PR #13, opened 2018-07-11
Italian localization

davidoskky:


berrnd:
I will provide a demo for each included language, so there is now one for Italian: :)

==> https://demo-it.grocy.info

PR #16, opened 2018-07-15
Norwegian localization

BlizzWave:
Norwegian localization


berrnd:
Thanks a lot! :)

I edited and included this in the release of today (version 1.14.0) and also create a demo at:
=> https://demo-no.grocy.info

PR #24, opened 2018-07-25
Updated for Version 1.16.0

BlizzWave:
Updated norwegian localization for version 1.16.0


berrnd:
Many thanks that you keep it updated! :)

PR #31, opened 2018-07-29
Update batteriesoverview.blade.php

BlizzWave:
Some german slipped in there too.


berrnd:
Thanks for testing/finding this! :)

PR #78, opened 2018-09-30
Grocy docker patch

talmai:
Hi Berrnd. Awesome job with Grocy. This PR allows anyone to run a dockerized version of Grocy.

PR #94, opened 2018-10-20
Grocy docker patch fix

talmai:
Fixes "blank white page after running docker".

Actually it fixes:

```
Warning: Use of undefined constant GROCY_USER_ID - assumed 'GROCY_USER_ID' (this will throw an Error in a future version of PHP) in /www/controllers/BaseController.php on line 47
```


berrnd:
Thanks - but I don't think this caused any problems, as it is only a WARNING for now...

talmai:
I agree, but that's the only thing coming up on my end, when I do a clean clone and build. What's happening on your end?

talmai:
You had mentioned (on the other pr), that you got white pages when trying to use it. The only thing I found was this warning - that when fixed, showed the app. Keep me posted if anyone says it isn't working with the docker version.

berrnd:
Sure - but was not me, was @cmantsch in https://github.com/berrnd/grocy/pull/78#issuecomment-430835742.

cmantsch:
> Sure - but was not me, was @cmantsch in [#78 (comment)](https://github.com/berrnd/grocy/pull/78#issuecomment-430835742).

yep it was me, and it seems like the changes fixed it. I now get the interface of grocy instead of a white page with only a warning.

Issue #139, opened 2019-01-07
API - A bit more RESTful


MrTarantula:
GET requests should not be able to modify or delete data. The correct verb should be used, according to [this article](https://restfulapi.net/idempotent-rest-apis/). Additionally, some routes can be made more RESTful. For example the Stock API can have a starting point for a product of `/api/stock/{productId}` and each function can be its own route. On the left are original routes and on the right are the RESTful routes:

```
GET /api/stock/add-product/{productId}/{amount} => POST/PUT /api/stock/{productId}/add
GET /api/stock/consume-product/{productId}/{amount} => POST/PUT /api/stock/{productId}/consume
GET /api/stock/open-product/{productId}/{amount} => POST/PUT /api/stock/{productId}/open
GET /api/stock/inventory-product/{productId}/{newAmount} => POST /api/stock/{productId}/inventory
GET /api/stock/get-product-details/{productId} => GET /api/stock{productId}/details OR /stock/{productId}
GET /api/stock/get-product-price-history/{productId} => GET /api/stock/{productId}/pricehistory
GET /api/stock/get-product-stock-entries/{productId} => GET /api/stock/{productId}/entries
```

I can come up with a complete list from the routes file. I could potentially make the required changes to the code to go from GET to POST but it might take me longer, as my PHP is a bit rusty.

berrnd:
Sounds good, and yes - the only place where I "followed the rules" was the `/api/file` group, however...

Feel free to provide the complete list, the API is used in various places in [/public/viewjs](https://github.com/grocy/grocy/tree/master/public/viewjs), so this needs also to be changed. And we need to increase the major version after that, or leave the old routes (maybe with redirects) to not confuse someone who is maybe already using the API for something...

berrnd:
I kind of completed this now - the Swagger UI now also reflects the current state, you can have a look at the already updated pre-release demo:
=> https://demo-prerelease.grocy.info/api

I also did some additional changes, e. g. grouped all stock-product-routes together under `/stock/products` and changed the response code to 204 when there is no response body.

I'm still unsure about some things:
- Use plural or singular for naming routes which return a single object - `/users/{userId}` vs. `/user/{userId}`
  - Currently the plural form is more used, so the `/file` routes should probably be changed too
[...]

PR #140, opened 2019-01-08
API - A bit more RESTful

gen2fish:
I tried to take a stab at https://github.com/grocy/grocy/issues/139

I left the generic entries as is since it seemed a lot of logic was built in and I didn't want to break anything big. I haven't spun up a Dev instance yet to test, but I would hate to have duplicate work.  If you see something glaringly bad, I can try to fix it. I'll do some testing in a docker lab here soon.

### DIFF
--- a/public/viewjs/batteriesoverview.js
+++ b/public/viewjs/batteriesoverview.js
@@ -27,7 +27,7 @@ $("#search").on("keyup", function()
 	{
 		value = "";
 	}
-	
+
 	batteriesOverviewTable.search(value).draw();
 });
 
@@ -41,7 +41,7 @@ $("#status-filter").on("change", function()
 
 	// Transfer CSS classes of selected element to dropdown element (for background)
 	$(this).attr("class", $("#" + $(this).attr("id") + " option[value='" + value + "']").attr("class") + " form-control");
-	
+
 	batteriesOverviewTable.column(4).search(value).draw();
 });
 
@@ -61,15 +61,15 @@ $(document).on('click', '.track-charge-cycle-button', function(e)
 	document.activeElement.blur();
 
 	Grocy.FrontendHelpers.BeginUiBusy();
-	
+
 	var batteryId = $(e.currentTarget).attr('data-battery-id');
 	var batteryName = $(e.currentTarget).attr('data-battery-name');
 	var trackedTime = moment().format('YYYY-MM-DD HH:mm:ss');
 
-	Grocy.Api.Get('batteries/track-charge-cycle/' + batteryId + '?tracked_time=' + trackedTime,
+	Grocy.Api.Post('batteries/' + batteryId + '/charged?tracked_time=' + trackedTime,
 		function()
 		{
-			Grocy.Api.Get('batteries/get-battery-details/' + batteryId,
+			Grocy.Api.Get('batteries/' + batteryId,
 				function(result)
 				{
 					var batteryRow = $('#battery-' + batteryId + '-row');
@@ -128,7 +128,7 @@ $(document).on('click', '.track-charge-cycle-button', function(e)
 function RefreshStatistics()
 {
 	var nextXDays = $("#info-due-batteries").data("next-x-days");
-	Grocy.Api.Get('batteries/get-current',
+	Grocy.Api.Get('batteries',
 		function(result)
 		{
 			var dueCount = 0;
@@ -146,7 +146,7 @@ function RefreshStatistics()
 					dueCount++;
 				}
 			});
-			
+
 			$("#info-due-batteries").text(Pluralize(dueCount, L('#1 battery is due to be charged within the next #2 days', dueCount, nextXDays), L('#1 batteries are due to be charged within the next #2 days', dueCount, nextXDays)));
 			$("#info-overdue-batteries").text(Pluralize(overdueCount, L('#1 battery is overdue to be charged', overdueCount), L('#1 batteries are overdue to be charged', overdueCount)));
 		},
--- a/routes.php
+++ b/routes.php
@@ -83,11 +83,11 @@ $app->group('/api', function()
 	$this->get('/get-openapi-specification', '\Grocy\Controllers\OpenApiController:DocumentationSpec');
 
 	// Generic entity interaction
-	$this->get('/get-objects/{entity}', '\Grocy\Controllers\GenericEntityApiController:GetObjects');
-	$this->get('/get-object/{entity}/{objectId}', '\Grocy\Controllers\GenericEntityApiController:GetObject');
-	$this->post('/add-object/{entity}', '\Grocy\Controllers\GenericEntityApiController:AddObject');
-	$this->put('/edit-object/{entity}/{objectId}', '\Grocy\Controllers\GenericEntityApiController:EditObject');
-	$this->delete('/delete-object/{entity}/{objectId}', '\Grocy\Controllers\GenericEntityApiController:DeleteObject');
+	$this->get('/object/{entity}', '\Grocy\Controllers\GenericEntityApiController:GetObjects');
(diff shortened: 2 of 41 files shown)


routes.php
```diff
@@ -102,46 +102,46 @@
...
-	$this->get('/stock/clear-shopping-list', '\Grocy\Controllers\StockApiController:ClearShoppingList');
-	$this->get('/stock/external-barcode-lookup/{barcode}', '\Grocy\Controllers\StockApiController:ExternalBarcodeLookup');
-	$this->get('/stock/undo-booking/{bookingId}', '\Grocy\Controllers\StockApiController:UndoBooking');
+	$this->post('/stock/{productId}/add/{amount}', '\Grocy\Controllers\StockApiController:AddProduct');
+	$this->post('/stock/{productId}/consume/{amount}', '\Grocy\Controllers\StockApiController:ConsumeProduct');
+	$this->post('/stock/{productId}/open/{amount}', '\Grocy\Controllers\StockApiController:OpenProduct');
+	$this->post('/stock/{productId}/inventory/{newAmount}', '\Grocy\Controllers\StockApiController:InventoryProduct');
+	$this->get('/stock/{productId}', '\Grocy\Controllers\StockApiController:ProductDetails');
+	$this->get('/stock/{productId}/pricehistory', '\Grocy\Controllers\StockApiController:ProductPriceHistory');
+	$this->get('/stock/{productId}/entires', '\Grocy\Controllers\StockApiController:ProductStockEntries');
```
MrTarantula: entires => entries

routes.php
```diff
@@ -102,46 +102,46 @@
...
+	$this->get('/stock', '\Grocy\Controllers\StockApiController:CurrentStock');
+	$this->get('/stock/volatil', '\Grocy\Controllers\StockApiController:CurrentVolatilStock');
```
MrTarantula: volatil => volatile

MrTarantula:
I like how the endpoints are structured, Good job. This is just a nit, but can you group the routes by verb, so all GETs are first, then POSTs, etc?

MrTarantula:
Also, there's the question of breaking API. Should it be a hard break? Or start versioning the API, e.g. `/api/v2/stock/{productId}`? @berrnd your call.

public/viewjs/batteriesoverview.js
```diff
@@ -61,15 +61,15 @@ $(document).on('click', '.track-charge-cycle-button', function(e)
...
 
 	Grocy.FrontendHelpers.BeginUiBusy();
-	
+
 	var batteryId = $(e.currentTarget).attr('data-battery-id');
 	var batteryName = $(e.currentTarget).attr('data-battery-name');
 	var trackedTime = moment().format('YYYY-MM-DD HH:mm:ss');
 
-	Grocy.Api.Get('batteries/track-charge-cycle/' + batteryId + '?tracked_time=' + trackedTime,
+	Grocy.Api.Post('batteries/' + batteryId + '/charged?tracked_time=' + trackedTime,
```
MrTarantula: Does this work? POST with query parameters is not typical. Data should be sent in POST body. POSTs will require changes on server side as well.

berrnd:
I would prefer a hard break, so a major version 2.x when this is ready. Currently it would be no problem, because the logic behind the API will not really change for now, but when this happens, it's maybe too cumbersome to keep the old version of the API alive...

routes.php
gen2fish: Fixed in https://github.com/grocy/grocy/pull/140/commits/e416695cc4aa7c77b9f5e2d0d4f7dbbfef843856

routes.php
gen2fish: https://github.com/grocy/grocy/pull/140/commits/e416695cc4aa7c77b9f5e2d0d4f7dbbfef843856

gen2fish:
Did some cleaning in https://github.com/grocy/grocy/pull/140/commits/e416695cc4aa7c77b9f5e2d0d4f7dbbfef843856

Moved the gets on top and posts on bottom per section and PUTs for user/object edits.

Working on removing the verbs in the routes.

gen2fish:
Would you rather have the same route used for GET and POST or for new items have a /new?

```php
$this->get(/object/{entity} // Get Object
$this->post(/object/{entity} // Create Object
```

Or

```php
$this->get(/object/{entity} // Get Object
$this->post(/object/{entity}/new // Create Object
```

I would suspect we can use the same route with different methods for each section.

berrnd:
I think the later (same route, different methods) is the recommended way...

gen2fish:
Verb-less generic routes

https://github.com/grocy/grocy/pull/140/commits/6b41e5212c27c881b96379fb993a7bbfd6636ca0

I'll spin up a test instance when I got home tonight and see if everything is broken.

I'll move on the Stock with the amount in the URL string next and try to put it in the JSON body.

berrnd:
Thanks.

At least all `PUT` routes will be affected, because `Grocy.Api.Put` does not exist as a JS function yet.
Also [grocy.openapi.json](https://github.com/grocy/grocy/blob/master/grocy.openapi.json) needs to be adjusted to reflect the changes in the integrated [Swagger UI](https://demo-en.grocy.info/api) instance.

gen2fish:
I created the function quick in https://github.com/grocy/grocy/pull/140/commits/3db750e2119bb546a106c85456c030a14d23bd6f

gen2fish:
Spun up a test instance. Testing is going well.  Only one issue with the volatile stock route.  Getting this error on startup

```
Fatal error: Uncaught FastRoute\BadRouteException: Static route "/api/stock/volatile" is shadowed by previously defined variable route "/api/stock/([^/]+)" for method "GET" in /www/vendor/nikic/fast-route/src/DataGenerator/RegexBasedAbstract.php:95 Stack trace: #0 /www/vendor/nikic/fast-route/src/DataGenerator/RegexBasedAbstract.php(30): FastRoute\DataGenerator\RegexBasedAbstract->addStaticRoute('GET', Array, 'route74') #1 /www/vendor/nikic/fast-route/src/RouteCollector.php(44): FastRoute\DataGenerator\RegexBasedAbstract->addRoute('GET', Array, 'route74') #2 /www/vendor/slim/slim/Slim/Router.php(227): FastRoute\RouteCollector->addRoute(Array, '/api/stock/vola...', 'route74') #3 /www/vendor/nikic/fast-route/src/functions.php(25): Slim\Router->Slim\{closure}(Object(FastRoute\RouteCollector)) #4 /www/vendor/slim/slim/Slim/Router.php(238): FastRoute\simpleDispatcher(Object(Closure), Array) #5 /www/vendor/slim/slim/Slim/Router.php(191): Slim\Router->createDispatcher() #6 /www/vendor/slim/slim/Slim/App.php(585): Slim\Router->di in /www/vendor/nikic/fast-route/src/DataGenerator/RegexBasedAbstract.php on line 95
```

Once I comment out line 116 in routes.php
```php
	$this->get('/stock/volatile', '\Grocy\Controllers\StockApiController:CurrentVolatilStock');
```

Application starts successfully.

So far I have tested the Stock GETs

MrTarantula:
Do the new POSTs work? They shouldn't, as far as I can tell. Although I'm not really a PHP dev, I'd like to contribute something. I am a web dev and MVC is similar enough in any language.

I should have a bit of time this weekend to work on the POSTs on the server side, unless you get to it first. I'll submit a PR to your branch when I'm ready.

gen2fish:
I haven't gotten to modifying the posts to have the data in the json body instead of the url. They work in their current form but I haven't spent much time on getting them fixed. Was hoping to get some time this weekend myself.

berrnd:
I'll merge this now and complete it this weekend - thanks @gen2fish and @MrTarantula! :)

Further progress will be tracked in #139.

PR #171, opened 2019-03-10
Add fetch by barcode API method

matejdro:
At some point in the future I plan on making small Android app that would just allow quick scanning of all items via phone's camera after coming from the store.

For that, API needs to have a way to retrieve product by barcode.

I've almost no experience in PHP so far, so please criticize as much as you want.

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -26,6 +26,19 @@ class StockApiController extends BaseApiController
 		}
 	}
 
+	public function ProductDetailsByBarcode(\Slim\Http\Request $request, \Slim\Http\Response $response, array $args)
+	{
+		try
+		{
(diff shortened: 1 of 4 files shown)


berrnd:
> At some point in the future I plan on making small Android app that would just allow quick scanning of all items via phone's camera after coming from the store.

Cool! Looking forward to it. :)

Looks good by a first look - will check and add it later...

rpbx:
Hello @matejdro ,

Did you ever end up developing an app for scanning? I know that this is an old thread and grocy has a built in scanner. However, I find that the scanner is giving a lot of wrong scans.

Thank you

matejdro:
Sorry, I've not gotten there yet.

PR #200, opened 2019-03-29
Fix typo (localization de)

skevas:


berrnd:
Thanks!

But localizations are managed on Transifex:
https://www.transifex.com/grocy/grocy

Do you want to change it there or should I just integrate it?

skevas:
You're welcome. I never used transifex. It would be nice if you could make the change.

berrnd:
Ok - done.

PR #212, opened 2019-04-25
Fixed missing #

rabbadab:
Added a # to a translation to fetch correct number of days until expiring.


rabbadab:
Ahh issed that! 
I'll fix it in transifex!

berrnd:
Thanks for fixing it on Transifex - closing this here.

PR #220, opened 2019-04-27
Adding the barcode scanning script

TheFuzz4:
So you inspired me to enhance what little I did know about python.  I wrote this script for adding new items to the system.  Increasing the quantity as well as decreasing. 
The script requires the user to print off a barcode (or use some barcode from something else) to setup as the ADD_ID.  When this is scanned it will put the script into add mode and start increasing.  Scan that code again and it shifts back to consume mode.  Out of the box this runs in consume mode.  I just bought a cheap scanner and plugged the USB dongle into one of the pi's in my house and just run the script locally there.  Right now I run the script in a screen session since I don't have any logging setup in the script.

I am by no means a developer and I am sure there are areas of my script that can use some enhancements or maybe some better logic.  I optimized it to the best of my abilities for the moment and I'm sure I'll have further enhancements with it.  But hey it gets the job done.  Your API is great and without it this would be a lot harder to accomplish.  I hope that this helps people.  

If there is a DB that you don't want it to search externally just set the token to empty and it will skip on over it.  

I have issues with rapid scanning, my scanner starts to report in only pieces of the UPC so I have to give it a few second pause between scans but only when its something that isn't in the system already.  If its local the script fires off really fast. 

Well let me know what you think about this and I'll continue to do my part to help contribute to the project.

### DIFF
--- a/scripts/barcode_reader.py
+++ b/scripts/barcode_reader.py
@@ -1,0 +1,240 @@
+#!/usr/bin/python
+import signal, sys
+import json
+import requests
+
+from evdev import InputDevice, categorize, ecodes
+
+#This will allow you to work the grocy inventory with a barcode scanner
+UPC_DATABASE_API='YOUR_TOKEN'
+buycott_token='YOUR_TOKEN'
+GROCY_API='YOUR_TOKEN'
+add_id='YOUR_ADD_BARCODE'
+base_url = 'YOUR_BASE_URL'
+ADD = 0
+barcode = ''
+found = ''
+device = InputDevice('/dev/input/event0') # Replace with your device
+scancodes = {
+	11:	u'0',
+	2:	u'1',
+	3:	u'2',
+	4:	u'3',
+	5:	u'4',
+	6:	u'5',
+	7:	u'6',
+	8:	u'7',
+	9:	u'8',
+	10:	u'9'
+}
+NOT_RECOGNIZED_KEY = u'X'
+
+def increase_inventory(upc):
+    #Need to lookup our product_id, if we don't find one we'll do a search and add it
+    product_id_lookup(upc)
+    print ("Increasing %s") % (product_name)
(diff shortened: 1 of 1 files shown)


berrnd:
Thanks! :)

Great idea, already also thought about a "scan mode" in the UI which does also this - removing or adding just one item per scanned barcode.

The only problem I see about this approach is that you don't get feedback when something has not worked. My initial idea was to provide maybe audio feedback, to still be able to scan "blindly" anyway, without having to look always on the screen.

So I don't know if it's the best place to just include the script here, maybe a separate repository "grocy-tools" or something fits better, or even a standalone repository just for this, because it would also need a bit of explanation how to use it and maybe there will be also issues/discussuions when somebody get to use it - so if you want, we would just need a repo name and I will set it up, give you write access and you can add/maintain everything there... (?)

berrnd:
> we would just need a repo name and I will set it up

What about "grocy-pyscanner"?

TheFuzz4:
That sounds great to me.  Yeah I've been trying to figure out some method
of giving you feedback.  Best idea I have right now is to have my script
call my home assistant via API and give feedback that route.  Still
thinking.  But yeah that repo name works for me.  Thank you for your
feedback.

Jason

berrnd:
Great, I created the repository and added you with write permissions:

https://github.com/grocy/grocy-pyscanner

Feel free to add there anything you want to make this a great little helper tool. :)
Of course we can then also add a hint on the main repos README and on the homepage...

So closing this here for now.

TheFuzz4:
Thank you I will do just that.

TheFuzz4:
Its saying that I don't have permissions to do a push to the repo

berrnd:
Seems that you don't have accepted the invitation yet - should be visible for you under https://github.com/grocy/grocy-pyscanner/invitations and for the org here: https://github.com/grocy

TheFuzz4:
Thanks that did it

PR #290, opened 2019-06-22
Allow possibility to choose custom homepage

### LINKED ISSUE #294: Redirection to Page Not found when feature deactivated
When clicking the Grocy logo, or when entering just the bare URL, I'm getting redirected to the equipment section as a sort of Start page. This is problematic because I actually deactivated that feature, and therefor getting a Page Not found error. Can the Start page be changed in any way? Any activated feature would be great, selectable would even be better.

nielstholenaar:
Hi.

I wanted to changed the default view to the shopping list instead of the stock overview. After a little bit thinking i thought why not make in configurable. 

This PR introduces a new Setting "GROCY_ENTRY_PAGE" which can be set to any of the available features. (stock,shoppinglist, .......)

This will close #294.

### DIFF
--- a/config-dist.php
+++ b/config-dist.php
@@ -46,8 +46,9 @@ Setting('STOCK_BARCODE_LOOKUP_PLUGIN', 'DemoBarcodeLookupPlugin');
 # set this to true
 Setting('DISABLE_URL_REWRITING', false);
 
-# By default the homepage will be set to the stock overview.
-# You can set this to any overview you want. Example: Use recipes to set the homepage to the recipes overview.
+# Specify an custom homepage if desired. By default the homepage will be set to the stock overview.
+# You can chosen one of the following values:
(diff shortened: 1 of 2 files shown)


berrnd:
Great idea, thank you! :)

But I think this would result in an `404` error when you set it for example to `batteries` and set at the same time `Setting('FEATURE_FLAG_BATTERIES', false);` ... so maybe some handling for this would be needed as well...

nielstholenaar:
I'm currently on a small vacation. When I'm back home I will add some checking to prevent any 404

nielstholenaar:
I have made the necessary changes to handle any the disabling of an specific feature.

I also added a fallback to handle the scenario when the requested entry page has been disabled. In that case the about page will be served instead.

nielstholenaar:
I have updated the docblock in the config-dist.php file

berrnd:
Thank you, merged it. :)

PR #315, opened 2019-07-13
Use list title when display a user field form

oncleben31:
To replace the default string `Product group`


berrnd:
Thank you :), great find - was probably my bad by copy & pasting things... :D

PR #338, opened 2019-08-04
Add API call to remove an item from the shopping list by productid

Forceu:
This adds an API call, so a shopping list item can easily be removed

Please note: This is **UNTESTED**, as I was unable to run the current grocy master commit on my server.

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -67,7 +67,7 @@ class StockApiController extends BaseApiController
 				throw new \Exception('An amount is required');
 			}
 
-			$bestBeforeDate = date('Y-m-d');
+			$bestBeforeDate = null;
 			if (array_key_exists('best_before_date', $requestBody) && IsIsoDate($requestBody['best_before_date']))
 			{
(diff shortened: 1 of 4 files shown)


berrnd:
Thank you! :)

Looks good to me at first sight, will review and integrate this soon...

Forceu:
Thanks for the quick reply! Sorry I am new to Github and am trying to make it two seperate PRs, somehow they got merged however.

The second PR is another change to the API as well, where the default best before date is used if no parameter is given for AddProduct().

berrnd:
I've merged this now with some little naming changes - see https://github.com/grocy/grocy/commit/cc1c6a6442cd2c958787950cadc59a83aa4f53cb.

Also when no best before date is supplied and also the given product has no "Default best before days" set, I changed it to still us today - maybe not the best thing, but it behaved like that so far and I don't want a breaking change for that now...

Thanks again. :)

Forceu:
I completely forgot that, yes you are right. Great to see the change merged! :)

PR #343, opened 2019-08-13
API: Return stock amount of 0 instead of null, if product is not in stock

Forceu:
Without this commit, the API returns null instead of 0 when queried for product stock amount


berrnd:
Thanks! :)

PR #349, opened 2019-08-30
Add API call for adding products to shoppinglist

Forceu:
Adds an API call to add items to a shoppinglist

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -289,6 +289,41 @@ class StockApiController extends BaseApiController
 	}
 
 
+	public function AddItemToShoppingList(\Slim\Http\Request $request, \Slim\Http\Response $response, array $args)
+	{
+		try
+		{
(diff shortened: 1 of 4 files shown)


berrnd:
Thank.

I merged it with little changes/fixes - see https://github.com/grocy/grocy/commit/a4454f825a74b4e51dfad94b0da5b74b8fa23246:

- Maybe you have not tested it - it did not work for existing shopping list entries because the product id was not taken here:
https://github.com/Forceu/grocy/blob/d4eb767f1b602e5279c23c991d1b5d52f6fa0c28/services/StockService.php#L559
- In the same line also the given shopping list id was not used
- There was no check that the given product id needs to be a valid one

But absolutely no problem, thanks again. :)

Forceu:
Thank you! I did some testing, but did not test your cases and always used the default shopping list, that's why I did not noticed these problems.

Issue #245, opened 2019-05-09
Shopping list output options


teamtmedia:
Firstly, as always, "Thank you" for sharing such a great project.

Just using the shopping list and realise that there are no options for outputting.  Now I realise that holding a device is a decent possibility but even that could do with an enhancement.  So my thoughts are:

1) Add a notes section to the shopping list page where we can add extra bits and bobs that are off the system (cleaning products, etc).  This would be useful for people using the app or option 2

2) Add a print button that produces a clear shopping list

3) (Potentially) Add a button that can send a calendar appointment (.ical) to an email address (handy for sharing the list with 3rd parties)

Keep on trucking, loving the work.

berrnd:
For 1.:
You can add items without a product, just fill in the notes field of the shopping list item...

hauweg:
For 2: 
would love to have a print button!! <3

For 1: yes, that would be great! bernd your suggestion is a working workaround, but not a really nice one :)

berrnd:
For 1:
Is it a that big difference if the textbox is inside a shopping list item or on the list page directly? Of course we can also add a textfield there...

hauweg:
1. it does not habe the posibility to ad a product group.
2. i am using the note field now, to mark what is ordered allreay and when.
3. in order to have a nice layout and by that a good overview, every item oth shoping list, should have the same layout

JoschaMiddendorf:
It would also be great to have the ability to subscribe to a shopping list as iCal Reminders list. In that way shopping lists could be viewed end checked by iOS reminders App.

berrnd:
Shopping lists now have an notes section/field and can also be printed. :)

See the pre-release demo (next release will be dropped on the coming Sunday): https://demo-prerelease.grocy.info/shoppinglist

But I'm unsure about the .ical export. Should that be just an appointment for today and the shopping list as the body/text? Do you have an example for such a file?

Issue #199, opened 2019-03-24
Feature Request: Android App that works offline?


ignacio82:
Is there an android app that works when I don't have internet access and syncs to my server when I do?

berrnd:
No.

The UI is responsive, works pretty well on mobile devices, I use it personally almost daily - but I don't have the problem to have areas without an internet connection...

The API should cover everything needed to build an app, feel free to build it - there was also already an idea about an app to do the scanning part, see https://github.com/grocy/grocy/pull/171#issue-259752589.

ASMfreaK:
@ignacio82 You can launch grocy from @termux (via nginx, for example) and use browser on your phone to navigate to it. Syncing may be not hard if you sync your database and config between devices using rsync/dropbox/syncthing/etc. AFAIK, sqlite databases are portable, if grocy does not rely on some sqlite extensions like fts5, which may not be compiled into termux's php sqlite binding. 
Please, @berrnd , cc which extensions grocy use.
**Beware of data losses and make backups of database before syncs!**
This method is only useable with exclusive device switch (e.g. you *use* **only** your phone , then *sync* and *switch* to *use* **only** your pc, sync, switch back to use **only** phone and so on). It will make you lose data if you don't sync all devices **before** you switch to them.

Altheran88:
I agree, mobile app should be another repo. Using API is a better approach vs DB cloning.

BlizzWave:
This is what alot of us want, but i requires someone to actually develop it. Bernd is a busy man trying to keep up with all feature requests and bugs. Is there anyone of you that could assist in developing a android/ iphone app?

berrnd:
grocy does not use any SQLite extensions - the default `pdo_sqlite` PHP driver works. The only requirement is that it needs to be at least SQLite version 3.8.3 - there will be SQL errors otherwise (I think Window Functions were not part of SQLite before, which are used in some views).

So syncing the database should absolutely work.

Running grocy completely on the phone with Termux could also work, although it would be a little bit heavy - I haven't really used Termux with a complete webserver so far...

Maybe Termux could also be used to package grocy into a standalone app (?), that's the way I do it with [grocy-desktop](https://github.com/grocy/grocy-desktop) - a (Windows) desktop application wrapper which runs the in PHP integrated development server to host grocy (available since this week also [directly through the Microsoft Store](https://www.microsoft.com/de-de/p/grocy/9nwb1trnnksf) 😄).

[...]

PR #352, opened 2019-09-01
Add a very simple javascript-only shopping list that works offline

gersilex:
The OfflineShoppingList disables auto-reloading by adding the 'fullscreen-card' class to the body. It also requests a confirmation on leaving the page because nothing is saved, not even into the local or session storage. Please consider this feature as a rough draft and foundation for discussion.

I really like the checkbox of the default shopping list, because it works with multiple users at once and everyone gets synced the updates. The space, on the other hand, is really rare because of all the other buttons that are not useful when being shopping right now. Maybe we can combine the features somehow. I still don't know how we could handle this when being offline or not connected to the home network by VPN. How do you handle shopping trips?

### DIFF
--- a/controllers/StockController.php
+++ b/controllers/StockController.php
@@ -81,6 +81,25 @@ class StockController extends BaseController
 		]);
 	}
 
+	public function OfflineShoppingList(\Slim\Http\Request $request, \Slim\Http\Response $response, array $args)
+	{
+		$listId = 1;
+		if (isset($request->getQueryParams()['list']))
+		{
+			$listId = $request->getQueryParams()['list'];
+		}
+
+		return $this->AppContainer->view->render($response, 'offlineshoppinglist', [
+			'listItems' => $this->Database->shopping_list()->where('shopping_list_id = :1', $listId),
+			'products' => $this->Database->products()->orderBy('name'),
+			'quantityunits' => $this->Database->quantity_units()->orderBy('name'),
+			'missingProducts' => $this->StockService->GetMissingProducts(),
+			'productGroups' => $this->Database->product_groups()->orderBy('name'),
+			'shoppingLists' => $this->Database->shopping_lists()->orderBy('name'),
+			'selectedShoppingListId' => $listId
+		]);
+	}
+
 	public function ProductsList(\Slim\Http\Request $request, \Slim\Http\Response $response, array $args)
 	{
 		return $this->AppContainer->view->render($response, 'products', [
--- a/routes.php
+++ b/routes.php
@@ -50,6 +50,7 @@ $app->group('', function()
 		$this->get('/shoppinglist', '\Grocy\Controllers\StockController:ShoppingList');
 		$this->get('/shoppinglistitem/{itemId}', '\Grocy\Controllers\StockController:ShoppingListItemEditForm');
 		$this->get('/shoppinglist/{listId}', '\Grocy\Controllers\StockController:ShoppingListEditForm');
+		$this->get('/offlineshoppinglist', '\Grocy\Controllers\StockController:OfflineShoppingList');
 	}
 
 	// Recipe routes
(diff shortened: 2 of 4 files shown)


berrnd:
> I still don't know how we could handle this when being offline or not connected to the home network by VPN.

Maybe the simpler option would be to just export the shopping list, as an mail or print it or even export it as an calendar entry as suggested in #245.

> How do you handle shopping trips?

I just use the shopping list - have exposed my private grocy install to the internet and (very, very) rarely have the case that I'm so offline that I cannot even load the shopping list...

gersilex:
My initial long-term goal is to have a Google Keep-like shared shopping list that syncs online but also works offline. [A syncing service for WebApps and native apps by AWS](aws.amazon.com/appsync) seems to make it very easy and I would like to sponsor the infrastructure cost for this amazing project to make it work.

I thought of something like this:

- Users click on the button "Sync to GrocyShopper" on /shoppinglist
- The grocy instance is uniquely registered by the GrocyShopper API
- A QR Code is shown
- Users open the GrocyShopper App which asks to scan the QR Code
- Shopping list data is there, including categories and synced with all other instances of GrocyShopper and with Grocy via the sync service. Checking off entries is equivalent to checking of in Grocy /shoppinglist (green tick)
- More users may join through the same QR code

I don't have a mock-up but I hope this description is somehow useful. That's what I aim for. :)

berrnd:
Ok, sounds interesting - I'm looking forward to any integration and I'm happy to help with any questions or adding API endpoints if there is something missing to get the job done... :)

But my first goal for grocy was to keep it simple. Well, this is maybe not true anymore because a lot of (detail) features came (and will come) by all the requests. Which is great and I enjoy working on. But that here sounds a little too complicated to me for the end result. I can think of a little "companion app" for things like the shopping list (see #199), to have it offline and "more native".

Second, but that's just my personal opinion for private stuff: I never host "my data" anywhere else than my "physically owned" infrastructure, regardless of how "private" the content is. Sounds maybe a little bit paranoid, but that's at the end what [#self-hosted](https://github.com/topics/self-hosted) and [#ownyourdata](https://github.com/topics/ownyourdata) is about (for me)... :)

gersilex:
> a little "companion app" for things like the shopping list

That's what I planned to do. I would create it as a PWA (or maybe Ionic App) instead, because I don't really need native functionality here.

> I never host "my data" anywhere else than my (physically owned) infrastructure

I strongly agree but unfortunately getting the family to enable the VPN before having a look into the shopping list and disabling afterwards is just too complicated for them. I can understand that they are looking for simplicity like Google Keep that just "works" everywhere, anytime, even offline.

>  I'm looking forward to any integration and I'm happy to help

Thank you very much. What would be the most acceptable way of integration a kind of entrypoint into my companion app when it's ready? Would you add a generic `Export ...` dropdown button and list my companion app there or do you want to keep all integrations away (apart from maybe printing or mailing) ? I would like to design it with you user workflow in mind.

berrnd:
> What would be the most acceptable way of integration...

Feel free to suggest whatever you think would work best. Currently there is no such integration, so would be the first one, but why not...

> getting the family to enable the VPN before having a look into the shopping list and disabling afterwards is just too complicated for them

I personally have a shortcut to https://grocy.berrnd.org/shoppinglist on my smartphones homescreen (my private grocy instance, I often don't see the benefit of a VPN in front of a TLS secured connection when I trust the authentication implementation), grocy has a [Web-App-Manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest) [included](https://github.com/grocy/grocy/blob/master/public/img/appicons/site.webmanifest) (although it utilizes no PWA features), the session never expires - so opening that shortcut already feels not that different and takes not longer than any other app (despite the offline functionality) ... just saying...

gersilex:
Closing this for now. Your last message inspired me to ditch the approach of creating something new and instead make a connection to the grocy instance available to the users.

I will continue to work on a simplified, even more shopping-friendly view of the shopping list but will re-introduce the backend syncing of checked-off items. I'll open another pull request as soon as I think this might fit into the grocy product :) Thank you for the interesting discussion.

PR #369, opened 2019-09-23
public mealplan: add fullscreen recipe in new tab (fixes #367)

### LINKED ISSUE #367: Feature Request: Open Recipe Fullscreen when click on MealPlan picture
Thank you for the new updates today, I am still exploring using Grocy and all the functions of it for my family.

It would be nice to be able to see the full recipe straight from the meal plan so that we can pull it up while cooking. Either through clicking the picture or another large icon to click above the current options.

kriddles:
fixes #367

### DIFF
--- a/public/viewjs/mealplan.js
+++ b/public/viewjs/mealplan.js
@@ -28,12 +28,12 @@ var calendar = $("#calendar").fullCalendar({
 		{
 			UpdateUriParam("week", view.start.format("YYYY-MM-DD"));
 		}
-		
+
 		$(".fc-day-header").append('<a class="ml-1 btn btn-outline-dark btn-xs my-1 add-recipe-button" href="#"><i class="fas fa-plus"></i></a>');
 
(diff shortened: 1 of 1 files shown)


kriddles:
Ok, give this popup a look. Thanks for providing an example to look at.

I noticed that the grid for mobile for the Meal Plan is not very responsive and the buttons are not all visible. - using google chrome pixel 3
Any ideas?

berrnd:
> I noticed that the grid for mobile for the Meal Plan is not very responsive and the buttons are not all visible. - using google chrome pixel 3

Hm, seems to be not responsive at all...

That's basically just the same as the calendar, forced to view only weeks - the component used is [FullCalendar](https://fullcalendar.io/)...

berrnd:
Merged this now with some little changes, see https://github.com/grocy/grocy/commit/6bb9d2c51db0fa82c8516ea26af0b0b9d3baa2f1.

- There were doubled y scrollbars in the popup
- I think it's better to hide the buttons in the recipe card header when displaying a recipe embedded, as the popups/questions would be behind the fullscreen-card

PR #373, opened 2019-09-23
public mealplan: if mobile change view to agenda day (Fixes #372)

### LINKED ISSUE #372: Meal Plan Mobile Responsive Switch to Agenda View
> I noticed that the grid for mobile for the Meal Plan is not very responsive and the buttons are not all visible. - using google chrome pixel 3

Hm, seems to be not responsive at all...

That's basically just the same as the calendar, forced to view only weeks - the component used is [FullCalendar](https://fullcalendar.io/)...

_Originally posted by @berrnd in https://github.com/grocy/grocy/pull/369#issuecomment-534240712_

Research
https://stackoverflow.com/questions/52751335/full-calendar-mobile-view

kriddles:
fixes #372

### DIFF
--- a/public/viewjs/mealplan.js
+++ b/public/viewjs/mealplan.js
@@ -120,6 +120,9 @@ var calendar = $("#calendar").fullCalendar({
 		{
 			$("#calendar").fullCalendar("gotoDate", GetUriParam("week"));
 		}
+		if( /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent) ) {
+			$('#calendar').fullCalendar('changeView', 'agendaDay');
+		} //IF MOBILE CHANGE VIEW TO AGENDA DAY
 	},
 });
 


berrnd:
(Just off topic: If you put [`fixes #372`](https://help.github.com/en/articles/closing-issues-using-keywords) in the description of the pull request, not the title, it will link the issue and automatically close it once this got merged...)

kriddles:
This was just a quick googling solution. 
It looks much better on my android mobile and tablet while retaining the week view on my computer screen. It provides the arrows to move forward and backward on days and includes the jump to "today" still.

I would think it would/could be responsive based on screen/viewport size and drop down to agendaDay and maybe there is a way?

berrnd:
Also googled around a little bit and maybe using the `agendaDay` view is a viable solution for now as FullCalendar seems to lack to be responsive in general for the month and week view...

Would also apply to the [/calendar](https://en.demo.grocy.info/calendar) page, and maybe needs to also listen on the `window.resize` event as the screen size could change after the view is rendered (e. g. rotating the mobile phone screen).

> I would think it would/could be responsive based on screen/viewport size and drop down to agendaDay and maybe there is a way?

Should be just `$(window).width()`, would use the same [breakpoints as Bootstrap](https://getbootstrap.com/docs/4.3/layout/overview/#responsive-breakpoints) and maybe switch the view to `agendaDay` on `< md`....

berrnd:
Did that now like you suggested, see https://github.com/grocy/grocy/commit/2a9f927a137b057c027797f6f6018d08ef61837c, thanks for the idea! :)

PR #397, opened 2019-09-27
Fixed navigation menu and full screen recipe smooth scrolling on iOS

DeeeeLAN:


DeeeeLAN:
Yeah for whatever reason iOS webkit doesn't use smooth scrolling on overflow divs without it. I guess you could make sure the change didn't break anything in your normal browser, but I imagine everything is fine. I am also looking at fixing how some of the buttons and other interface items wrap on smaller screens so that everything is a little more clean.

PR #415, opened 2019-10-10
Fix meal plan start page requirements

lwis:
I missed a requirement on my previous pull request, sorry.

### DIFF
--- a/controllers/SystemController.php
+++ b/controllers/SystemController.php
@@ -85,9 +85,9 @@ class SystemController extends BaseController
 		if ($entryPage === 'calendar' && constant('GROCY_FEATURE_FLAG_CALENDAR')) {
 			return '/calendar';
 		}
-		
+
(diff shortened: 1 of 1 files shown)


lwis:
If I enable the meal plan without enabling calendar, I didn't get the calendar functionality (I'd need to double check that, though). I thought the feature might be dependant?

berrnd:
Not intentional, I will check that...

berrnd:
So I think the only problem was that the menu entry in the sidebar was hidden when `FEATURE_FLAG_CALENDAR` was set to `false` (the meal plan is included in `FEATURE_FLAG_RECIPES`).

So should be fixed by https://github.com/grocy/grocy/commit/3328c789d41820edc1896312fdad4ebbbd1881f8.

Thanks for finding that! :)

lwis:
Ah, np!

PR #416, opened 2019-10-10
Rename location to "Default location" on product page

lwis:
Given that the location can be set on the purchase page, I thought it was more appropriate to list this as "Default location".

I wasn't sure how the translations work, so if I need to make a change to a different file please let me know.

### DIFF
--- a/localization/en/strings.po
+++ b/localization/en/strings.po
@@ -159,6 +159,9 @@ msgstr "Add"
 msgid "Name"
 msgstr "Name"
 
+msgid "Default location"
+msgstr "Default location"
+
 msgid "Location"
 msgstr "Location"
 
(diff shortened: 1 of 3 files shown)


lwis:
Hmm, I've tried running in `dev` mode but I'm not having much luck - anything else I need to do?

berrnd:
Thanks, merged it. :)

> Hmm, I've tried running in `dev` mode but I'm not having much luck - anything else I need to do?

No, it runs directly in the localization function ([backend](https://github.com/grocy/grocy/blob/3328c789d41820edc1896312fdad4ebbbd1881f8/services/LocalizationService.php#L135-L147), [frontend](https://github.com/grocy/grocy/blob/9ba66aeac2a023f00a0a01f0a5f72ed0b9181c2a/public/js/grocy.js#L232-L236)) and only checks for the `GROCY_MODE` setting, maybe the `localization` directory is not writable?

PR #417, opened 2019-10-11
Calendar descriptions (iCal export)

### LINKED ISSUE #368: Feature request: Meal Plan included in Calender
Can the meal plan be included in the calendar as an option so that the meals flow alongside with the calender ical syncs?

kriddles:
Adjustment to include description with a link to the meal plan in the calendar event.
fix #368

### DIFF
--- a/controllers/CalendarApiController.php
+++ b/controllers/CalendarApiController.php
@@ -38,6 +38,7 @@ class CalendarApiController extends BaseApiController
 				$vEvent->setDtStart($date)
 					->setDtEnd($date)
 					->setSummary($event['title'])
+					->setDescription($event['description'])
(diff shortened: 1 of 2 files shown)


kriddles:
Thanks, I updated to add the week query parameter 

I also tested syncing a battery calendar event and I did not see any errors. It showed up in my google calendar without a description

berrnd:
Great! :)

I merged it with just a little change:
`GROCY_BASE_URL` most of the time is only set to `/` - so to get the proper full URL the class [`UrlManager`](https://github.com/grocy/grocy/blob/master/helpers/UrlManager.php) is used, see https://github.com/grocy/grocy/commit/21b9c1a8aaf3c0a6d7e2de7a8be28afab36f1bff.

PR #423, opened 2019-10-15
Purchase factor

kriddles:
Add Value and Purchase Factor Information


kriddles:
I'll retake a look at this and close it for now.

PR #424, opened 2019-10-15
Improve mobile performance on stock overview

lwis:
I have 143 products and 442 units on stock overview, when I search on my phone the page performance is really poor - this aims to alleviate that.

### DIFF
--- a/public/viewjs/stockoverview.js
+++ b/public/viewjs/stockoverview.js
@@ -1,5 +1,7 @@
 ﻿var stockOverviewTable = $('#stock-overview-table').DataTable({
 	'paginate': false,
+	'deferRender': true,
+	'searchDelay': 200,
 	'order': [[3, 'asc']],
 	'columnDefs': [
 		{ 'orderable': false, 'targets': 0 },


berrnd:
Thanks! :)

But I think only `searchDelay` makes then the relevant difference, as [`deferRender`](https://datatables.net/reference/option/deferRender) only changes something when the table data is loaded from JavaScript - that would be be the real performance improvement (including paging and so on), but I had no time so far to implement this as an API endpoint and then for all tables. Currently all tables are populated server side as plain HTML (the worst site on my private instance is the stock journal page with about 7600 lines, takes (on a fast desktop machine) about 10 seconds to fully load...

lwis:
@berrnd yikes! I think deferRender also works for re-draws on the table (which happen when searching). I can't imagine how a 7600 row table load would perform on my phone 😝 let alone searching!

I can remove the searchDelay, the only reason I added it was to debounce the potentially expensive redraws while typing, but maybe deferRender will help with that?

Yeah, I was trying to work out if there was a common configuration applied to DataTables, but couldn't find anything - would be handy!

berrnd:
If it feels faster, let's add it like that... :)
Already did some other "workarounds" like hiding the table while DataTables is initialising (also made a bigger difference in loading performance).

> Yeah, I was trying to work out if there was a common configuration applied to DataTables, but couldn't find anything - would be handy!

Yes, currently all options are on every page (and are mostly the same everywhere).

But the only real solution would be to load the data via JS/API, implement paging and also searching (server side / via API)...

berrnd:
I've merged and tested this, but I cannot see that [`searchDelay`](https://datatables.net/reference/option/searchDelay) makes a difference (tried to set this for example to `5000`, it does not wait 2 seconds after typing something in the serch field), as this only works for the DataTables built-in search and we use a own search field/handler function ([example](https://github.com/grocy/grocy/blob/9e1804252e0b22f004076df497d0193593598464/public/viewjs/stockoverview.js#L56-L65)).

I've also centralized the always-the-same default init settings in https://github.com/grocy/grocy/commit/9e1804252e0b22f004076df497d0193593598464 ([doc reference](https://datatables.net/examples/advanced_init/defaults.html)).

berrnd:
> we use a own search field/handler function

I've added now a delay for all table search fields (https://github.com/grocy/grocy/commit/e1335088148b3f5d546ab4b8db9cbe100c983273).

lwis:
@berrnd nice! Thanks.

PR #436, opened 2019-10-29
Set reasonable Step for Currency

### LINKED ISSUE #435: Incrementing Inventory Price with Arrow Buttons alters price with wrong precision
Attempting to adjust a price, for example `21.37`, by using the form's clickable arrows increases/decreases the value by `0.0001`.  For example:

Initial State:
image

After 3 taps of the Up Arrow:
image

Sanfam:
@berrnd, Feel free to reject and close this PR this if you have a particular reason for setting it to `0.0001`, but it appeared super random and was super easy to alter so I thought I'd give this a shot. 

Adjusts currency selector step from `$0.0001` to `$0.01`

Fixes #435

### DIFF
--- a/views/inventory.blade.php
+++ b/views/inventory.blade.php
@@ -59,7 +59,7 @@
 				'id' => 'price',
 				'label' => 'Price',
 				'min' => 0,
-				'step' => 0.0001,
+				'step' => 0.01,
 				'value' => '',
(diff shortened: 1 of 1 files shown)


Sanfam:
@berrnd , Thanks! I'll do some additional digging and update this PR.

berrnd:
@Sanfam thanks again for your help on this - but this was no done in #421, so I'm going to close this.

PR #470, opened 2019-12-19
Freeze items

### LINKED ISSUE #426: Option to freeze products
Analogue to "Default best before days after opened" - option per product, then an action from the stock overview product conetext menu.

kriddles:
Closes #426 

This allows a location to be set as a "Freezer".

When items are purchased for any location set as a freezer then the best buy date will be automatically set to "never". The best buy date will refresh back to the product default best buy date if a location is selected that is not a freezer.

Future project:
Set best buy date to never when transferring products to a location that is set as a freezer. 
Set the best buy date to best buy date opened or default best buy date when transferring from a location set as a freezer?

### DIFF
--- a/grocy.openapi.json
+++ b/grocy.openapi.json
@@ -3368,6 +3368,9 @@
 					"description": {
 						"type": "string"
 					},
+					"freezer": {
+						"type": "integer"
+					},
 					"row_created_timestamp": {
 						"type": "string",
 						"format": "date-time"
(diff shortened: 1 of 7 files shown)


kriddles:
Fixed merge conflict

berrnd:
This is now done in https://github.com/grocy/grocy/commit/71a57c9dcb1db7c5ca34770fe4a3c6f05214e1cb.

> - New product options "Default best before days after freezing/thawing" to set how the best before date should be changed on freezing/thawing
> - New location option "Is freezer" to indicate if the location is a freezer
> - => When moving a product from/to a freezer location, the best before date is changed accordingly
> - There is also a new sub feature flag `FEATURE_FLAG_STOCK_PRODUCT_FREEZING` to disable this if you don't need it (defaults to `true`)

PR #474, opened 2019-12-20
Remove divisor by purchase factor for recipe costs

### LINKED ISSUE #473: Edit Recipe Questions/Bug?
Heya,

I can't seem to track down what's happening but maybe you know offhand
My issue is selecting an ingredient to edit in a recipe changes the product picker to a different ingredient with a similar name.

1. Select a recipe
2. Click edit on an ingredient in the Ingredient List
3. The product picker for the Edit Recipe Ingredient, does it
- prefill by name?
If so, I have a few ingredients that have similar names.

If instead it uses the product id, then it seems to be sending the wrong product id?

The other issue I have is the Recipe cost looks to be off.
Why is there a divisor in the recipes_pos_resolved by the purchase factor?
I removed it from the view and it seems to have cleared up my issue.

kriddles:
fix #473

### DIFF
--- a/migrations/0096.sql
+++ b/migrations/0096.sql
@@ -1,0 +1,89 @@
+DROP VIEW recipes_pos_resolved;
+CREATE VIEW recipes_pos_resolved
+AS
+
+-- Multiplication by 1.0 to force conversion to float (REAL)
+
+SELECT
+	r.id AS recipe_id,
+	rp.id AS recipe_pos_id,
+	rp.product_id AS product_id,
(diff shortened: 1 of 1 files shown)


berrnd:
As mentioned in https://github.com/grocy/grocy/issues/473#issuecomment-567925191, I think that's ok as it currently is.

> Prices are always per purchase quantity unit.
See "testproduct1" and "testrecipe1" currently on the english demo: https://en.demo.grocy.info/recipes?recipe=11
1 Pack purchased = 10 Pieces in stock; 1 Pack costs 10 $.
If you remove the division as you proposed in #474, it would result that 1 Piece costs 10 $ (instead of 1 $)...

PR #476, opened 2019-12-20
Recipe form updates

### LINKED ISSUE #475: Product picker: Use "prefill-by-id" instead of "prefill-by-name" everywhere
When products have similar names, "prefill-by-name" may picks the wrong product. So the prefilled product should always be selected by its id. Affects at least the recipe ingredient edit page, but all places where the product picker is used should be checked.

References #473.

kriddles:
For the Recipe Ingredient Add and Edit buttons I moved them to instead be bootbox as you have elsewhere.

Edit Ingredient will use the prefill by id. Fixes #475 
I didn't see anywhere else using prefill by name?

Upon success it will currently just reload the page instead of building out a Refresh method.

### DIFF
--- a/public/viewjs/recipeform.js
+++ b/public/viewjs/recipeform.js
@@ -343,7 +343,7 @@ $(window).on("message", function(e)
 
 	if (data.Message === "IngredientsChanged")
 	{
-		window.location.reload(true);
+		window.location.href = U('/recipe/' + Grocy.EditObjectId);
 	}
 
 });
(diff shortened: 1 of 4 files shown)


kriddles:
Thank you!

I just noticed that if you create a new recipe and update the name, then go and add ingredients, since I didn't save the recipe, when it window reloads it sets the recipe name as "New Recipe".

Perhaps when you go to add or update ingredients it should trigger to save recipe to make sure and capture other changes?

Eventually, instead of a window reload the ingredient table should just be refreshed similar to the Stock Overview table.

berrnd:
Yes, that was also the problem before, and that's why a new recipe is immediately created with the name "New recipe" (when adding an ingredient, a `recipe_id` is needed to save the `recipe_pos`. It saved the recipe form when editing or adding an ingredient ([here](https://github.com/grocy/grocy/blob/6c7420ea08ba589d339b380db2f30f153ec2ef12/public/viewjs/recipeform.js#L199), haven't tested if everything still works about that)...

Issue #448, opened 2019-11-26
Convert services to singletons and use lazy loading to improve performance


zebardy:
Performance on my setup with grocy running in a docker container on a ROC-RK3328-CC (https://libre.computer/products/boards/roc-rk3328-cc/) is very slow. The stock page can take upto 25s to fully load, db-changed-time calls would take ~7s. Digging deeper into what was going on I discovered that the DB connection was being created multiple times for a single call, service objects were being created multiple times even if not needed for the logic that would be executed. 

As a proof of concept I've implemented the singleton pattern for some of the most used services and lazy loading. My fork is here (https://github.com/zebardy/grocy/tree/35e88ba5f1d0f7845b6d92301da15c39fabab3d7). These changes have significantly improved performance for me, and should improve performance for others. 

The fork is very much a proof of concept and needs further work before potentially submitting a pull request. For instance it also contains other changes changes done as part of my investigations and testing out using apcu for storing localization data in memory between requests instead of loading from files every time.

What are thoughts on these changes?
Should I put work into preparing and submitting a pull request for this?

berrnd:
Many thanks for your work on this! :)
Every improvement is very welcome, so feel free to prepare a pull request on this.

To be honest, for grocy I personally don't really care about coding-best-practices so far and focus more on the end result. But I also don't run grocy on lowend SoCs. Just as a reference: The public demos (24 instances) run on a Intel Xeon Gold 6140 (2.3 GHz, 2 cores) and have a limit of 512 MB RAM (not per instance, all together) and the performance is pretty good I would say...

BlizzWave:
> What are thoughts on these changes?
> Should I put work into preparing and submitting a pull request for this?

I have complained about speed before.
#120 
So if you have the time to do this it would be appreciated!

berrnd:
Yeah, table loading performance could really see an improvement, especially for the journal pages as mentioned in #120 - but that affects mostly the client, would probably result in (a lot) more (API)requests when done "lazy"...

zebardy:
Coding-best-practices are at the end of the day best advice and tools to help build something and keep it working. There is a lot to be said for building something that runs and is useful. This definitely runs and is useful! Worrying too much about doing things right can often mean nothing gets built at all.

Happy to share back my work. Will probably create a branch of my fork to create a cleaner version of the changes. Might take a bit to find the time to complete this. 

Looking at #120, these changes probably will not solve that issue, but definitely will not hurt.

[...]

PR #479, opened 2019-12-22
Convert services to singletons and use lazy loading to improve performance

zebardy:
This is a first pass at a pull request for changes to turn all services to singletons and for controllers to lazy load services #448. This is a large change and I expect these changes will need further refining.


berrnd:
Thank you a lot for your efforts on this!

Just wanted to try it out, haven't looked into the code changes in detail so far - but I cannot get it to work...

When visiting the root route (`/`):
```
Type: Error
Message: Call to undefined method Grocy\Services\BaseService::MigrateDatabase()
File: C:\Users\bb\Workspace\grocy\controllers\SystemController.php
Line: 20
```

When visiting `/stockoverview`:
```
Type: Error
Message: Call to undefined method Grocy\Services\BaseService::IsValidSession()
File: C:\Users\bb\Workspace\grocy\middleware\SessionAuthMiddleware.php
Line: 38
```

And just about the `get<Service>()` functions in all the controllers: Would it make a performance difference if we just add these functions for alle services to `BaseController` to have them available in all controllers without adding them to each single one seperately? Would make it more clearly, I think...

zebardy:
Fixed the bug I had introduced due to refactoring the singleton pattern to the parent class. Testing locally I'm building a docker container and running from there. However loading for javascript assets is broken for me at the moment, I think this is due to a bug in the docker container I'm building not the code. However it is 2am, so I'm leaving this here for now.

Shouldn't have a performance impact moving the lazy loading functions to the parent controller as a refactoring. However that's a code refactor I'm not going to attempt at ~2am.

zebardy:
Hi, 

managed to find some more time (and rig up a drone CI environment). I've synced up to master, resolved all of the merge conflicts and applied the changes to the code that has changed since I opened this pull request. I've also moved all of the `get<Service>()` functions to the base controller. 

I've built an arm64 docker container with this code and tested it on my setup. It looks good to me. Please let me know if there is anything else you would like me to change before this can be merged.

berrnd:
Thank you a lot for your work.

Just did some tests and some things did not work, starting with the login page. But I hopefully already found and fixed them all. I also removed a lot of unrelated files before merging.

But most important: I did not measure it, but grocy feels "snapier" now, especially API calls. And even on a "normal" machine. Great! :)

zebardy:
Hi,

Sorry I missed a few things, and thankyou for fixing them. I eyeballed the code untill it became blury and tested as much as i could think of, but obviously missed a few usecases! Cant believe I missed checking in the data files.

Thankyou for getting this merged and glad you can see some benefit. Would be interesting to see if it leads to reduced resource usage on the demo server under heavy load.

I also started experimenting with using APCu to cash some state and data structures between invocations. However that would add another dependency and not sure if it would give as significant of an impact. Will let you know if it comes to anything.

Thanks for your work!

Aaron

berrnd:
>  Would be interesting to see if it leads to reduced resource usage on the demo server under heavy load.

The grocy demo is not the only thing running on that server (it's a LXC container), so I don't have specific statistics, but I don't see a real difference in the total resource usage so far (red line is when the pre-release demo was updated after this was merged)...

grafik

grafik

Issue #488, opened 2020-01-08
Little stock API refinements


jglauche:
For a my interface I am writing I need some kind of way to update the stock database directly. My motivation is to update the best_before_date of items without having to remove the item from stock and entering it again.

For my application I hacked it into the api:

```
diff --git a/grocy.openapi.json b/grocy.openapi.json
index eb6a246..199247b 100644
--- a/grocy.openapi.json
+++ b/grocy.openapi.json
@@ -2858,7 +2858,8 @@
                                        "userfields",
                                        "userentities",
                                        "userobjects",
-                                       "meal_plan"
+                                       "meal_plan",
+                                       "stock"
                                ]
                        },
                        "ExposedEntitiesPreventListing": {
```

This is probably not the best solution as this does not touch the stock_log table.

jglauche:
.. I should have checked master branch, this was already merged into.

berrnd:
This was all introduced by the huge amount of changes of #421 (not yet released).

But I think exposing the `stock` table directly was not intentional, and is also not necessary (and maybe dangerous) - there is the new endpoint `PUT /stock` to update stock entries, which then handles journal updates for every change.

We should check/undo that, maybe also renaming to new endpoint to something like `PUT /stock/{entryId}` would be better. I'll reopen this to track that here.

//CC @kriddles

jglauche:
Sounds reasonable. Maybe related, there's an inconsistency with the stock api: When querying 
/stock/products/{productId}/entries
, you get an array with Stock items that have an "id" element. However, the PUT /stock/ api in master requires you to submit "stock_row_id"

kriddles:
Ok I'll take a look at removing the stock access and updating the endpoints.

kriddles:
> Sounds reasonable. Maybe related, there's an inconsistency with the stock api: When querying
> /stock/products/{productId}/entries
> , you get an array with Stock items that have an "id" element. However, the PUT /stock/ api in master requires you to submit "stock_row_id"

I updated the "stock_row_id" to be named "id" for the put requirement

berrnd:
Done by #493.

PR #493, opened 2020-01-15
Stock detail updates

kriddles:
#488 fix

Take a look at the removal of stock and feel free to make further updates as desired.

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -580,6 +580,11 @@ class StockApiController extends BaseApiController
 		return $this->ApiResponse($this->StockService->GetProductStockLocations($args['productId']));
 	}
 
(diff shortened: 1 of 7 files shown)


kriddles:
It looks like the Undo was calling the wrong function for the stock detail refresh.

Also, an issue I see but don't quite know where to look to fix...
I was playing around and set a very large stock detail row of over 12,212. After I edit another stock detail row then it looks to format this amount row to be 12 instead of showing 12,212.

example:
Cold cuts one of them set to 121,228
image
I edit the first 1pack to instead be a 3 pack.
Now the 121,228 changed to be 121 which is just the display in the table.
image

berrnd:
Thanks.

...and I cannot reproduce the problem that large numbers get cut off...

PR #494, opened 2020-01-16
productcard.js check null location

### LINKED ISSUE #489: Bug: productcard.js throws when product properties are missing because of disabled FEATURE flags
With disabled location tracking FEATURE flag, loading the productcard info fails because the missing location is `null` and therefore has not property `name`:

```text
productcard.js?v=2.5.2:18 Uncaught TypeError: Cannot read property 'name' of null
    at productcard.js?v=2.5.2:18
    at XMLHttpRequest.xhr.onreadystatechange (grocy.js?v=2.5.2:17)
```

image

This leads to the rest of the `productcard.js` to not run and breaks the buttons and all info that comes below that line.

## Reproduction
- Run Grocy release 2.5.2 with `Setting('FEATURE_FLAG_STOCK_LOCATION_TRACKING', false);`
- Create product
- Open `/inventory`
- select your product

## Expected Behaviour
- Productcard shows existing information about the product and leaves out what's missing
- Buttons in the productcard window work

## Actual Behaviour
- Productcard shows part of the information (all collected infos above the erroring line)
- Buttons in productcard remain disabled

Probably we can just wrap the relevant part with a `if( productDetails.location ) { ... }` block for each of the properties, as I found no way to use a *safe navigation* or *elvis* operator implementation in Javascript.

kriddles:
fix #489

### DIFF
--- a/public/viewjs/components/productcard.js
+++ b/public/viewjs/components/productcard.js
@@ -15,7 +15,10 @@ Grocy.Components.ProductCard.Refresh = function(productId)
 			$('#productcard-product-last-purchased-timeago').attr("datetime", productDetails.last_purchased || '2999-12-31');
 			$('#productcard-product-last-used').text((productDetails.last_used || '2999-12-31').substring(0, 10));
 			$('#productcard-product-last-used-timeago').attr("datetime", productDetails.last_used || '2999-12-31');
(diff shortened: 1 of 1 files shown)


berrnd:
Great, thank you! :)

Issue #487, opened 2020-01-08
Feature request : api/stock can return detailed products


BlueBlueBlob:
Hi :)

Can you add an argument to api/stock to get products with details ?
To avoid running api/product/{product_id} on every products.

Please :)

PR #496, opened 2020-01-16
Feature request : api/stock can return detailed products #487

gsacre:
Added a new API: /api/stock/full

This API returns the same values as "/api/stock" except that it also
gives the details of the corresponding product.

This also answers the request from: https://www.reddit.com/r/grocy/comments/enu427/api_retrieve_all_products_from_given_product_group/

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -388,11 +388,6 @@ class StockApiController extends BaseApiController
 		return $this->ApiResponse($this->StockService->GetCurrentStock());
 	}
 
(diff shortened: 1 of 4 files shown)


gsacre:
Hi berrnd,

I think that keeping '/stock' untouched would be the best approach to be backward compatible.
Your proposal to use a parameter to '/stock' makes perfect sense, but just to be sure, at the end, the URL would be: .../api/stock?product=show

Is this what you expect?

berrnd:
> Is this what you expect?

Yes, and at the end it's just a new property/field in the JSON response, show should not hurt, we don't change anything existing by that approach...

Maybe also just include the product object always, without an optional query parameter...

Issue #374, opened 2019-09-23
Tie a Recipe to a Product ("Self produced products")


kriddles:
Just an idea, not fully thought out... 

Can consuming a recipe also be able to increase a product qty tied to the recipe?

For example, if you make your own ranch sauce or gyro meat for use in other recipes.

The product then could also be checked if used as a nested recipe.

berrnd:
Sorry, maybe don't understand that...

So when your recipe needs 2 eggs and you use the "consume this recipe" button, the recipe ingredient amount should be increased (to 4) on that action?

blubbel42:
I Think its more like:
My recipe uses 3 carrots, 5 potatos and 1 kohlrabi, so they are consumed from stock.
It produces 6 portions of vegetable-soup. These will be frozen and added to stock.
so the amount of individual vegetables ist decreased but the amount of soup-portions is increased.
works the like for bread etc., there are even recipes for shampoos..
Especially useful for homemade stuff.

berrnd:
Ah, ok, got it now, great idea!

So kind of "self produced products"...

kriddles:
Thanks!

Yes, I would then use that self produced product which is tied to a recipe in another recipe as an ingredient instead of a nested recipe. That way I can make a 'batch' of whatever the product is to be used across multiple recipes in a week.

A useful part would be making it a clickable link that opens the recipe full screen for that product in the recipe screen.
Or if in the new Meal Plan Recipe Modal, switch to that Product recipe

berrnd:
And just thinking: Even the price of the self produced product should be possible to track automatically, as we already have the price what 1 serving costs based on the product purchase prices...

berrnd:
This has now landed, done by #501.

PR #501, opened 2020-01-17
recipes can create products

kriddles:
#374 
Initial work for self producing products.

A product can be assigned to a recipe. On recipe consumption the product is created using the price of the recipe along with desired servings.

An idea is perhaps on recipe consumption a checkbox could toggle the creation of the product?

### DIFF
--- a/migrations/0095.sql
+++ b/migrations/0095.sql
@@ -20,6 +20,25 @@ ADD correlation_id TEXT;
 ALTER TABLE stock_log
 ADD transaction_id TEXT;
 
+ALTER TABLE recipes
+ADD product_id INTEGER;
+
+DROP VIEW recipes_resolved;
+CREATE VIEW recipes_resolved
+AS
+SELECT
+	1 AS id, -- Dummy, LessQL needs an id column
+	r.id AS recipe_id,
+	IFNULL(MIN(rpr.need_fulfilled), 1) AS need_fulfilled,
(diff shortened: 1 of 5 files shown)


kriddles:
Also need to update the success toaster

berrnd:
Cool, thanks for all your work! :)

Will have a detailed look later or tomorrow, but looks good so far I think.

> An idea is perhaps on recipe consumption a checkbox could toggle the creation of the product?

We already have a "do you really want"-popup when consuming a recipe, so maybe a button there to trigger the stock transaction. (?) Or maybe it's enough when the recipe has a product linked, when it has one, always book the transaction on consume...

And maybe a separate stock transaction type like `TRANSACTION_TYPE_SELF_PRODUCTION` would be good.

kriddles:
Yeah I agree there should be a separate type.

zebardy:
🧠💪👨‍💻🙌👍

necr0dncr:
Hello! Hate to revive an older thread, but I was unsure if this should go here or be brought up as another potential feature.

I really love this aspect of the recipe and product management. But is there a way for the recipe to generate the product automatically?
As of right now I am assuming a product must be manually created, and then linked to the recipe with the "Produced Product".

berrnd:
> Hate to revive an older thread

Hate that too, thanks for spamming ([README ref](https://github.com/grocy/grocy?tab=readme-ov-file#questions--help--bug-reports--feature-requests)).

&nbsp;
> As of right now I am assuming a product must be manually created, and then linked to the recipe with the "Produced Product".

There are no hidden gems anywhere, so I don't see how that should be possible as of today.

PR #503, opened 2020-01-18
Feature request : api/stock can return detailed products #487

### LINKED ISSUE #487: Feature request : api/stock can return detailed products

gsacre:
The response of the call '/api/stock' now returns a new attribute
('product') which contains the details of the related product.

Created new pull request after discussion with previous PR to keep things clean.

Closes #487.

### DIFF
--- a/grocy.openapi.json
+++ b/grocy.openapi.json
@@ -163,7 +163,7 @@
 											{
 												"$ref": "#/components/schemas/StockEntry"
 											}
-										]	
+										]
 									}
 								}
 							}
@@ -626,7 +626,7 @@
 					}
 				}
 			}
-		},		
+		},
 		"/files/{group}/{fileName}": {
 			"get": {
 				"summary": "Serves the given file",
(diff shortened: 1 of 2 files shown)


berrnd:
Thank you, works perfectly, but I tried to simplify it a little bit, see https://github.com/grocy/grocy/commit/d0036e8034fd3daf208894bfde5151e786250a73.

https://github.com/grocy/grocy/blob/d0036e8034fd3daf208894bfde5151e786250a73/services/StockService.php#L29-L38

gsacre:
Thank you!

Just for my understanding, and as I'm not that familiar with this framework, don't you execute twice the $sql query on the stock_current table (lines 29 and 31)?

I'm asking also so that I can better help next time :-)

berrnd:
Line 31 queries all products which are part of the query before by using it as a subquery (so yes, it's executed twice) - just thought that it's cheaper to do it that way as one foreach-loop less is necassary, but haven't measured it...

`$this->Database` is an instance of the used ORM LessQL, see here :https://github.com/morris/lessql/blob/master/doc/api.md

So `$this->Database->products()` is basically the query `SELECT * FROM products`....

gsacre:
Thanks for the explanation!

Issue #506, opened 2020-01-21
TODOs before releasing stock edit & transfer products feature (#421)


berrnd:
I think it would be good to have a new release soon, did some more tests and fixed a couple of things regarding #421, the following things need to be done before releasing:

- [x] /stockdetail: Consume and undo from the success toast: The row is not shown again (because it has a different stock row id and refreshing is done based on that id)
- [x] /stockedit: Make the purchased date editable (problem about 2 datetimepickers on the same page)
- [x] New stock entry API endpoints: I would prefer to have ~~`GET /stock/{entryId}` and `PUT /stock/{entryId}` (route scheme colides with `/stock/valatile`, needs to be handled somehow)~~
  - [x] Or better/more restful `GET /stock/entry/{entryId}` and `PUT /stock/entry/{entryId}`

//CC @kriddles

PR #507, opened 2020-01-22
421

kriddles:
#421 {excited for the release!}

Some updates for the release:

updated get/put endpoint to be stock/entry/{entryId}

I added the stock edit back to the stockdetail dropdown

For the recipeform I find it useful to have the product workflow so that I can make a recipe and if needed create a product from that interface rather than have to create a product and then create a recipe.

Without the on document ready it seems that some of the filters do not work such as on the stockdetail and stockedit

With your cleanup I am not sure but the stock edit is not showing your new purchase date and is not saving the changes when attempted to save?

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -124,11 +124,6 @@ class StockApiController extends BaseApiController
 				throw new \Exception('Request body could not be parsed (probably invalid JSON format or missing/wrong Content-Type header)');
 			}
 
-			if (!array_key_exists('id', $requestBody))
-			{
-				throw new \Exception('A stock row id is required');
-			}
-
 			if (!array_key_exists('amount', $requestBody))
 			{
 				throw new \Exception('An amount is required');
@@ -152,7 +147,7 @@ class StockApiController extends BaseApiController
 				$locationId = $requestBody['location_id'];
 			}
 
-			$bookingId = $this->StockService->EditStockEntry($requestBody['id'], $requestBody['amount'], $bestBeforeDate, $locationId, $price, $requestBody['open']);
+			$bookingId = $this->StockService->EditStockEntry($args['entryId'], $requestBody['amount'], $bestBeforeDate, $locationId, $price, $requestBody['open']);
 			return $this->ApiResponse($this->Database->stock_log($bookingId));
 		}
 		catch (\Exception $ex)
--- a/grocy.openapi.json
+++ b/grocy.openapi.json
@@ -1050,86 +1050,9 @@
 						}
 					}
 				}
-			},
-			"put": {
-				"summary": "Edits the stock entry",
-				"tags": [
-					"Stock"
-				],
-				"requestBody": {
-					"required": true,
-					"content": {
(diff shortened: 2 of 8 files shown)


berrnd:
Thanks.

> I added the stock edit back to the stockdetail dropdown

Ok, I made a button in the first row to edit the entry, I think this should be more prominent on that page is it's may only about editing there ... so really needed as button and in the drodpown menu?

> For the recipeform I find it useful to have the product workflow so that I can make a recipe and if needed create a product from that interface rather than have to create a product and then create a recipe.

Also ok, but then we have to test that no recipe changes get lost when leaving the page (save the recipe before).

> Without the on document ready it seems that some of the filters do not work such as on the stockdetail and stockedit

Just put this at the right place (end) of the script, after everything else is initialized...

> With your cleanup I am not sure but the stock edit is not showing your new purchase date and is not saving the changes when attempted to save?

The field is commented out currently on the stockedit page, as mentioned in #506, there is the problem when having 2 datetimepickers on the same page (the same problem as for all components, I initially thought we would never need two on the same page (which is of course stupid...).

kriddles:
I updated with some fixes.

I didn't notice you moved the Stock Edit so I removed it from the dropdown. It's better where you put it, I just have to get used to it there now:)

A few bug fixes, the stock edit on the detail page was in the if statement for the open feature check.
I updated the stock service to better handle Undo for the open when editing stock.
I moved my stock detail to be further down in the script instead of using document ready(thanks!)

I've not thoroughly tested whether the changes are saved for recipes if you go through the create product workflow yet.

Question: Can we keep the Consume All for the stock detail page next to the Consume 1? That button was useful to me.

berrnd:
Great, thanks! :)

> Question: Can we keep the Consume All for the stock detail page next to the Consume 1? That button was useful to me.

Don't know, just wanted to focus the "edit action" on that page. The "raw" stock lines shown there can be split up into multiple different lines, especially when using partial amounts. So I think exactly `1` is more or less arbitrary there...

berrnd:
> I've not thoroughly tested whether the changes are saved for recipes if you go through the create product workflow yet.

This works now, I added to just save the recipe on every product picker change.

Also updated the list in #506, two (hopefully) little points left. :)
So let's say to finish that and maybe some other little things this weekend and push a new release next friday or saturday next week...

kriddles:
> So let's say to finish that and maybe some other little things this weekend and push a new release on friday or saturday next week...

Just a heads up, I will not be available over this coming weekend to work on code updates but I'll keep an eye out to review and test next week.

berrnd:
Absolutely no problem, thanks for all your help. :)

I finished the two open points in #506 by more or less dirty workarounds (but as always - get it to work first, improve it later... :D):
- Just copied the datetimepicker to "datetimepicker2" and used that for the purchased date on the /stockedit page
- Just reload the /stockedit page when after consume/undone the stock row is not found by the new id

So from that side the release is ready. Will also do some more tests and maybe take care about the other open `bug` and `ui-bug` issues or some other little things...

Issue #450, opened 2019-11-26
Ability to add individual products as well as recipes to meal plans


zebardy:
As a user of grocy I want to add products as well as recipes to the meal plan so that I can plan out food for the week for planning and shopping purposes.

If #374 was implemented, then just being able to add products (recipes would be self produced products) would resolve this.

zebardy:
🥳🙌👍

Issue #477, opened 2019-12-21
Feature request: Free input field for mealplan


Jeltel:
When we go out (to family or a restaurant) the meal for a certain doesn't need to be filled in. But it is handy to show something there, so that it is clear we do not need to think of something to eat that day.
Just like the note field when adding an item to the shopping list. When you only fill in the note, that gets added to the shopping list.

kriddles:
Hi Jeltel, could you create a recipe called "Eating Out" for now to use?

To expand this as a feature, maybe there could be another selection for Eating Out which could then provide an optional picker list of restaurants to select from which could be either added to the locations table with a new field tagging it or a separate object.

Or even better, an integration with Yelp? It looks like there is an open yelp REST api.

Jeltel:
@kriddles That is an option. But more complicated than I need.

In your solution you indeed cover the restaurants, but for family I need to make a list for every option. And I just want to fill in where I'm going to eat.  Just as a placeholder, so that I know I don't have to think of something that day and don't have to do any shopping.

For now I have a 'pizza' meal without any ingredients. But for the other things now, when I have the mealplan open, I would have to go back to the recipes. Create a 'recipe' like Eating at my Mom, go back to mealplan and select the recipe there. 
I would just want to fill in the title 'eating at my mom' one time and go on with the other days. Also with all empty recipe, the recipe section becomes cluttered.

berrnd:
That's now possible: New option in the dropdown of the add button per day:

![grafik](https://user-images.githubusercontent.com/2161815/73019209-6fe0c380-3e23-11ea-9655-2a87c455c36e.png)

Maybe have a short test on the pre-release demo if this would work for you: https://demo-prerelease.grocy.info/mealplan

PR #508, opened 2020-01-22
product in mealplan

kriddles:
#450 

Now that recipes can create products then adding products to meal plans will be useful and would be nice to have for the release.

This is a work in progress commit not quite up to master just to show the work if you want to take it a bit further. It's almost there.

It may be useful to add a new product db column called ?"include_in_mealplan"? that way products can be limited in the mealplan dropdown. This would be a checkbox option on the product creation/edit pages.

### DIFF
--- a/controllers/RecipesController.php
+++ b/controllers/RecipesController.php
@@ -132,6 +132,7 @@ class RecipesController extends BaseController
 	public function MealPlan(\Slim\Http\Request $request, \Slim\Http\Response $response, array $args)
 	{
 		$recipes = $this->Database->recipes()->where('type', RecipesService::RECIPE_TYPE_NORMAL)->fetchAll();
+		$products = $this->Database->products()->orderBy('name');
 		
 		$events = array();
 		foreach($this->Database->meal_plan() as $mealPlanEntry)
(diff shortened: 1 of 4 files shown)


berrnd:
Haven't thought about this in detail, but as the meal plan is based on (internal) nested recipes (per day and week), it should be only necessary to include the product in the day recipe (maybe also the week recipe, don't know right now if the week recipe is just a recipe which includes all day recipes)...

Also the `recipe_id` column of the `meal_plan` table does not allow `NULL` values currently, need to be changed then, but also need to be checked that then either `recipe_id` or `product_id` is set...

Or maybe just create a new (internal) recipe (same name as the product, at least for displaying) for those case and don't change anything else (maybe easier)...

kriddles:
Definitely do not merge this at the moment; it's still set up just for dev since I updated an earlier migration. Just a little more tinkering on it.

I removed the constraints on the meal plan table however it looks like in order to actually remove them we'll have to create a new table, copy the data over, delete the original, rename the new table.

berrnd:
> I removed the constraints on the meal plan table however it looks like in order to actually remove them we'll have to create a new table, copy the data over, delete the original, rename the new table.

Yes, that's unfortunately the only way in SQLite as altering or removing columns is not possible...

See for example [this migration](https://github.com/grocy/grocy/blob/dfc05e0bec2c524dc2f35a833f8fff6435f63a46/migrations/0048.sql) where the `shopping_list` table was "edited".
Important is also `PRAGMA legacy_alter_table`, as since a maybe two year old version of SQLite all references of a table (in views and triggers) get renamed when renaming a table...

berrnd:
I made some changes for #477 which also may help here:
- The `meal_plan` table now has a column `type` (currently to distinguish between recipes and notes, can also be used for products)
- The `recipe_id` column is not longer a `NOT NULL` column

kriddles:
I updated my work in progress for allowing products for meal plans based on your changes that helped.

Take a look when you get a moment. I won't be able to take it much further this weekend.
It currently allows adding a product to the meal plan, deleting the product once added, and consuming the product with Undo working.

My last commit attempts to clear the product or recipe picker but for some reason it is not actually clearing the values so both product_id and recipe_id are being set in the database? This will cause an error on the page. An error will not occur if you open the modal and only select a product and save.

berrnd:
I will have a look, but maybe just a separate modal would be easier, like now for the note a new link in the "add button dropdown"...

kriddles:
That sounds great and much simpler

berrnd:
I finished this now in https://github.com/grocy/grocy/commit/e111d07f4e381ec064abb19b5e227c0efdd99a19.

PR #527, opened 2020-01-29
389 Display ingredient amount as fractions

kriddles:
I was just playing around with a few of these functions found from https://stackoverflow.com/questions/59727517/how-to-convert-between-decimals-and-fractions-in-php

It did require me to enable php-gmp on my system to work.

I then played with an example in the views/recipes.blade.php changing the selectedRecipes Ingredients recipe_amount to instead be {{ decimalToFraction($selectedRecipePosition->recipe_amount) }}

In the demo this was now displayed 
/recipes?recipe=5

image

### DIFF
--- a/config-dist.php
+++ b/config-dist.php
@@ -101,6 +101,7 @@ DefaultUserSetting('shopping_list_disable_auto_compact_view_on_mobile', false);
 
 # Recipe settings
 DefaultUserSetting('recipe_ingredients_group_by_product_group', false); // Group recipe ingredients by their product group
+DefaultUserSetting('recipe_ingredient_display_as_fractions', false); // Display the ingredient amounts as fractions
 
 # Chores settings
 DefaultUserSetting('chores_due_soon_days', 5);
--- a/helpers/extensions.php
+++ b/helpers/extensions.php
@@ -236,3 +236,62 @@ function string_ends_with($haystack, $needle)
 
 	return (substr($haystack, -$length) === $needle);
 }
+
+function FractionToDecimal($fraction)
+{
+	// Split fraction into whole number and fraction components
+	preg_match('/^(?P<whole>\d+)?\s?((?P<numerator>\d+)\/(?P<denominator>\d+))?$/', $fraction, $components);
+
+	// Extract whole number, numerator, and denominator components
+	$whole = $components['whole'] ?: 0;
+	$numerator = $components['numerator'] ?: 0;
+	$denominator = $components['denominator'] ?: 0;
+
+	// Create decimal value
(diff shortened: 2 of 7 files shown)


kriddles:
This would allow the db to remain unchanged and continue to store as decimals.

A variable ?"displayAsFraction"? could be checked whether or not to display the decimalToFraction conversion for ingredients or even by product?

It would be nice to be able to add an ingredient to a recipe using a string such as "1/4" which would then fractionToDecimal to be stored in the db. This could be allowed if the variable "displayAsFraction" were at the product level?

berrnd:
Cool, maybe it would work to only change how the number is displayed and keep decimal numbers internally. But let's test this first with a `1/3` amount - this is most of the time the problem which leads to that there is `0.00...01` of something left in stock at some point...

kriddles:
I rewrote part of the decimalToFraction to better handle displaying 1/3 using a tolerance and precision.

I also have an example update to the consume so that a stock will be consumed if left with less than .01 remaining in the stock amount. Perhaps the .01 should be more? .0001?

kriddles:
Using the DecimalToFraction function is relatively easy to show the user fractions, however I am not having much luck on adding to the RecipePosForm the ability to let the user input a fraction and then use the FractionToDecimal function to store the decimal value.

Is there an example somewhere of using the php functions in the js? Or how would you go about this?

berrnd:
Thanks for trying this, I will have no time to look into this before sunday or so (but will push the new release tomorrow - seems to be everything new/included as of now is working... :)).

kriddles:
No rush on it!

Thanks for getting a release ready

kriddles:
This is rebased off of #556 so that the display as fractions can be included in the new recipes settings.

Currently this only allows switching to displaying the ingredient amount as fractions.

I have not attempted to allow the user to input as fractions. I would be ok with having a future enhancement to allow entering as fractions. 
I played around with eval() and it would seem to work to take the fraction string and convert to decimal. However right now entering as a fraction is not a big priority. There seem to be very few needs to enter 1/3 as an amount so most ingredient adds can be in decimal with viewing in fraction.

berrnd:
> There seem to be very few needs to enter 1/3 as an amount so most ingredient adds can be in decimal with viewing in fraction.

Really? Sorry, still haven't tested it myself yet: But then I have to enter something like `0.33333` when I want to have `1/3`? I think also entering fractions would be needed, otherwise I don't know if anyone would find that that much useful...

So we need the convert functions on the frontend (JS) and also need to solve the input validation to only allow proper fractions...

berrnd:

PR #528, opened 2020-01-30
Limit mealplan products dropdown based on include_in_mealplan product…

kriddles:
… variable

New product variable called "include_in_mealplan" that adds a checkbox on product create/edit.
This is used in the mealplan to limit the product dropdown to just desired products.


berrnd:
Don't really know if we need this, we already have a lot product options - we would then also need "show in recipes", "show in chores" and so on...

Searching works ok I think, maybe highlighting self produced products or list them first would be a thing, as I think for the meal plan it will be mostly about them...

kriddles:
Searching does work well.

 I like the idea of some sort of highlight or maybe an asterisk next to products that can be self produced.

We'll see if the community has suggestions after using the new feature for awhile.

Closing this for now

kriddles:
Another quick idea for future reference...

If we would need to add for products the include in recipes and include in chores, etc. Then it could be hierarchy of set at product_group which could be overridden at the product level.

I'd imagine a dropdown multselect option would work well for this

PR #539, opened 2020-02-02
Fix horizontal scrolling on compact shopping list

gersilex:
I could not find any negative side-effects


berrnd:
Next try is https://github.com/grocy/grocy/commit/7235d9c0c9395bb9b827fece068debb19b9a7fe9.

The page fits again, so no horizontal scrollbar, but it leaves the default padding on the left and right side - but ok in my opinion...

gersilex:
Absolutely fine right now. Maybe I'll get back to it later again. With proper testing then. Thanks!

PR #541, opened 2020-02-02
Use full width for compact shopping list

gersilex:
I had luck with not enforcing any width, not even 100% and removed the
padding and margin overrides from the shoppinglist.js

I also tested it on a real Android device to be really sure.


berrnd:
Sorry, seems that the recipe fullscreen card was already broken before, don't know since when ... will create a separate issue for that...

PR #545, opened 2020-02-04
Mealplan recipe servings resolution

kriddles:
It does not appear that the mealplan recipe_servings is used when adding ingredients to the shopping list or consuming ingredients.
The recipe desired_servings looks to resolve a lot of information that is used when adding ingredients to the shopping list or consuming a recipe ingredients. 
The first commit sets the recipe desired_servings to be the mealplan recipe_servings so that the correct information is resolved.

I fixed the selectedRecipePostion but I did not look elsewhere in the recipe blade for other possible bugs such as if there is a sub recipe.

### DIFF
--- a/public/viewjs/mealplan.js
+++ b/public/viewjs/mealplan.js
@@ -135,9 +135,9 @@ var calendar = $("#calendar").fullCalendar({
 					<h5> \
 						<a class="ml-1 btn btn-outline-danger btn-xs remove-recipe-button" href="#"><i class="fas fa-trash"></i></a> \
 						<a class="ml-1 btn btn-outline-info btn-xs edit-meal-plan-entry-button" href="#"><i class="fas fa-edit"></i></a> \
-						<a class="ml-1 btn btn-outline-primary btn-xs recipe-order-missing-button ' + recipeOrderMissingButtonDisabledClasses + '" href="#" data-toggle="tooltip" title="' + __t("Put missing products on shopping list") + '" data-recipe-id="' + recipe.id.toString() + '" data-recipe-name="' + recipe.name + '" data-recipe-type="' + recipe.type + '"><i class="fas fa-cart-plus"></i></a> \
-						<a class="ml-1 btn btn-outline-success btn-xs recipe-consume-button ' + recipeConsumeButtonDisabledClasses + '" href="#" data-toggle="tooltip" title="' + __t("Consume all ingredients needed by this recipe") + '" data-recipe-id="' + recipe.id.toString() + '" data-recipe-name="' + recipe.name + '" data-recipe-type="' + recipe.type + '"><i class="fas fa-utensils"></i></a> \
-						<a class="ml-1 btn btn-outline-secondary btn-xs recipe-popup-button" href="#" data-toggle="tooltip" title="' + __t("Display recipe") + '" data-recipe-id="' + recipe.id.toString() + '" data-recipe-name="' + recipe.name + '" data-recipe-type="' + recipe.type + '"><i class="fas fa-eye"></i></a> \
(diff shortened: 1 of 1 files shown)


berrnd:
Haven't tried to reproduce the problem yet, but all week/day recipes are created by a trigger on `INSERT` into the `meal_plan` table, they are never created elsewhere manually via the API - so I think this should be addressed there, last time edited in migration [0096](https://github.com/grocy/grocy/blob/fcfe62f4d3ff097e2fda67815cb669e51d11d886/migrations/0096.sql#L25-L162).

About the other problem:
I think `$selectedRecipePosition` can never have a property `need_fulfilled`, as it*s a `recipes_pos` row. That property is from a `recipes_pos_resolved` row and that rows are hold in `$recipePositionsResolved` in that view...

kriddles:
If using the week/day recipe information then there could be issues if the same recipe is included on the same day. Updating the current recipe serving_amount to resolve the desired servings looks to be the only way currently to make sure the accurate product amount is consumed or added to the shopping list.

I also updated the Recipe Overview commit based on your recent changes. It appears that the recipePositionsResolved is looking at the day or week recipe instead of the base recipe. I remedied this by only including recipe_id > 0

kriddles:
The recipe bugs can be seen in the prerelease demo.

1. In shopping list, purchase 1 qty each of sieved tomatoes and minced meat as is already populated.
2. Go to the Meal Plan page.
3. Click the Spagetti Bolognese Display Button
4. Change the serving size amount to be 3.
5. Notice the helper text "missing" does not reflect changing the serving size of the recipe
6. Close the popup
7. Click the Consume Recipe button - this will fail with a yellow popup saying not enough in stock. This is due to the Consume attempting to resolve against 3 servings instead of the desired meal plan 1 serving since we changed the recipe desired_serving in step 4.

berrnd:
Ok, thanks for the description, so it's only about the serving size which you expected to change when the serving size of the recipe changed. But that's set when adding a recipe to the meal plan, saved in the column `meal_plan.recipe_servings` and not changed again automatically. So maybe we just need an edit button for the meal plan entry. Don't know if it's good to couple that with `desired_servings` of the recipe, so you could never have a different serving size (on any day) for that recipe on the meal plan...

And sorry - I think I was wrong about
> I think `$selectedRecipePosition` can never have a property `need_fulfilled`, as it*s a `recipes_pos` row. 

It's a `recipes_pos_resolved` row at that place...

berrnd:
> I remedied this by only including recipe_id > 0

Better do this by the `type` column of the `recipes` table...

kriddles:
> > I remedied this by only including recipe_id > 0
> 
> Better do this by the `type` column of the `recipes` table...

They type column is not available for that view. Would you want it added?

kriddles:
> Ok, thanks for the description, so it's only about the serving size which you expected to change when the serving size of the recipe changed. But that's set when adding a recipe to the meal plan, saved in the column `meal_plan.recipe_servings` and not changed again automatically. So maybe we just need an edit button for the meal plan entry. Don't know if it's good to couple that with `desired_servings` of the recipe, so you could never have a different serving size (on any day) for that recipe on the meal plan...
>

I don't mean to edit the meal_plan.recipe_servings right now. I think it's fine right now to delete the recipe and add it again it if necessary with a different recipe_servings.

Let me detail out each scenario because I think I did not explain the issue and what I would expect to occur.

berrnd:
> They type column is not available for that view. Would you want it added?

If it's necessary for a where clause or something, no problem that it get added where needed... :)

And ok, but maybe an edit button for all entry-types would not hurt and should be easy to add - saves 1 one page reload compared to delete/add again...

berrnd:
I've added an edit button to be able to edit meal plan entries.

Maybe it's better to move the other changes about things mentioned above to a separate pull request?

kriddles:
> I've added an edit button to be able to edit meal plan entries.
> 
> Maybe it's better to move the other changes about things mentioned above to a separate pull request?

Sounds good, I'll take a look at splitting them in a few hours.

kriddles:
This pull request is now just to fix the mealplan serving size recipe position resolved issue.

berrnd:
Just as a side note: `Grocy.Api.*` functions are executed async, so when doing two dependent actions it's not ensured that they get executed in the same order/timing - the success/error handlers need to be used (I already changed that in https://github.com/grocy/grocy/commit/49edd011b45f694b6294fd0ee1d98e8d837bd760).

PR #552, opened 2020-02-09
CalendarService reflect config options for GetEvents

kriddles:
Reflect flags in config option for calender service getevents

### DIFF
--- a/services/CalendarService.php
+++ b/services/CalendarService.php
@@ -29,65 +29,77 @@ class CalendarService extends BaseService
 
 	public function GetEvents()
 	{
-		$products = $this->Database->products();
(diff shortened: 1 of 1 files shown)


berrnd:
Thanks.

To make it complete I also added to only include meal plan entries when `FEATURE_FLAG_RECIPES` is enabled (meal plan is currently part of that flag).

But should also make only a difference when a feature flag was disabled but the feature was used before (when `FEATURE_FLAG_TASKS` is disabled, there should be no tasks). Except for expiring products, the best before date is set to today when `FEATURE_FLAG_STOCK_BEST_BEFORE_DATE_TRACKING` is disabled.

kriddles:
Thanks!

PR #555, opened 2020-02-09
View base servings on recipecard

kriddles:
Include a hint on the recipeCard to see the base servings.


berrnd:
Just tried to make the hint for the desired serving amount picker a little smaller to prevent a new line on most screen sizes to keep a line with "Energy" and "Costs"...

PR #556, opened 2020-02-09
Include product_group in recipe ingredient list

kriddles:
This will now include a section in the recipe card ingredient list to separate by product group.

If there is already entries for the ingredient group then it will include that feature.

image

I'm not sure if you want to make this a feature or not?


berrnd:
Hm, product groups were originally introduced to group products by "supermarket assortments", so mainly for the shopping list. Of course maybe and depending how the groups are used that would also be useful for recipe ingredients. But if that's done by default now, all recipes would display differently now...

So I would prefer to make that optional. (?)
How about a new "Recipe settings" page (top right corner settings menu, like the others), and there an option like "Group recipe ingredients by product groups" (default `false`)?

kriddles:
> So I would prefer to make that optional. (?)
> How about a new "Recipe settings" page (top right corner settings menu, like the others), and there an option like "Group recipe ingredients by product groups" (default `false`)?

I understand, yes adding the recipe settings would be great!

kriddles:
Created recipe settings page for this feature.

I just added some basic information for display which you may want to update to make it easier to understand the setting.

PR #557, opened 2020-02-09
Recipe Overview recipePositionsResolved should only include normal re…

kriddles:
…cipes

Bug: It appears that the recipePositionsResolved is looking at the day or week recipe instead of the base recipe. I remedied this by only including normal recipes.

### DIFF
--- a/controllers/RecipesController.php
+++ b/controllers/RecipesController.php
@@ -62,7 +62,7 @@ class RecipesController extends BaseController
 		return $this->AppContainer->view->render($response, 'recipes', [
 			'recipes' => $recipes,
 			'recipesResolved' => $recipesResolved,
-			'recipePositionsResolved' => $this->Database->recipes_pos_resolved(),
+			'recipePositionsResolved' => $this->Database->recipes_pos_resolved()->where('recipe_type', RecipesService::RECIPE_TYPE_NORMAL),
 			'selectedRecipe' => $selectedRecipe,
 			'selectedRecipePositionsResolved' => $selectedRecipePositionsResolved,
 			'products' => $this->Database->products(),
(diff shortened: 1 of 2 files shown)


berrnd:
When is this a problem?

controllers/RecipesController.php
```diff
@@ -62,7 +62,7 @@ public function Overview(\Slim\Http\Request $request, \Slim\Http\Response $respo
 		return $this->AppContainer->view->render($response, 'recipes', [
 			'recipes' => $recipes,
 			'recipesResolved' => $recipesResolved,
-			'recipePositionsResolved' => $this->Database->recipes_pos_resolved(),
+			'recipePositionsResolved' => $this->Database->recipes_pos_resolved()->where('recipe_type', RecipesService::RECIPE_TYPE_NORMAL),
```
berrnd: Needs to be done based on the query parameter `include-internal` like above in line 25 - is only for dev/debugging purposes, but when supplied, also internal recipes (so all types) will be shown on the recipes page...

kriddles:
> When is this a problem?

This is used by the ingredient missing check on the recipe card. If the recipe is used in a mealplan then the array to look for the missing information in recipePositionsResolved will use the mealplan information since it's a negative id and shows up first in order.

If you go to the demo prerelease pizza recipe and increase the desired servings. You'll see that the missing tomatoes does not increase or decrease.

image

berrnd:
Ok, thanks for the explanation/example.

So maybe we could also remove the thing about the `include-internal` query param to not do too much differently in the controller just for that - for debugging it's also ok to look into the database I think...

kriddles:
> Ok, thanks for the explanation/example.
> 
> So maybe we could also remove the thing about the `include-internal` query param to not do too much differently in the controller just for that - for debugging it's also ok to look into the database I think...

Ok, I'll leave it up to you for the remaining changes needed.

Issue #553, opened 2020-02-09
Feature Request: Meal plan rules based on rules


Crupuk:
Hi,

Just discovered this project on Reddit and I have one question : 

Do you plan to have meal plan's rules ?
And generating meal plan with this rules ( based on last time used ) 

Like : 
At least one soup 
Max 1time meat
Min 1time fish 
At least one time vegan
... 

Thanks for this project i will have a deeper look next week.

berrnd:
Sounds like #310, feel free to add further ideas about this there.

Crupuk:
Thanks for replying.

I'm not sur that's the same subject.

The issue #310  talk about a meal plan that you can repeat after x week.

You only need to prepare 3/4 week to have a full month planned.

I was talking about a motor to create a meal plan from nothing just with somes rules.

Do you want that I add this as a reply to the issue #310 ? 

Thanks

berrnd:
Ok, I've reopened this issue to track this here.

For your example above we would also need something like a recipe categorie (soup/meat/fish).

Crupuk:
Thanks !

Exactly , tags would be perfect

PR #560, opened 2020-02-10
New Feature Recipe Catagories

kriddles:
This is not yet complete just a work in progress for you to checkout and make additional adjustments to.

New master data - Recipe catagories
New component dropdown
Additional Recipe option using component
Filter by Recipe Catagory on Recipes page

After adding the recipe catagory to the recipe it's not loading that saved option value when loading the recipe again.
I have not finished updating the filter on the recipes page so it's not currently working.


kriddles:
Might want to instead of adding a column to the recipes table to instead create a new table id, recipe_id, catagory_id so that recipes could be assigned multiple catagories.

I saw somewhere you have a multi select dropdown but I can't recall where.

berrnd:
> I saw somewhere you have a multi select dropdown but I can't recall where.

On the chore edit page, user assignments: https://demo.grocy.info/chore/2

When it's about new dropdowns - I would be happy to migrate everything (productpicker and so on) to [Select2](https://select2.org/), this should cover all our needs and seem to just work better: https://select2.github.io/select2-bootstrap-theme/4.0.3.html

Maybe we can even skip the separate recipe categories page and just allow selecting and adding categories at the same time (should be doable with Select2)...

kriddles:
> When it's about new dropdowns - I would be happy to migrate everything (productpicker and so on) to [Select2](https://select2.org/), this should cover all our needs and seem to just work better: https://select2.github.io/select2-bootstrap-theme/4.0.3.html
> 
> Maybe we can even skip the separate recipe categories page and just allow selecting and adding categories at the same time (should be doable with Select2)...

That looks great! Do you want to take this branch/feature from here and I'll hold off on making any changes?
That way you can make an example select2 the way you want it and then I can help migrating the other dropdowns to using select2?

berrnd:
Don't know when I will have time for it...

But also the recipe categories alone will not add that much alone without the other "generate a meal plan base on rules" things mentioned in #553...

berrnd:
> Maybe we can even skip the separate recipe categories page and just allow selecting and adding categories at the same time (should be doable with Select2)...

That's the example where items can be added by just typing them: https://select2.org/tagging

kriddles:
> But also the recipe categories alone will not add that much alone without the other "generate a meal plan base on rules" things mentioned in #553...

That meal plan feature has quite a few steps before we could get there and seems very specific to their needs. 
Just providing recipe catagorization seems to be a common main feature of other recipe apps.
Initially solving for more of a recipe divider feature.
Appetizers, Beverages.
Soups, Salads.
Vegetables.
Main Dishes.
Breads, Rolls.
Desserts.
Miscellaneous.

I would think the next simple step would be just to add the recipe catagory filter dropdown to the recipe add module for the meal plan. The catagory filter could then be used to filter the list in the recipe dropdown by catagory.

kriddles:
> Maybe we can even skip the separate recipe categories page and just allow selecting and adding categories at the same time (should be doable with Select2)...

A pro for also including the recipe categories page would be if the user wants to clean up any categories. They could use the Select2 to input new categories for the dropdown however they may accidentally create one they do not wish to always see in the dropdown. The interface would be needed to delete the category.

manojknaikade:
I think tags would be a better option than category. If a user needs to filter recipes, say, appetizer and vegetarian, it can be done by placing a filter list next to the search box where the user can select multiple values. 
while creating the recipe the user can get autocompletion from previously populated list of tags to avoid duplicates and faster tagging and also can create new tags on the fly. https://select2.org/tagging

berrnd:

PR #574, opened 2020-02-19
Button to Flashlight ON in Barcodescanner

radim-ek:
I made only button to ON flashlight, not to OFF, but thats not needed.
For better UI I allow CloseButton o righttop.

You need a make translate string "Torch".


berrnd:
Great, thank you!
Will do a quick test in the next days...

Do you know if it's possible to detect if the device has a flash light, to show the button only if that's the case?

radim-ek:
It's possible to test, but don't have time to implement it to this commit. Need some time to test it...

PR #588, opened 2020-03-01
Localization fix

kuchosauronad0:
I noticed a typo in an error message in the English localization when trying to scan a barcode. Looks like brainfart kinda thing, if not not feel free to ignore this PR :)

Change string 'Camera access is on only possible [..]' to 'Camera access is only possible [..]'

Cheers,


kuchosauronad0:
Ok thanks,
I tried to fix it in transifex and will close this PR now.

PR #592, opened 2020-03-06
Added chore name when getting list of all chores.

DarienFord:
When hitting the API to get a list of all chores, the response does not include the chore name.

I didn't want to then have to iterate through each one to get the name after getting the list of the chores as it could be significantly computationally expensive.

This PR just adds a join on the chores table to include the name of the chore when the list is retrieved.

### DIFF
--- a/services/ChoresService.php
+++ b/services/ChoresService.php
@@ -25,7 +25,7 @@ class ChoresService extends BaseService
 
 	public function GetCurrent()
 	{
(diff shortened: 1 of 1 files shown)


berrnd:
I just renamed the field to `chore_name` (instead of `name`).

DarienFord:
Well, that's ok I guess.  
I wouldn't have since the context of the query is already `chores` so the fields should all be in relation to that context.

PR #638, opened 2020-03-21
Fix recipes page

### LINKED ISSUE #637: With no recipes, errors are shown on page.
Notice: Trying to get property 'id' of non-object in /home/forge/food.runkel.org/controllers/RecipesController.php on line 36

Adding a recipe removes the errors.

mrunkel:
This fixes #637 
I don't know if you have a more generic way of handling this, but at least this prevents the errors from being displayed on the page when no recipe is selected.


berrnd:
Thank you.

(I don't really care about PHP notices and warnings, just disable them - and it's pretty rare that you access the recipes page while you don't have any recipe...)

PR #639, opened 2020-03-22
Check if all dependencies are installed

Forceu:
A lot of people had problems with installing grocy, as it only displays an empty page if there is an error.  This little helper does a quick and cheap check if all dependencies are installed and config.php / config-dist.php exists.

If something is missing, the script stops and outputs an error.


Forceu:
Done, I put gd in the list as well.

PR #657, opened 2020-03-24
Optionally use number pad input in best-before-date fields on mobile Chrome

Mik-:
If you set the feature flag FEATURE_FLAG_STOCK_BEST_BEFORE_DATE_FIELD_AS_NUMBER to True, the best-before-date input field gets additionally the `inputmode = "numeric"`. This leads to a number pad on some [mobile Browsers](https://caniuse.com/#search=inputmode) and allows a faster input of dates (using the [shorthands](https://github.com/grocy/grocy#input-shorthands-for-date-fields)).

And there was a missing translation string while undoing an action.


localization/de/strings.po
```diff
@@ -1957,3 +1957,6 @@ msgstr "Rezept-Karte"
 
 msgid "Group ingredients by their product group"
 msgstr "Zutaten nach Produktgruppen gruppieren"
+
+msgid "Transaction successfully undone"
+msgstr "Vorgang wurde rückgängig gemacht"
```
berrnd:
```suggestion
```

That's not needed, needs to be translated on Transifex (see https://github.com/grocy/grocy#localization).

config-dist.php
```diff
@@ -145,6 +145,7 @@
 Setting('FEATURE_FLAG_STOCK_PRICE_TRACKING', true);
 Setting('FEATURE_FLAG_STOCK_LOCATION_TRACKING', true);
 Setting('FEATURE_FLAG_STOCK_BEST_BEFORE_DATE_TRACKING', true);
+Setting('FEATURE_FLAG_STOCK_BEST_BEFORE_DATE_FIELD_AS_NUMBER', false);
```
berrnd: Maybe that better needs a short comment what it does...

Mik-:
I did some tests with other options, but the `inputmode` was the only working for me.
Neither `input type="number"` nor `pattern="[0-9]*"` allowed the dashes after leaving the input field and the real date was inserted. So the date validation always failed.
I saw some hacks, which change the `input` and `pattern` on focus and blur, but that looked a bit awkward to me.

config-dist.php
Mik-: Updated the naming of this option and added a comment.

Issue #447, opened 2019-11-23
Have a way to specify shop places / Price tracking per store


immae:
I’m not sure it is feasible with the current state of the code (with "user entities"), but it would be nice to be able to specify a list of (location, price) for a product, so that we can for instance compare prices between two shopping places (or see the evolution of price separately per shop), or prepare a shopping list specific to a given place, etc.

berrnd:
Userfields are attached to a product, it's currently not possible to track them on purchase/consume/etc.

Similar to #443, but I also leave that open as it is about price tracking per purchase location.

berrnd:
_Originally posted by @aaruni96 in https://github.com/grocy/grocy/issues/490#issue-548462452_

> It'd be really nice if while making a purchase, we can record what shop/supermarket the thing was purchased at, and the price tracker tracked prices per store.

> So, if I sometimes purchase product from store A, and sometimes from store B, I want to see two graphs in the price tracker : one for store A and one for store B

Belafestim:
i see this in the pre-release and am really pumped, can we additional fields address and phone numbers

PR #658, opened 2020-03-24
Add shopping location for price tracking

immae:
This PR adds the concept of shopping location to the price tracking feature.

Tracking price usually makes sense associated to a shopping location, since the price will evolve in that location but will have a different base when you buy it somewhere else. This PR adds this notion to it, and the corresponding charts will now have more than one line to show this difference, allowing the user to chose his/her shopping location depending on the price too.

This fixes https://github.com/grocy/grocy/issues/447

Things to look at specifically:
- [x] Translations were added manually. They seem to work but there might be automated tools to handle the many languages.

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -82,13 +82,19 @@ class StockApiController extends BaseApiController
 				$locationId = $requestBody['location_id'];
 			}
 
+			$shoppingLocationId = null;
+			if (array_key_exists('shopping_location_id', $requestBody) && is_numeric($requestBody['shopping_location_id']))
+			{
+				$shoppingLocationId = $requestBody['shopping_location_id'];
+			}
+
 			$transactionType = StockService::TRANSACTION_TYPE_PURCHASE;
 			if (array_key_exists('transaction_type', $requestBody)  && !empty($requestBody['transactiontype']))
 			{
 				$transactionType = $requestBody['transactiontype'];
 			}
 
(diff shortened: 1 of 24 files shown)


berrnd:
Wow, thanks! :)

> Translations were added manually. They seem to work but there might be automated tools to handle the many languages.

Yes, is done via Transifex, see: https://github.com/grocy/grocy#localization
So you only need to add the strings to `localization/strings.pot` (which is also automatically done when the file is writable and when `MODE` is `dev`).

immae:
Thanks @berrnd I added them to the file. I thought I was in dev mode but it turns out I didn’t change the default during development so I missed it.

berrnd:
Everything looked good, thanks again! :)

I did some smaller changes/fixes, see the referenced commits above.

immae:
Thanks for the change, store is indeed a better name!

jayaddison:
This is an introduction / mention for who @sumeetmoray who posted the [Nearby Shops](https://github.com/NearbyShops) project on Hacker News [post](https://github.com/NearbyShops/Nearby-Shops-API).

I like the fact that it's open source and internationalized (just like `grocy`), and that it's paid for by food vendors rather than users (i.e. it is financially supported by the businesses that generate revenue from it).

It'd be an impressive user experience to see real-time local product availability and pricing using `NearbyShops` app and then have purchased baskets loaded into `grocy`.

In general I think it's best to continue to focus on our own software projects & components to make sure they work as well as possible for their individual tasks, but I figured I'd make the connection/introduction here so that we could consider integration in future.

SumeetMoray:
hi @jayaddison thanks for making a mention. Sure i think it would be a great idea. As the project grows we would surely love to have integrations with other open source projects. We would put this on our To-Do list.

Issue #577, opened 2020-02-23
Feature request (API): search object by other fields


smhgit:
Currently search support only name. Will be helpful if it will support other (any) fields.

Use case: I want to get all items in shopping list which belongs to specific shopping list id.

berrnd:
That's now possible, excerpt from the changelog:

> - The output of the following endpoints can now be filtered (by any field), ordered and paginated (thanks @fipwmaqzufheoxq92ebc)
>   - `/objects/{entity}/search`
>   - `/stock/products/{productId}/entries`
>   - `/stock/products/{productId}/locations`
>   - `/recipes/fulfillment`
>   - `/users`
>   - `/tasks`
>   - `/chores`
>   - `/batteries`
>   - There are 4 new (optional) query parameters to use that
>     - `order` The field to order by
>     - `limit` The maximum number of objects to return
>     - `offset` The number of objects to skip
>     - `query[]` An array of conditions, each of them is a string in the form of `<field><condition><value>`, where
>       - `<field>` is a field name
>       - `<condition>` is a comparison operator, one of
>         - `=` equal
>         - `~` LIKE
>         - `<` less
>         - `>` greater
>         - `>=` greater or equal
>         - `<=` less or equal
>       - `<value>` is the value to search for

PR #667, opened 2020-03-27
added new shopping list api endpoints

sourcefactory-at:
I added some more api endpoints to receive shopping lists, receive their entries and update the todo state.

* /stock/shoppinglist
* /stock/shoppinglist/{listId}
* /stock/shoppinglist/{listId}/entries
* /stock/shoppinglist/set-done/{entryId}

Btw. thanks for the tool, it's great.

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -419,6 +419,59 @@ class StockApiController extends BaseApiController
 		));
 	}
 
+    public function CurrentShoppingLists(\Psr\Http\Message\ServerRequestInterface $request, \Psr\Http\Message\ResponseInterface $response, array $args)
+    {
+        $lists = $this->getStockService()->GetCurrentShoppingLists();
+
+        return $this->ApiResponse($response, $lists);
+    }
+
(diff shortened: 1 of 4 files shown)


berrnd:
Thank you, but that's not need - all base entities/objects can be get and set through the ["Generic entity interaction"-endpoints](https://demo.grocy.info/api#/Generic%20entity%20interactions):

Get all shopping lists:
```
https://demo.grocy.info/api/objects/shopping_lists
```

Get a single shopping list:
```
curl -X GET "https://demo.grocy.info/api/objects/shopping_lists/1" -H  "accept: application/json"
```

Get all items of all shopping shopping lists:
```
curl -X GET "https://demo.grocy.info/api/objects/shopping_list" -H  "accept: application/json"
```

Set the done flag of a specific shopping list item:
```
curl -X PUT "https://demo.grocy.info/api/objects/shopping_list/1" -H  "accept: */*" -H  "Content-Type: application/json" -d "{\"done\":\"1\"}"
```

It's currently not possible to get the items of specific list, you need to filter them yourself. I prefer to implement that in the same generic way, not specific for each entity - already tracked in #577.

sourcefactory-at:
well... I did see that for products but was not thinking about also using it for shopping lists. Indeed, it's better to use it in that generic way. Thanks for your prompt reply.

PR #668, opened 2020-03-27
Return amount of stock in location API request

Forceu:
For the `/stock/products/{productId}/locations` request, the amount of stock for each location is returned as well

I added it to the already existing 99.sql - is that the correct way or would it be better to use a 100.sql? The 99.sql is not in the current release.

### DIFF
--- a/grocy.openapi.json
+++ b/grocy.openapi.json
@@ -3547,6 +3547,9 @@
 					"product_id": {
 						"type": "integer"
 					},
+					"amount": {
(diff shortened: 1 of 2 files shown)


berrnd:
> I added it to the already existing 99.sql - is that the correct way or would it be better to use a 100.sql?

Database migrations should work between releases, so adding that to an existing not yet released migration is ok, I sometimes created multiple files to separate different topics, but doesn't really matter...

PR #690, opened 2020-04-02
Add ability to import json data from Kroger grocery stores

zebfross:
I shop at Kroger grocery stores (QFC, Fred Meyer, etc.), so I wanted to be able to import my purchases from their api.  Unfortunately, the Kroger api doesn't allow access to purchases, BUT through their website you can access purchase data in json format.  This change allows us to import that data into our database.

Here is a gif of the process and result: grocy_kroger

The main changes are as follows:

- Added KrogerToGrocyConverter that is in charge of converting the Kroger json data to the Grocy equivalent.  It is able to pull quite a bit of information from the json data.  Specifically:
    - Quantity
    - Price
    - Location - can be guessed based on product category
    - Barcode - translated from UPC
    - Picture - they give a picture url if it exists
If the location or quantity unit cannot be determined, I use the location or quantity unit from user settings.  Everything else have reasonable defaults.

- Added a menu option to "Upload JSON".  The form has only a few options but definitely room to grow.

- Added a few unit tests for the converter.  I noticed there weren't unit tests in the whole project.  Not sure how you've lived without them; I'd be happy to add more if they'd be appreciated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,5 +23,8 @@
 		"files": [
 			"helpers/extensions.php"
 		]
+	},
+	"require-dev": {
+		"phpunit/phpunit": "9"
 	}
 }
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "9c7420e77a3f1afb59e00560d5b9081c",
+    "content-hash": "087239c88ff18285970bf9079e165232",
     "packages": [
         {
             "name": "doctrine/inflector",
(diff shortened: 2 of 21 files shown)


berrnd:
Thank you a lot for your work!

But this seems to be very specific for Kroger - and don't know if this works for an "average user" when you need to copy & paste something from the browser developer console...

I'm more thinking about a general (plugin) way to do those things, like mentionend [here](https://github.com/grocy/grocy#barcode-lookup-via-external-services) there is already something that can import products, no purchase data yet and also an UI is missing, only possible through the API currently.

I also saw that you hardcoded location ids - that can never work...

>  I noticed there weren't unit tests in the whole project. Not sure how you've lived without them; I'd be happy to add more if they'd be appreciated.

Sure, feel free to add as many you want. :)

zebfross:
Yeah that's true; it would really only help a subset of people, at least until other grocery stores make data available.  I'll recall this and maybe think about it some more.

For the locations, it would be nice if we had something similar to how email folders work.  You can add whatever email folders you like, but there are always the main Inbox, Sent, Deleted, etc.

PR #692, opened 2020-04-02
favicon.ico retrieval: Specify profile for rel='icon', following W3C howto

### LINKED ISSUE #691: Firefox retrieves favicon from incorrect path
This is a very minor issue.  While checking for HTTP 404 response statuses during development of https://github.com/grocy/grocy-docker/pull/70, I discovered that Firefox (74.0) is attempting to retrieve the grocy `favicon.ico` from an incorrect (root) path despite [full paths specified in the source code](https://github.com/grocy/grocy/blob/v2.6.1/views/layout/default.blade.php#L14-L18).

favicon

jayaddison:
From reading the [W3C How to Add a Favicon to your Site](https://www.w3.org/2005/10/howto-favicon) it looks like specifying a '[profile](https://www.w3.org/2005/10/howto-favicon#profile)' may encourage browsers to retrieve page icons from the correct path.

Caveat: I have not yet been able to test this locally. 

Resolves #691

### DIFF
--- a/views/layout/default.blade.php
+++ b/views/layout/default.blade.php
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <html lang="{{ GROCY_CULTURE }}">
-<head>
+<head profile="http://www.w3.org/2005/10/profile">
 	<meta charset="utf-8">
 	<meta http-equiv="x-ua-compatible" content="ie=edge">
 	<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">


berrnd:
Also saw this with Firefox (not always, however) and just thought that's a Firefox thing to theck for it in the root path.

Also haven't tested it so far, if it resolves that, ok - otherwise we could also add or move the icon to the root path...

jayaddison:
Yep, also possible that this is an unrelated Firefox behaviour - I'll take a bit of a look through their bug tracker.  Until I can test this locally, we can just leave the ticket open.  It'd be nice to resolve 'correctly' if possible, but moving / duplicating the icon is certainly an option too.

jayaddison:
Firefox seems to be running through this line of code when I experience this locally:

https://hg.mozilla.org/mozilla-central/annotate/f53ecf8eee2a0d2a7705d7ea3ec8682028e922a0/browser/modules/FaviconLoader.jsm#l670

I'll try not to be too noisy on this issue thread as I continue to investigate.  Feel free to unsubscribe if not interested :)

favicon-loader

jayaddison:
This may be an issue with Firefox, and repro steps are available in https://bugzilla.mozilla.org/show_bug.cgi?id=1186324

Adding the `profile` attribute doesn't seem to affect the repro case or resolve the issue in that bug, so I don't think it is likely to help here either.  Closing this issue for now.

PR #702, opened 2020-04-06
API: Add some parameters to answer of /stock request

### LINKED ISSUE #701: API enhancement: Add some parameters to /stock response
Hey Bernd, could you please add the following parameters to the answer of the [/stock](https://de.demo.grocy.info/api#/Stock/get_stock) request?

- last_purchased
- last_used
- last_price
- average_shelf_life_days
- spoil_rate_percent

They're already in the answer of the [/stock/products/{productId}](https://de.demo.grocy.info/api#/Stock/get_stock_products__productId_) request, but for my Android App this request would be necessary for all products which the user clicks on and therefore there would be a loading delay after every click on a stock item.
I avoid this delay by making the /stock request on App start and then I already have most of the information which I want to display when clicking on a product. But: With this request, the parameters from above are currently missing, so I cannot display them.

It would be great if you can add these to the /stock request answer. :-)

MrJohnZoidberg:
This would fix #701.

I simply copied the statements from the function `GetProductDetails` below (with my humble PHP knowledge).
But I don't know if this simple approach is possible, because the database accesses for the parameters are made for every item in the stock overview. Maybe this could involve some latency?

Feel free to close this pull request if my approach is not so good.

### DIFF
--- a/services/StockService.php
+++ b/services/StockService.php
@@ -34,6 +34,30 @@ class StockService extends BaseService
 		{
 			$currentStockMapped[$product->id][0]->product_id = $product->id;
 			$currentStockMapped[$product->id][0]->product = $product;
+			
+			$productLastPurchased = $this->getDatabase()->stock_log()->where('product_id', $product->id)->where('transaction_type', self::TRANSACTION_TYPE_PURCHASE)->where('undone', 0)->max('purchased_date');
+			$currentStockMapped[$product->id][0]->last_purchased = $productLastPurchased;
+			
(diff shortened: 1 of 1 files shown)


berrnd:
> But I don't know if this simple approach is possible, because the database accesses for the parameters are made for every item in the stock overview. Maybe this could involve some latency? 

It will, for sure. The current `GetCurrentStock` function is already not the best one, but this would add another 5 SELECT statements per row. So I think we should do this better.

And at least not copy the existing `GetProductDetails` function to get those product details, just call it...

MrJohnZoidberg:
You're right, this is very inefficient. I will close this pull request and the issue, because I've implemented it now with making the `/stock/products/id` request after click on a stock item like you do it on the Grocy stock overview page.

PR #733, opened 2020-04-13
Add camera picker button

MichaelMadsen:
I have added a button to pick what camera to scan with. The preferred camera ID is saved in local browser storage, so it is persisted for a reasonable amount of time. No change to default behaviour has been made, so the initial camera choice is still determined by the Quagga constraints.

I have two primary reasons for adding this button:
1. A lot of newer phones have a frightening amount of cameras on the back. On my own Galaxy S10, Quagga chose the 0.5x landscape camera by default. This camera has no autofocus and distorts the image, making it very bad for barcode scanning.
2. In the future I'd like to be able to hang a tablet on the fridge and use the front camera for scanning barcodes.

Here are a couple of screenshots of the proposed change:

1. Using the back camera
scanner1

2. After clicking the camera picker button
scanner2

3. Changing the camera
scanner3

4. Using the front camera
scanner4

Picking a new camera re-initialises Quagga, which also means the flash button is added/removed if you start using a camera with different capabilities. Notice how the flash button is visible when using the back camera. Neat!

I also changed the `Cancel` button to be red, to make the buttons more visually distinct. I can undo it if you don't like it :)

### DIFF
--- a/public/viewjs/components/barcodescanner.js
+++ b/public/viewjs/components/barcodescanner.js
@@ -1,11 +1,18 @@
 Grocy.Components.BarcodeScanner = { };
 
-Grocy.Components.BarcodeScanner.CheckCapabilities = function()
+Grocy.Components.BarcodeScanner.CheckCapabilities = async function()
 {
 	var track = Quagga.CameraAccess.getActiveTrack();
 	var capabilities = {};
 	if (typeof track.getCapabilities === 'function') {
 		capabilities = track.getCapabilities();
+	}
+	
+	// If there is only 1 camera, hide the camera select button
+	var cameras = await Quagga.CameraAccess.enumerateVideoDevices();
+	var cameraPickerButton = document.querySelector('.cameraPicker');
+	if (cameraPickerButton) {
+		cameraPickerButton.style.display = cameras.length > 1 ? 'inline-block' : 'none';
 	}
 	
 	// Check if the camera is capable to turn on a torch.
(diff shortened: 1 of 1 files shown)


berrnd:
Great idea, makes sense, thanks! :)

But a dialog over another dialog is maybe not really the best way, how about to put that camera dropdown just in a section below the camera stream picture / above the buttons?

Cancel button color: I'm by far no UI guy, but I think red should only be used for "critical" actions...

> In the future I'd like to be able to hang a tablet on the fridge and use the front camera for scanning barcodes.

Here off-topic, but just my personal recommendation: If not a real portable solution is needed, just use any cheap USB (wireless) barcode laser scanner, works 1000 % faster, in any lighting condition, and from any angle...

MichaelMadsen:
Great suggestion! I think I stared myself blind on trying to make the nested dialog work. Appending a select to the existing dialog is a much simpler solution.

scanner5


off-topic:
I think a lot of people would not spend money on a scanner without *knowing* that grocy works for them. So I think making the camera scanner as robust as possible is a good investment for growing the user base. 
Apart from that, I don't have a dedicated machine/tablet that I want to set up in my kitchen right now. But everybody in the household has a smartphone with bookmarks to the grocy Purchase and Consume pages.

Of course, I know that a cheap scanner would be objectively quicker, and I still might get one later :)
Do you have any recommendations?

berrnd:
> Do you have any recommendations?

I also have to do with such hardware in my job, for basic barcodes, like used for food, I would say there is no real difference - any does the cheap pretty good, also the ones below 20 EUR...

I personally have a wireless one by Inateck in my kitchen, most similar recent model seems to be this one: https://www.amazon.de/dp/B075XKWG9M

Have grocy always running with the consume page open and "Scan mode" (which provides audio feedback) on, so removing something works without looking at the screen by just scanning the item...

Issue #263, opened 2019-05-23
Feature Request: User Rights Assignment


CyberRad:
I believe there is a need to restrict what to allow users to do within this product. For example a child could only have access to remove stock and view their chores list.

berrnd:
Great idea, also relates to #207.

So first we would need to define which roles/groups with which permissions would be needed. Would be great if you could make a proposal.

CyberRad:
I would think the ability to make your own roles/groups would be beneficial. I assume each implementation (household) would want to define what they are comfortable with instead of using out of the box groups. For example I might be comfortable with my child adding items to the shopping list but others may not.

berrnd:
Sure, I didn't mean the roles, more the "actions", or call it "permissions".

So from your example:
- View own chores
- Remove stock

These need to be defined because they need to be implemented in the destination (page) - of course you then define a own named role/group "Childs" and assign those actions/permissions to it...

CyberRad:
Oh I see... let me brainstorm and I will get back with you.

CyberRad:
Here is my crack at it. I am sure it is missing some functionality.

View:
Stock Overview
Shopping List
Recipes
Chores
Tasks
Batteries
Equipment
Calendar
Meal Plan

Function:
Add/Remove inventory items (Purchase/Consume/Chore/Battery)
Chore tracking/check off chore (user specific #253)
Battery tracking
Add/Remove equipment

Edit:
Any inventory item
Meal plan
Recipes
Equipment

Wordna1:
The ability to add notes to the Meal Plan should be separate from everything else such that a given user could only have permission to mark their chores/tasks as complete, consume recipes assigned to them and add notes to the Meal Plan (eg. "I think the provided recipe should be changed in such and such manner..." or "Too many chicken-based meals have been scheduled for this day", etc)

klausmcm:
I'd like to add that the ability to change passwords for other users should be restricted - either to admin or have the ability to create user accounts with admin privileges.

Germs2004:
This is a great idea!  Children should not have access to "delete chores" :)

berrnd:
This is now done, see #960.

Excerpt from the changelog:
> ## New feature: Permissions
> - Users can now have permissions, can be configured per user on the "Manage users" page (lock icon)
> - Default permissions for new users can be set via a new `config.php` setting `DEFAULT_PERMISSIONS` (defaults to `ADMIN`, so no changed behavior when not configured)
[...]

PR #735, opened 2020-04-13
Add email

kriddles:
WIP 

These current two commits just lay some framework to add Email to the users, add phpmailer, and a possible table for user password reset.

Initially taking a look at this link
https://www.allphptricks.com/forgot-password-recovery-reset-using-php-and-mysql/

I think the next step is to make a ResetUserPassword in the UserApiController, what do you think?
Have you looked a phpmailer or did you have something else in mind for emails?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,8 @@
 		"gettext/gettext": "^4.8",
 		"eluceo/ical": "^0.16.0",
 		"erusev/parsedown": "^1.7",
-		"gumlet/php-image-resize": "^1.9"
+		"gumlet/php-image-resize": "^1.9",
+		"phpmailer/phpmailer": "^6.1"
 	},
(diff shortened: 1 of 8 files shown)


berrnd:
> Have you looked a phpmailer or did you have something else in mind for emails?

Don't have a preference, but I think PHPMailer is well known and so the thing to use for this...

But when we talk about a password reset function, maybe we also need to start something about permissions in general (#263) - currently every user can change everything and therefore the password of anyone...

But thanks! :)

kriddles:
> But when we talk about a password reset function, maybe we also need to start something about permissions in general (#263) - currently every user can change everything and therefore the password of anyone...

Very true, unless there's only one user... who happened to forget their password...

berrnd:

PR #745, opened 2020-04-16
Remove NodeJS dependency check

### LINKED ISSUE #744: [v2.7.0] Prerequisite Checker: Workaround required when using separate host / CDN for static content
If `grocy v2.7.0` is deployed from release tarball in an environment where PHP content is served by separate host(s) from static web content, the [PrerequisiteChecker::checkForYarn](https://github.com/grocy/grocy/blob/1cb0ba5e68b0b4069416ed3e9565bebb12c808b4/helpers/PrerequisiteChecker.php#L43-L49) step fails.

It's possible to work around this by creating a file or directory to satisfy the check, but that doesn't achieve the intent of the check.

Relates to https://github.com/grocy/grocy/pull/639

jayaddison:
It may be worth performing this NodeJS dependency validation check outside of the `grocy` application.

PHP code is well-suited to self-test for presence of configuration file and PHP library dependencies, while other infrastructure may be better suited to check for the presence and integrity of static content.

Previous `grocy` zipfile releases have included the full PHP and NodeJS file dependencies in the expected file path locations.  If that convention continues, then users deploying a simple installation from a release zipfile should be unlikely to encounter situations where PHP requirements are fulfilled and yet NodeJS dependencies are missing.

Fixes #744

### DIFF
--- a/helpers/PrerequisiteChecker.php
+++ b/helpers/PrerequisiteChecker.php
@@ -11,7 +11,6 @@ class PrerequisiteChecker
         self::checkForConfigFile();
         self::checkForConfigDistFile();
         self::checkForComposer();
-        self::checkForYarn();
         self::checkForPhpExtensions();
     }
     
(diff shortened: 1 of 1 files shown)


berrnd:
Thanks.

(I also noticed other problems just after releasing the new version with the new PrerequisiteChecker - for example the data directory path can be overriden by https://github.com/grocy/grocy#embedded-mode which is also not handled yet...)

PR #753, opened 2020-04-18
Cosmetic Update

zsarnett:
### Stock Overview
image

### Shopping List
image

### Recipe
image

### Chore Overview 
image

### Tasks
image

### Batteries
image

### Equipment
image

### Products
image

And more :)

Here to work with you on any changes :)

Issue #740, opened 2020-04-16
Shopping List Printing Template


hmronline:
Hi, first of all, thank you for developing this!

The print button provides an export of the shopping list that look nice, however the font size was pretty small making it difficult to read at the supermarket.

Please find below some suggestions to make it more compact (in order to carry a smaller sheet of paper at the supermarket) and more readable (avoid blank spaces and increase font size):

Item/Product quantity should be placed on the first column instead of second column as it is right now.
Quantity text is shorter so it would fit better on the left, and column size should not be fixed to avoid wasting blank space between columns.
I think it also reads better: 2 pieces of product X

It would be great to have a few additional options in the configuration menu:
 * Font size: at least small, medium, large
 * Layout: 1 column (as it is right now), 2 columns (to save some space and make it fit in one page)
 * Include Headers (yes/no), or at least make it smaller.

Hope this helps to improve that feature.

Thanks in advance!
Hernán.-

CRTified:
It would also be helpful to include the product group headers in the print view.

The 2-Column option can be achieved by using the printer option "Print 2 pages on 1". The font size is fine (at least for me, YMMV).

lousando:
![image](https://user-images.githubusercontent.com/1214300/79630787-36a9ae00-8109-11ea-9286-5bf8e1d1c8a1.png)

Currently working on this enhancement :)
I will add options for headers (yes/no) next.
Re-ordering of columns might take a bit longer, but I'll make a PR before that sub-feature and possibly and that in a future revision.

bfayers:
I would like to see the product groups included on the printed list in the same way they are included on the web view list.

hosios:
I would also urge these options for printing. I know Grocy is suposed to be used with the app, and it is handy, but our times with COVID-19 changed that. We now opt for delivery, and then you have to have a printed list (or a pdf for that matter). 
So, pelase make these options as soon as possible. 

Thank you vor a great product

edenhaus:
I found https://www.reportbro.com which is generating from a template a pdf. Also it supports a template designer so the user can easy adopt and configure the required layout for it.
Before starting to implement there are a few questions to solve:
- The pdf is generating in the backend, which is written in python. @berrnd  Would you accepted a PR which is forwarding a pdf request to a python application? The first idea in my mind was a python rest server, which is only listening on localhost and so only the php backend can call it. If you have a better idea, then feel free to write it 😊
- Opensource can use it for free but we have to check the license. I will do that when the first one is answered.

[...]

PR #754, opened 2020-04-18
Add print options onto Shopping List page

lousando:
## Edited: 06/12/2020

**Enhancements:**

- Allows users to change text size of the print table
- Also allows users to hide the list headers

_Note:_ None of the options are saved to localStorage or the database.

**Current Modal Options:**

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 /vendor
 /.release
 embedded.txt
+
+# jetbrains IDE
+.idea
--- a/localization/en/strings.po
+++ b/localization/en/strings.po
@@ -1205,3 +1205,9 @@ msgstr "This is required and can only contain letters and numbers"
 
 msgid "Edit userfield"
 msgstr "Edit userfield"
+
+msgid "Print Text Size"
+msgstr "Print Text Size"
+
+msgid "Show Print Headers"
+msgstr "Show Print Headers?"
--- a/localization/strings.pot
+++ b/localization/strings.pot
@@ -1810,3 +1810,9 @@ msgstr ""
 
 msgid "Save & return to recipes"
 msgstr ""
+
+msgid "Print Text Size"
+msgstr ""
+
+msgid "Show Print Headers"
+msgstr ""
--- a/public/viewjs/shoppinglist.js
+++ b/public/viewjs/shoppinglist.js
@@ -343,12 +343,6 @@ function OnListItemRemoved()
 }
 OnListItemRemoved();
 
-$(document).on("click", "#print-shopping-list-button", function(e)
-{
-	$(".print-timestamp").text(moment().format("l LT"));
-	$("#description-for-print").html($("#description").val());
-	window.print();
-});
 
 $("#description").on("summernote.change", function()
 {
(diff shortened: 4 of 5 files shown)


berrnd:
Thanks.

Have you noticed that there is currently a separate tables used for printing only?
https://github.com/grocy/grocy/blob/59650728a2abf126d703cca355b1f869b06f9ed2/views/shoppinglist.blade.php#L200-L255

Maybe it would be the best thing to create a new page for printing a shopping list with all required options...

I would also prefer to merge #753 before, so a separate page would also make it simpler because of avoiding conflicts...

And I don't know if it's really necessary to introduce Vue for that small part, of course it would be maybe great to use it everywhere, but that would be a rewrite of the frontend in my opinion...

lousando:
Ok, I can move to a separate page 👌

I agree that Vue is a bit overkill for that one feature, but it makes data binding much easier in the front end. If you would like I can also remove it.

berrnd:
> I agree that Vue is a bit overkill for that one feature, but it makes data binding much easier in the front end. If you would like I can also remove it.

From the current added functionality, I see nothing special which can't be handled "natively" in not more or complexer lines of code...

Another idea:
Use the existing table and just apply style changes when using the print dialog, things like sorting or column reordering is already possible since all tables use [DataTables](https://datatables.net/)...
(Can't remember why I have added that second table only for printing and have not done it that way when I've added the printing option...)

lousando:
> Use the existing table and just apply style changes when using the print dialog, things like sorting or column reordering is already possible since all tables use DataTables...
(Can't remember why I have added that second table only for printing and have not done it that way when I've added the printing option...)

I might've been because you get more flexibility from creating a separate container for it 🤷‍♂️

lousando:
>From the current added functionality, I see nothing special which can't be handled "natively" in not more or complexer lines of code...

Performance-wise I will still agree with you that Vue is overkill; we gain almost nothing there. 😞 
I still think developers can still benefit from DOM manipulation even in such a small feature though as it takes an imperative approach to UI building. The benefits of Vue, in this case, are with the `v-model`, `v-if`, and `@click` statements: 🚀 

- `v-model`:  Updates `printTextSize` and allows it to be printed to the screen without manually selecting the DOM node. e.g. `$("#print-text-size").value(printTextSize)`

- `v-if`: Hides/Shows `showPrintHeaders` without manually needing to do `$("#print-staging-area th").hide()`

- `@click`: Binds event handlers once without listener on the parent element and without using inline `onclick`.

zsarnett:
I don't know if moving the Print button makes sense. Especially with the new UI changes.

lousando:
@zsarnett Just got around to looking at the changes; awesome work 👌. I will rebase my branch off of that and re-do my changes.

With the style changes though, the button still only appears when `GROCY_FEATURE_FLAG_SHOPPINGLIST_MULTIPLE_LISTS` is enabled. I think the print button should show even in single list mode, right?

zsarnett:
I do agree with that. It can be removed from the if statement and left in the header, no?

hmronline:
Hi @lousando, 

Have you had any chance to continue this work ?

I will be glad to test this feature and provide feedback, so please let me know if that helps.

Thanks in advance!
Hernán.-

lousando:
@hmronline Not yet, I think I will have time later this weekend however. I will ping you when I have something 👍

lousando:
@hmronline Online you can go ahead and test it now :)

@berrnd I was having trouble adding this to a new page; my PHP is not very good. _However_, I did remove Vue :)

@zsarnett This now builds off of your new style changes :rocket:

berrnd:
This is now done, see https://github.com/grocy/grocy/issues/740#issuecomment-748536983.

Issue #760, opened 2020-04-18
UI Enhancement: Swap purchase Best Before and Amount position


lyz-code:
When `purchasing` a new product, the Best before is usually automatically generated, while the Amount is not.

So I need to hit TAB twice, if they were swapped, I'd only have to hit it once.

It's a small improvement but still.

Thanks!

berrnd:
Makes sense. The inventory page has it unintentionally already ordered that way...

berrnd:
Fixed by #767.

Issue #395, opened 2019-09-26
Internal enhancement: Persistent demo intances to share test scenarios


berrnd:
References https://github.com/grocy/grocy/issues/384#issuecomment-535642827.

As grocy gets more and more complex, also testing scenarios get more and more complex and it's maybe sometimes not enough to describe them by text...

> Maybe it would be good to have some kind of "persistent" demo instances where scenarios could be saved/shared (currently the stable demos are recreated every 24 hours, the pre-release demo past 10 minutes every hour)

Kind of also references #129.

PR #767, opened 2020-04-19
Purchase Price Hints

kriddles:
A number of items to review

- Update products_average_price view with rounding
- Add average price to added to product card
- Recipe costs, use default consume rule which is "First expiring first, then first in first out"
   Achieved by updating the recipes_pos_resolved to use the new products_oldest_price view
- Update product details to use products_last_purchased view to reduce db calls
- deprecate products_current_price due to using products_last_purchased and changing recipe_pos_resolved

**update to recipe costs calculation in recipes_pos_resolved**
- when using only_check_single_unit_in_stock, now set the case to use 0 so that the costs is not included in the recipe costs. Logic is, if the product will not be consumed, then why add it to the costs?
- move rp.amount outside of if statement to always be used
- remove logic divisor of p.qu_factor_purchase_to_stock, the stock price is saved at the qu level and not the purchase level so is not necessary

### DIFF
--- a/public/viewjs/components/productcard.js
+++ b/public/viewjs/components/productcard.js
@@ -80,7 +80,7 @@ Grocy.Components.ProductCard.Refresh = function(productId)
 
 			if (productDetails.last_price !== null)
 			{
-				$('#productcard-product-last-price').text(Number.parseFloat(productDetails.last_price).toLocaleString() + ' ' + Grocy.Currency);
+				$('#productcard-product-last-price').text(Number.parseFloat(productDetails.last_price).toLocaleString() + ' ' + Grocy.Currency + ' per ' + productDetails.quantity_unit_purchase.name);
 			}
 			else
 			{
--- a/public/viewjs/purchase.js
+++ b/public/viewjs/purchase.js
@@ -161,6 +161,8 @@ if (Grocy.Components.ProductPicker !== undefined)
 						Grocy.Components.LocationPicker.SetId(productDetails.location.id);
 					}
 
+					$('#amount_qu_unit').attr("qu-factor-purchase-to-stock", productDetails.product.qu_factor_purchase_to_stock);
+					$('#amount_qu_unit').attr("quantity-unit-stock-name", productDetails.quantity_unit_stock.name);
 					if (productDetails.product.qu_id_purchase === productDetails.product.qu_id_stock)
 					{
 						$('#amount_qu_unit').text(productDetails.quantity_unit_purchase.name);
@@ -169,6 +171,12 @@ if (Grocy.Components.ProductPicker !== undefined)
 					{
 						$('#amount_qu_unit').text(productDetails.quantity_unit_purchase.name + " (" + __t("will be multiplied by a factor of %1$s to get %2$s", parseFloat(productDetails.product.qu_factor_purchase_to_stock).toLocaleString({ minimumFractionDigits: 0, maximumFractionDigits: 2 }), __n(2, productDetails.quantity_unit_stock.name, productDetails.quantity_unit_stock.name_plural)) + ")");
 					}
+
+					var priceTypeUnitPrice = $("#price-type-unit-price");
+					var priceTypeUnitPriceLabel = $("[for=" + priceTypeUnitPrice.attr("id") + "]");
+					priceTypeUnitPriceLabel.text(productDetails.quantity_unit_purchase.name + " price");
+
+					refreshPriceHint();
 
 					if (productDetails.product.allow_partial_units_in_stock == 1)
 					{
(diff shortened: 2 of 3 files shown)


berrnd:
Thank you. Maybe will some time to review this, just one thing, maybe I understood it just wrong:

> remove logic divisor of p.qu_factor_purchase_to_stock, the stock price is saved at the qu level and not the purchase level so is not necessary

Prices in `stock` and `stock_log` are currently always related to the products **purchase QU**.

kriddles:
Ah, there was a change I didn't notice that updated how prices are saved to the stock table commit 9a93c32

I will make an example to see if I understand based on the demo products.

kriddles:
Eggs in the demo are set to pack and piece with a qu_factor_to_purchase_stock of 10
I consumed all current entries for eggs to start fresh.
The question is, How much should an individual egg cost?

Purchase using the Unit Price of 1
image

Purchase using the Total Price of 10
image

Now when I checkout my stock entries, I have the amount I expect to see of 20 total eggs.
The price saved using current code set's the unit price to $1 and $10.
image

Should an individual egg cost $1 or $0.10 based on the qu_factor_to_purchase_stock and how does the $10 fit into the math now?

If the expected price per egg is supposed to be $0.1, then the stock price will need to be adjusted every time a unit is consumed in order to keep the price correct in the stock table.

If the expected price per egg is supposed to be $1, then I believe the 9a93c32 should be reverted?


On a side-note, I recognized I need to rethink my changes for "Update product details to use products_last_purchased view to reduce db calls" since stock_log is required. My changes only work if there is current stock which I did not properly think through.

berrnd:
When a pack costs 10 then one eggs of course costs 1. But currently prices displayed on the productcard/stock entries page are related to the products **purchase QU**. **Only for recipes** the price is based on the _stock QU_.

Total/unit price on the purchase page is related to the amount only. Means for an amount of 1 (pack) id doesn't matter if you select unit or total price. If you buy 3 packs, select "Total price" and enter $10, then 1 pack gets $3.33 in stock. If "Unit price" is selected, then each pack costs $10.

I also did the example with eggs and can't reproduce a problem. Regardless of total/unit price and buying 1 pack, I got $10 in stock (stock entries apge) both times.
When I use 1 egg in a recipe, it costs $1 there.

(We need #395 to be able to share those test cases better.)

views/recipes.blade.php
```diff
@@ -174,7 +174,7 @@
 					@if(GROCY_FEATURE_FLAG_STOCK_PRICE_TRACKING)
 					<div class="col-6">
 						<label>{{ $__t('Costs') }}&nbsp;&nbsp;
-							<span class="small text-muted">{{ $__t('Based on the prices of the last purchase per product') }}</span>
+							<span class="small text-muted">{{ $__t('Based on the prices of the default consume rule which is "First expiring first, then first in first out"') }}</span>
```
zsarnett, line 177: It may be more UI friendly if we removed this from the small text and just added it as hover text?

berrnd: Yes, also thought about that for all places where we have that currently, maybe a little help/info icon where the text will be shown on hover...

But fits better in your other pull request #776, as this is more about the logic behind recipe costs.

kriddles: Thanks for the suggestion. Hover text is not used anywhere else at the moment that I am aware of so I would not want to start with just this. Hints (small text) are used throughout the UI.

kriddles:
> When a pack costs 10 then one egg of course costs 1. But currently prices displayed on the productcard/stock entries page are related to the products **purchase QU**. **Only for recipes** the price is based on the _stock QU_.
> 
> Total/unit price on the purchase page is related to the amount only. Means for an amount of 1 (pack) it doesn't matter if you select unit or total price. If you buy 3 packs, select "Total price" and enter $10, then 1 pack gets $3.33 in stock. If "Unit price" is selected, then each pack costs $10.
> 
> I also did the example with eggs and can't reproduce a problem. Regardless of total/unit price and buying 1 pack, I got $10 in stock (stock entries page) both times.
> When I use 1 egg in a recipe, it costs $1 there.
> 
> (We need #395 to be able to share those test cases better.)

Due to the total_price issue that was fixed, I need to update prices in my production db to match this logic as described. I agree it is now working as it should.

I am working through some changes to push for this, but they are more UI centric to provide more information to the user. Such as, instead of saying "Unit Price" on the purchase page, I would recommend including the name of the QU_purchase such as in this case eggs = "pack". It would say "Pack price" instead of "Unit Price" and then provide the calculated stock cost in a hint. If "Total Price" is picked then a hint will provide the pack cost breakdown against the amount as a hint.

Hopefully that will make it so I don't have to question how the prices work again in 3 months when I look at it.

berrnd:
> It would say "Pack price" instead of "Unit Price" and then provide the calculated stock cost in a hint. If "Total Price" is picked then a hint will provide the pack cost breakdown against the amount as a hint.

Great idea! :)

views/recipes.blade.php
zsarnett: Agree. Will look to add it in my PR. @kriddles Only one way to start 😉

kriddles:
Reduced commit to just adding price hints

- product card last price now included 'per $qu_purchase.name'
- purchase page now includes dynamic price hint for unit price and total price calculations

I'll create issues for the other initial ideas so this can be closed out if approved.

kriddles:
Added commits for 
#760 
#752

berrnd:
I just fixed some things:
- The added texts have not used the translate function
- Numbers/prices were not displayed browser locale specific

kriddles:
> I just fixed some things:
> 
> * The added texts have not used the translate function
> * Numbers/prices were not displayed browser locale specific

Thanks!

PR #772, opened 2020-04-19
Japanese translation file(string.po) adding new stirng and minor fix

kshigeyama:


berrnd:
Thank you.

But translations are manage via [Transifex](https://www.transifex.com/grocy/grocy), as [stated in README](https://github.com/grocy/grocy#localization). Please add/change the translations there, thanks.

PR #788, opened 2020-04-22
Create Inverse QU Conversions

### LINKED ISSUE #388: Add inverse when adding quantity conversion
If I add "1 tablespoon = 3 teaspoons" it would be awesome if it simultaneously adds "1 teaspoon = 1/3 tablespoon".

kriddles:
Fixes #388 

When creating a QU conversion there is a new checkbox to offer creating the inverse.

Editing QU Conversions would require some refactoring or much more in depth updates.

### DIFF
--- a/public/viewjs/quantityunitconversionform.js
+++ b/public/viewjs/quantityunitconversionform.js
@@ -5,6 +5,11 @@
 	var jsonData = $('#quconversion-form').serializeJSON();
 	jsonData.from_qu_id = $("#from_qu_id").val();
 	Grocy.FrontendHelpers.BeginUiBusy("quconversion-form");
+	if ($("#create_inverse").is(":checked"))
+	{
+		var inverse_to_qu_id = $("#from_qu_id").val();
+		var inverse_from_qu_id = $("#to_qu_id").val();
+	}
 
 	if (Grocy.EditMode === 'create')
 	{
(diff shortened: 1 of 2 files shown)


kriddles:
> Thanks, but have you once tried it?
> 
> The inverse conversion is created as stated in the hint...
> 
> firefox_2020-04-23_20-41-33

lol oops... err.. yeah, I fixed that

I tried it but didn't look at the math very well as I was trying to muck with handling the hard part of editing and offering to edit the inverse too but just said good enough with offering to create the inverse.

berrnd:
Also have spontaneously no idea how to display/handle that when editing a conversion...

Another approach for this would maybe to not create the inverse conversions, but just populate them in the `quantity_unit_conversions_resolved` view, but then again how to display this / what happens when a auto-populated conversion is also added manually (with a different factor)...

kriddles:
> Also have spontaneously no idea how to display/handle that when editing a conversion...
> 
> Another approach for this would maybe to not create the inverse conversions, but just populate them in the `quantity_unit_conversions_resolved` view, but then again how to display this / what happens when a auto-populated conversion is also added manually (with a different factor)...

I thought about updating the database view too but there were too many places that used the table that I was not sure it would be enough to just update the view.

I figure this is most helpful when setting up new QU so covers 95% of the situational need for it... 

Some sort of QU Overview page could be helpful to see all the QU setup in one screen for review.

kriddles:
rebased to master

Issue #787, opened 2020-04-22
Recipe gallery images do not retain aspect when window width exceeds 1920px


vorkuta:
Recipe gallery image lose their original aspect ratio when window width exceeds 1920px, they are stretched. No vertical stretching from what I can see. The images seen in the recipe window itself (right hand side) do not appear to be affected.

![Correct aspect](https://user-images.githubusercontent.com/6390017/79988595-be474200-84a6-11ea-998c-bec1454b8af6.PNG)

![Stretched aspect](https://user-images.githubusercontent.com/6390017/79988605-c0a99c00-84a6-11ea-8150-4665bb05ef82.PNG)

zsarnett:
Working on a PR that I can include a fix for this

PR #795, opened 2020-04-24
Recipe updates

zsarnett:
- Update UI of Recipe Card
- Update UI of Gallery Cards
- Update logic for recipe creation - Only creates recipe once saved
- Update Recipe controller logic

### Before
image

### After
d886c07f3b5b09c8b97ccac509f76093
image

Fixes: #787

### DIFF
--- a/controllers/RecipesController.php
+++ b/controllers/RecipesController.php
@@ -17,27 +17,41 @@ class RecipesController extends BaseController
 		$recipesResolved = $this->getRecipesService()->GetRecipesResolved();
 
 		$selectedRecipe = null;
-		$selectedRecipePositionsResolved = null;
+
 		if (isset($request->getQueryParams()['recipe']))
 		{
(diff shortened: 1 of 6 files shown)


zsarnett:
There is a lot here. a Lot to test. Sorry about that. Let me know if you have any questions.

berrnd:
Wow, thanks again for all your work, looks amazing! 🤩

> Update logic for recipe creation - Only creates recipe once saved

The problem was that all was on one page so far, so including ingredients, and for adding/saving an ingredient a `recipe_id` was needed - but splitting the recipe "head" and ingredients creation should also work - will test it later...

zsarnett:
Yea this way the system isn't prematurely creating a new Recipe. If I go to create a recipe and then end up forgetting something or now don't want to add it. I would just click back and would expect there not to be a new recipe created.

vorkuta:
Nice one, thanks.

Issue #570, opened 2020-02-17
Weekly chores scheduled wrong when the next execution is in the same week


shadow7412:
When you set the period type to "Weekly" the hint text reads;

> This means the next execution of this chore is scheduled 1 day after the last execution, but only for the weekdays selected below

I read this to mean that if I tick, say, Thursday and Saturday then after I complete the chore on Thursday the next run should be on Saturday (and vice versa).

The current behaviour seems to be that when completing a chore on Thursday, it schedules it for next Thursday.

Period is currently set to 1 - the lowest setting that the UI allows.

How do I get a chore to behave in this way? Am I doing something wrong?

berrnd:
>Thursday and Saturday then after I complete the chore on Thursday the next run should be on Saturday (and vice versa).

Yes, that's how it should work - I can reproduce this, so is of course a bug.

berrnd:
Unfortunately there is still a problem about this.

Noticed by @shadow7412 in _https://github.com/grocy/grocy/commit/c6e06ab07c4fa734837a95aba5ef48c35ed8003a#commitcomment-38734078_

> Eg: I have a chore that recurs on Saturday and Wednesday. If I complete it on Saturday, it'll set it for next Saturday because it's the first listed in that switch statement despite the fact that Wednesday is closer...

shadow7412:
This is what I was thinking;

```sql
SELECT
  x.chore_id,
  x.last_tracked_time,
  CASE WHEN x.rollover = 1 AND DATETIME('now', 'localtime') > x.next_estimated_execution_time THEN
    DATETIME(STRFTIME('%Y-%m-%d', DATETIME('now', 'localtime')) || ' ' || STRFTIME('%H:%M:%S', x.next_estimated_execution_time))
  ELSE
    x.next_estimated_execution_time
  END AS next_estimated_execution_time,
  x.track_date_only,
  x.next_execution_assigned_to_user_id
FROM (

SELECT
  h.id AS chore_id,
  MAX(l.tracked_time) AS last_tracked_time,
  CASE h.period_type
    WHEN 'manually' THEN '2999-12-31 23:59:59'
    WHEN 'dynamic-regular' THEN DATETIME(MAX(l.tracked_time), '+' || CAST(h.period_days AS TEXT) || ' day')
    WHEN 'daily' THEN DATETIME(IFNULL(MAX(l.tracked_time), DATETIME('now', 'localtime')), '+' || CAST(h.period_interval AS TEXT) || ' day')
[...]

PR #800, opened 2020-04-26
#570 Added fix for weekly chores with multiple days

shadow7412:


berrnd:
Just tested this, but seems to be the same at the end, the view still uses the first day starting by saturday (?), or maybe I'm wrong now...

Had a chore scheduled for wednesday and saturday, last execution on 2020-04-25, next execution will be 2020-05-02...

shadow7412:
It was working yesterday... which makes me think that it's basing the calculations on "now" instead of the last run.

Now that I have a chore that was supposed to be done yesterday (and thus should be overdue) I'll have a fiddle.

shadow7412:
See https://github.com/grocy/grocy/pull/802

PR #801, opened 2020-04-26
Qu factor purchase to stock & Product Barcode Details

### LINKED ISSUE #204: Link products of different size
During my last grocery shop, we got an 18 pack of eggs; we usually buy a 12pk. It would be great to have two different products linked to one inventory item: "eggs".

Barcode XXXX = +18 eggs
Barcode YYYY = +12 eggs

Some sort of "sub-product" system would be ideal for things like this.

### LINKED ISSUE #565: Hide/archive products instead of deleting
Problem: I bought these awful vegan broccoli balls. I won't ever buy them again, but they come up every time I search for broccoli in purchase/inventory/etc.

Solution #1: Delete the product - but then, I'm worried that a possible future dashboard/analytics would be incorrect and skewed.

Solution #2: Add an option to hide or archive products. The product continues to exist in the master data section, but doesn't show up in searches in other areas. 

This is very low priority.

### LINKED ISSUE #785: Changing qu_factor_purchase_to_stock affects historical prices and current recipe costs
Low priority since this may rarely occur...
but since price is saved in the stock and stock log table based on the point in time factor qu_factor_purchase_to_stock, if this factor is changed then the current and historical prices will be wrong.

Consider a method to adjust the prices by the factor difference in order to try to keep historical information?

kriddles:
Initial commit for review and comments

Fixes #785
Saves qu_factor_purchase_to_stock to the stock and stock_log table
On Purchase, if the product qu_to_stock is different than the qu_to_purchase then the user will be allowed to edit the qu_factor_purchase_to_stock for that individual purchase.

Fixes #204 
Add product barcode details is available on the product form page.
On Purchase, if a barcode is inputted then it will attempt to pull the new barcode details to populate the purchase page information overriding the product defaults for qu_factor_purchase_to_stock and shopping location .

Also fixes #565.

Remaining items
- [x] Use new qu_factor_purchase_to_stock in stock and stock_log table for recipe costs
- [x] Better couple the new barcode details with current barcode input
- [x] Handle making sure barcodes are unique
________________________
New TODOs after the discussion below:
- [x] Migrate all current prices to be related to 1 QU stock
  - [x] Remove all `qu_factor_purchase_to_stock` conversion where then not longer needed
  - [x] Make sure that on purchase the correct price is used (no problem to change the API to accept only the price related to the stock QU)
  - [ ] Use the productamountpicker on the purchase/consume/inventory page to allow any related QU to be used while always prefilling the products default QU
- [x] Migrate current barcode(s) (comma separated in the `barcodes` field of the `products` table) to the new table for barcodes
  - [ ] Use the productamountpicker on the new barcode details page
  - [x] The barcode input for create or edit is not setup to receive a barcoded scanned input
  - [x]  Make the Barcode create or edit a popup
- [x] Don't allow to change the product stock QU when editing a product and when the product was already used (means when it has any `stock_log` entry)


berrnd:
Thanks - haven't tested it, just some comments:

>  Saves qu_factor_purchase_to_stock to the stock and stock_log table

When the purchase QU of the product has changed, the price will still be maybe wrong, the factor alone doesn't say much...

> On Purchase, if the product qu_to_stock is different than the qu_to_purchase then the user will be allowed to edit the qu_factor_purchase_to_stock for that individual purchase.

Is this really practical to always/when needed think about the factor manually? There is already the ["productamountpicker"](https://github.com/grocy/grocy/blob/master/public/viewjs/components/productamountpicker.js), currently only used on the recipe ingredients page, which does the amount conversion handling and allows to select all QUs which directly (product overrides) or indirectly (global conversions) have conversions for the product...

>  Add product barcode details is available on the product form page.

I think two inputs/places for barcodes will be confusing. Probably the best way would be to create a separate table for all barcodes, also to ensure that they are unique is then easier. UI-wise I would say the new table should be enough (showing the edit/add page in a popup instead of switching pages).

migrations/0102.sql
```diff
@@ -0,0 +1,19 @@
+ALTER TABLE stock_log
+ADD qu_factor_purchase_to_stock REAL NOT NULL DEFAULT 1;
```
berrnd, line 2: Adding a column with a `NOT NULL` constraint to a table with existing data is not possible in SQLite...

kriddles:
> Thanks - haven't tested it, just some comments:
> 
> > Saves qu_factor_purchase_to_stock to the stock and stock_log table
> 
> When the purchase QU of the product has changed, the price will still be maybe wrong, the factor alone doesn't say much...
> 
> > On Purchase, if the product qu_to_stock is different than the qu_to_purchase then the user will be allowed to edit the qu_factor_purchase_to_stock for that individual purchase.
> 
> Is this really practical to always/when needed think about the factor manually? There is already the ["productamountpicker"](https://github.com/grocy/grocy/blob/master/public/viewjs/components/productamountpicker.js), currently only used on the recipe ingredients page, which does the amount conversion handling and allows to select all QUs which directly (product overrides) or indirectly (global conversions) have conversions for the product...
> 

I do not want to suggest changing the QU, such as from Pack to Piece.
Only the QU Factor, such as today I bought a Pack of 12 Pieces instead of a Pack of 18 Pieces.


> > Add product barcode details is available on the product form page.
> 
> I think two inputs/places for barcodes will be confusing. Probably the best way would be to create a separate table for all barcodes, also to ensure that they are unique is then easier. UI-wise I would say the new table should be enough (showing the edit/add page in a popup instead of switching pages).

I agree and agree with your best way ideas. 
This is only a proof of concept with the 2nd checkbox on the remaining items to address this.
I am just not yet familiar with the tagManager to change over the management.

migrations/0102.sql
kriddles: Hmm, there were no sql errors and it looks to have been added in my dev db?
I did not test adding a null yet though.

berrnd:
> I do not want to suggest changing the QU, such as from Pack to Piece.

Then when we say the QU stock is "better" and _should never_ be changed, maybe it's easier in general to just relate all prices everywhere to 1 QU stock (instead to 1 QU purchase as of now)?

migrations/0102.sql
berrnd: Yes, sorry, seems to work - did only a quick test via my preferred SQLite editor where it did not work, but seems to work when executed through PHP (whatever the difference is)...

kriddles:
> > I do not want to suggest changing the QU, such as from Pack to Piece.
> 
> Then when we say the QU stock is "better" and _should never_ be changed, maybe it's easier in general to just relate all prices everywhere to 1 QU stock (instead to 1 QU purchase as of now)?

Yes, I agree the QU stock should never be changed. Perhaps I'll even add some comments in code to that effect. 

I know there was a request about products being deleted which causes issues due to now missing rows/ids. Its most likely the same issue if a QU is deleted. Perhaps they should instead be archived or deactivated.

I have quite a few commits coming for review to help with the pricing. Right now I am looking just to keep the stock price as the QU Purchase as is currently setup and just provide more price hint information throughout.

kriddles:
Well... here you go. Review at your leisure...

Instead of removing the core search and functionality of the product.barcode I just adjusted the new product barcode table to add/remove from the product.barcode and then removed the tagManager from the product form page.

Also updated quite a bit to help show the qu_factor and price information in several place.

kriddles:
Some screenshots to help with review and double check myself

image
- [ ] I need to round stock value

image

image
- [ ] The barcode input for create or edit is not setup to receive a barcoded scanned input

image
If the qu_purchase is different than qu_stock then the Amount will include the qu_factor information

Recipe costs updated along with the hint.

I noticed the new hint icon does not always put the hints as hover on several pages.

kriddles:
Branch added for rounding and include currency for stock value in product card.

kriddles:
- [ ] Add qu_factor_purchase_to_stock to the stock edit entry page

kriddles:
Branch added for product card stock amount to include factor_purchase_amount only if used just like on the stock overview page.

image

kriddles:
Commit added for stock entries page updates

berrnd:
I just gave it a try, basically works, but I think especially the price factor thing on the purchase page is unhandy / looks complicated (was also always hidden, don't know why, just made it visible manually via the browser console for testing)...

About all the `qu_factor_purchase_to_stock` conversions/calculations:
If we change so much about price handling now, I think it would be much easier if we always store all prices related to 1 QU stock. To make it more clear that this can never be changed, make the QU stock dropdown read only when editing a product. Based on the QU conversions we could always convert the price to all related QUs which have conversions defined. (For example a little dropwn on the product card, defaults to the products QU purchase.) What do you think?

About the price factor:
Is it really ideal to always prefill the last factor used on the purchase page? The factor defined on the product is then more or less useless. Also still would prefer to just use the productamountpicker on the purchase page, is much more "user friendly", I would say.

About "barcode details":
Seems that the new barcodes did not yet get looked up in the productpicker (?). And we should definitely combine them, just transfering the old barcodes into the new table and only using that. (If it should be no breaking change, the old `barcode` field of the `products` table would be needed to "virtually" stay...)

Sorry if this was not intended to be tested yet, just wanted to have a look. :)

berrnd:
> Also still would prefer to just use the productamountpicker on the purchase page

Would also fit for the new barcode details page: "I want to attach this barcode to x pieces (and that store)" can be easier/clearer defined that way I think...

kriddles:
> Sorry if this was not intended to be tested yet, just wanted to have a look. :)

Thank you for looking, it is in a state to review. The only outstanding issue I had was changing the add barcodes to instead be a popup and including on the barcode form the barcode scanned logic.

And I expected changes after review :)

> I just gave it a try, basically works, but I think especially the price factor thing on the purchase page is unhandy / looks complicated (was also always hidden, don't know why, just made it visible manually via the browser console for testing)...

The factor to stock is only going to be shown if the product is setup with different QU Purchase vs QU Stock. In the demo, it should only be visible for Eggs.

 
> About all the `qu_factor_purchase_to_stock` conversions/calculations:
> If we change so much about price handling now, I think it would be much easier if we always store all prices related to 1 QU stock. To make it more clear that this can never be changed, make the QU stock dropdown read only when editing a product. 

Changing the price stored in the stock and stock_log tables to be QU stock would prevent quite a few places from doing a calculation and allow better visualizing on the stock entries table. I agree that would be a good change. It would require a conversion script for all current stock/stock_log entries to adjust the prices for those entries.

To prevent setup issues, perhaps making the qu stock dropdown read only when editing a product - (only if there has been a stock_log entry?) otherwise it can be changed. For example, a user could click the wrong dropdown for a new product and click save... then not be able to edit that correctly.

>Based on the QU conversions we could always convert the price to all related QUs which have conversions defined. (For example a little dropwn on the product card, defaults to the products QU purchase.) What do you think?
> There would then be no need to store the factor on `stock` and `stock_log`...

I'm not sure I understand, can you provide an example?

> About the price factor:
> Is it really ideal to always prefill the last factor used on the purchase page? The factor defined on the product is then more or less useless. Also still would prefer to just use the productamountpicker on the purchase page, is much more "user friendly", I would say.

Prefill the last factor was important since the purchase page prefills the last price which are important together.

> About "barcode details":
> Seems that the new barcodes did not yet get looked up in the productpicker (?). And we should definitely combine them, just transfering the old barcodes into the new table and only using that. (If it should be no breaking change, the old `barcode` field of the `products` table would be needed to "virtually" stay...)

I just made it work without having to adjust the products table removing the barcode column and having to figure out what all would need to be fixed for that...
The purchase page is the only current place that uses the new barcode details. It does lookup based on the barcode that is inputed in the productpicker for the purchase page.

kriddles:
> > Also still would prefer to just use the productamountpicker on the purchase page
> 
> Would also fit for the new barcode details page: "I want to attach this barcode to x pieces (and that store)" can be easier/clearer defined that way I think...

I'll take a look at it.

Also, I'll take a look at adjusting the stock price to instead be QU stock instead of QU purchase

berrnd:
> >Based on the QU conversions we could always convert the price to all related QUs which have conversions defined. (For example a little dropwn on the product card, defaults to the products QU purchase.) What do you think?

> I'm not sure I understand, can you provide an example?

On the productcard for example where you added the purchase QU amount in braces, this QU could just be some (inline" dropdown where all related QUs could be selected, or for example display just all other QU values in a tooltip or something like...

Great, then let's do this all here :), of course I can also help (and will not work on anything else inbetween to avoid conflicts, as this will most likely require changes nearly everywhere...).

berrnd:
I've added some TODO points above to not forget anything.

Also let's don't invest to much time in backwards compatibility, I don't have a problem to bump the major version after this, would also fit with the great new style improvements by @zsarnett... :)

zsarnett:
@berrnd 

When are you wanting the next release to finish up by? Do you have a dev cycle you keep too? I have a few more things I am working on currently for UI updates that may be best to get in before the next release if a major version bump

kriddles:
> @berrnd
> 
> When are you wanting the next release to finish up by? Do you have a dev cycle you keep too? I have a few more things I am working on currently for UI updates that may be best to get in before the next release if a major version bump

This has evolved now to be touching quite a bit so may need some time to check and recheck things and it may even effect @Forceu. 

I think you have time @zsarnett  :) based on @berrnd [roadmap](https://github.com/grocy/grocy#roadmap)

berrnd:
Yes, no hurry, I will personally have not that much to work on grocy for the next couple of weeks ... based on the last releases with bigger features changes, I would say maybe about the end of July or so - but as always, it's ready when it's ready (and when everyone who works on changes says it's ready)... :)

This here is about to change everything related to price and barcode handling internally, so let's better take the time and make sure that everything works, that no prices will be migrated wrong, no barcodes get lost, and so on... :)

zsarnett:
Sounds good. End of July? Oh big changes are coming ;)

kriddles:
A few progress commits

- Updated price to be 1 QU to stock
- Removed product barcode column and now using just the product_barcodes table

and since removing the product barcode column requires rebuilding the table...
- Products can be deactivated now... #565

kriddles:
rebased to master

kriddles:
Rebased and made product form updates for barcode and qu-conversion to be embedded popups

kriddles:
@berrnd  Can you take a look at this recent branch 4eca9a9 for the barcode scanner?
I don't have one setup to test and also the css may need a look at for it too.

image

berrnd:
About the CSS:
The `barcodescanner-input` needs to wrapped inside an `input-group` - like here:
https://github.com/grocy/grocy/blob/98f214e9f14d5197f85aaefd38de856bff5da208/views/barcodescannertesting.blade.php#L28-L33

kriddles:
Fixed d2964b2 barcode scanner css

Included fix for stock overview refresh in d61f353

Added work in progress commit for the productamountpicker

kriddles:
A fix done for 050a3d2

Also, @berrnd what do you think instead if can edit product stock QU if stock is currently at 0?
instead of checking stock_log entry 559717e
"Don't allow to change the product stock QU when editing a product and when the product was already used (means when it has any stock_log entry)"

berrnd:
> instead if can edit product stock QU if stock is currently at 0?

I would say then we have the same problem about historic prices, as long as we not also save the used QU stock in `stock_log`...
It should be somehow possible to migrate all data correctly when changing the QU stock, I think - maybe we just don't allow this for now and implement such a migration in a second step (so not part of this here)...

kriddles:
> > instead if can edit product stock QU if stock is currently at 0?
> 
> I would say then we have the same problem about historic prices, as long as we not also save the used QU stock in `stock_log`...
> It should be somehow possible to migrate all data correctly when changing the QU stock, I think - maybe we just don't allow this for now and implement such a migration in a second step (so not part of this here)...

Ok, I updated the commit to instead check against productDetails.last_purchased so that editing cannot occur after the first purchase occurs. 

Yes, there should be a way to migrate the historical information... I don't know how often this scenario would really occur so don't know how important this will be. I guess we'll find out the first time someone complains about not being able to change the qu_stock lol...

Wordna1:
Any ETA for this feature?  We are really looking forward to #204 which is apparently related

kriddles:
> Any ETA for this feature? We are really looking forward to #204 which is apparently related

I'll take a look at where I left off with the product amount picker updates which was causing some issues and see if I can get it ready in the next couple of weeks. I've had quite a bit going on recently that took me away from making updates.

kriddles:
Hi Bernd,

I rebased this to master and removed my work in progress commit for the product amount picker changes.
When you have time, this branch can be reviewed to be pushed to master with all the changes. Feel free to make changes you need to for this chain.
I'll make a separate commit with my work in progress which hopefully I'll have time for soon or you can take it and update as you wish.

bmalinconico:
Awesome & Congrats!

Life is still too crazy to update PantryParty so please post a warning when you release it.

berrnd:
> Life is still too crazy to update PantryParty so please post a warning when you release it.

Have never planned releases :D, but [let's say](https://www.reddit.com/r/grocy/comments/hxeg31/when_can_we_expect_an_update/fz6apyt/) to get grocy 3.0 (and hopefully MUCH more, but still also here a lot of real life todo) ready as a christmas present. So still some time left.

The API has the [`/system/info` endpoint](https://demo.grocy.info/api/system/info) which you could use to prevent the API connecting to an unknown grocy version.

hannahswain:
What an amazing set of changes! Thank you for all of your hard work!

Issue #805, opened 2020-04-28
Add SQLite version check to PrerequisiteChecker


berrnd:
_Originally posted by @bendauphinee in https://github.com/grocy/grocy/issues/104#issuecomment-620270089_

> Might be worth adding a prerequisite check script that runs before anything else to make sure the server has the right versions of everything. I've run into the same issue, and it took me a few minutes to locate this thread via Google. CentOS 7, SQLite 3.7.17

PR #806, opened 2020-04-28
Added SQLite check for #805

Forceu:
Added SQLite check for #805


Forceu:
Changed in https://github.com/grocy/grocy/pull/806/commits/06bc8c271e51f449c3ccd96e43f6478175f39c51

PR #811, opened 2020-05-01
[MISC] - Added DEVELOPMENT.md to help contributers to set up the dev …

NokiDev:
Hello @berrnd , I added a .md file precising, how to set up dev environnment. We sure can add a bit more of information. 

There is a part missing here, it's how to actually view the HTML pages. Blade component don't render on my computer (blank page), if you have an insight about how to setup this, I'll add it to the doc.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,0 +1,48 @@
+
+This page is here to help you setup your devel environnment for starting to contribute to grocy.
+Feel free to share your setup if you think it can increase developper productivity.
+
+## Environment Set up
+
+Install PHP7 : 
+ - [Windows](https://windows.php.net/download#php-7.4)
+ - [Linux](https://www.php.net/distributions/php-7.4.5.tar.bz2) but using your distro package manager will do the job.
+
+Install Composer :
+
+- [Windows](https://getcomposer.org/doc/00-intro.md#installation-windows)
+- [Linux](https://getcomposer.org/doc/00-intro.md#downloading-the-composer-executable) but using your distro package manager will do the job
+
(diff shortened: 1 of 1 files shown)


berrnd:
Thanks, maybe good to have an example dev setup description, but I think that depends a lot personal preference - some like to just use a plain text editor, others (me) use VSCode, others use maybe a even bigger IDE...

I could also describe my setup more detailed if that helps - I don't install PHP, composer or node/npm/yarn on the system for example, because I need to use different PHP versions for different projects and also use different machines...

>  Blade component don't render on my computer (blank page),
I think there must be a PHP error somewhere then...

NokiDev:
> I could also describe my setup more detailed if that helps 
Why not, surely install PHP system wide is not good. I give the path to VSCode, but there should be some version management tools for this. 
Would be great for the community, if you share your setup. Less waste of time for setup and more for developing.

> I think there must be a PHP error somewhere then...
Well, I'll dig into then.

NokiDev:
Find out that I was missing viewcache directory, and node_module installation. 

Added thoose steps in the doc

berrnd:
Thanks again, but I decided to not add this, see https://github.com/grocy/grocy/pull/904#issuecomment-675562451.

NokiDev:
@berrnd Understood, but I clearly can't see your point. 
The steps described in the doc, are not "personal preferences" it's just how the project works at the bare minimum.
I agree people can do whatever they want but for someone, that usually do not use PHP, it's more clear to have at least a doc explaining how to setup a dev-env that work. 
The dev is able to know how it work and start contribute or customize the setup, with docker or a php version manager, or whatever.

Anyway, that's just my two cents. 
Keep up the great work. 

See ya.

Issue #799, opened 2020-04-26
Upgrade to Quagga2 (again)


berrnd:
References https://github.com/grocy/grocy/issues/749#issuecomment-619499413.

ericblade:
If you find any problems, or need any assistance, give me a yell, thanks :)

ericblade:
just dropping a line to let you know that it's up to 1.0.1 now, it should work just the same, but there's been a big amount of internal changes and a slight difference that impacts how people using it with Image and File inputs rather than LiveStream inputs might see it working.

it should still just drop in and replace the old quagga

TheBird956:
What would be the testing requirements for this? I might do the upgrade and see what happens. I have a few issues that I want to know if the updates fixes (camera not focusing right and not scanning, returning the wrong code)

ericblade:
Most of what I've done with Q2 so far has been code modernization, maintenance, updating dependencies, things of that nature, so just replacing the original with the fork probably won't affect much of that.   As far as focus and accuracy goes, there are some things that can be done in the user code (ie, grocy) -- there are various ways to implement error rejection, as well as having camera selection (i'm not familiar with grocy's feature set, so i'm not sure if either are a thing it does already).  Although I've discovered that some phones just plain don't support auto-focus in the browser/webview camera capture, i've also discovered that many phones have different logical or physical cameras that produce very different results.   (ie, my LG G6 has two rear cameras, one is a wide-angle, which is useless for scanning barcodes becuase it fisheyes the image... and it's also the one that the browser selects by default)

PR #844, opened 2020-05-11
Additional configuration options for Quagga2

andrelam:
Hello, @berrnd.

This is an initial commit for review and discussion.

When trying to make the barcode scanner work on Grocy on my iPhone, I had to tweak some of the Quagga settings to achieve optimal results, which differ from the default. For example, to me I need to set halfSample to true, patchSize to small and frequency to 5. These took lots of time to tweak, and I thought a PR enabling these configurations on config.php along with numOfWorkers might be valuable to some users. Also, to enhance performance, I added an option to enable/disable the debug options for Quagga.

### DIFF
--- a/config-dist.php
+++ b/config-dist.php
@@ -126,9 +126,14 @@ DefaultUserSetting('auto_reload_on_db_change', true);
 # Show a clock in the header next to the logo or not
 DefaultUserSetting('show_clock_in_header', false);
 
-# Component configuration
+# Component configuration for Quagga2 - read https://github.com/ericblade/quagga2#configobject for details
+# Default for Grocy is as below
+# For iPhone 7 PLus, halfsample = true, patchsize = small, frequency = 5 yields very good results
 DefaultUserSetting('quagga2_numofworkers', 4);
-
+DefaultUserSetting('quagga2_halfsample', false);
+DefaultUserSetting('quagga2_patchsize', 'medium');
+DefaultUserSetting('quagga2_frequency', 10);
+DefaultUserSetting('quagga2_debug', true);
 
 # Feature flags
 # grocy was initially about "stock management for your household", many other things
--- a/public/viewjs/components/barcodescanner.js
+++ b/public/viewjs/components/barcodescanner.js
@@ -69,25 +69,25 @@ Grocy.Components.BarcodeScanner.StartScanning = function()
 			}
 		},
 		locator: {
-			patchSize: "medium",
-			halfSample: false,
+			patchSize: Grocy.UserSettings.quagga2_patchsize,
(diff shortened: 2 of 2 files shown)


berrnd:
Thank you.

Should make no difference, but currently we don't use Quagga2 but again QuaggaJS, see #799.

ericblade:
hi there :) present maintainer of Quagga2 here :)  There are quite a few configuration settings, such as those, which can greatly affect your scanner performance.    Unfortunately, it's not entirely clear, exactly how these various tweaks can affect how well a scan works, given all the different other variables.

In general, higher resolution camera will give you a better image to work with, and increasing the patchSize and turning on halfSample seems to be better on higher resolution cameras than on lower.   Lower resolution cameras seem to do better with much smaller patch and halfSample off.  

Per the README, which mostly comes from the original author of the library:

[quote]The `halfSample` flag tells the locator-process whether it should operate on an
image scaled down (half width/height, quarter pixel-count ) or not. Turning
`halfSample` on reduces the processing-time significantly and also helps
finding a barcode pattern due to implicit smoothing.
It should be turned off in cases where the barcode is really small and the full
resolution is needed to find the position. It's recommended to keep it turned
on and use a higher resolution video-image if needed.

The second property `patchSize` defines the density of the search-grid. The
property accepts strings of the value `x-small`, `small`, `medium`, `large` and
`x-large`. The `patchSize` is proportional to the size of the scanned barcodes.
If you have really large barcodes which can be read close-up, then the use of
`large` or `x-large` is recommended. In cases where the barcode is further away
from the camera lens (lack of auto-focus, or small barcodes) then it's advised
to set the size to `small` or even `x-small`. For the latter it's also
recommended to crank up the resolution in order to find a barcode.[/quote]

personally, in my app, to avoid having to present a user with a bunch of options, i set 800x600 camera res, and use some settings that seem to work well for me.  there's probably some ability to greatly increase scanning ability and speed if you tweak all the different things.  

i've also considered maybe trying different default settings based on resolution, or some sort of auto-adjustment of the settings, as well to try to scan for what works based on the image presented.. but that stuff is all future, if it becomes a thing

andrelam:
Thanks for the additional information, @ericblade. I agree that the tweaking will take several tries until well-refined. For me, on an iPhone 7s Plus, the best settings were those I mentioned above (halfSample = true, patchSize = small, and frequency = 5. With these settings, I can disable all debugging options, which improve battery consumption as well.

Maybe creating a page on Grocy or Quagga2 documentation to list tried configurations shared by users based on the camera model for each mobile phone would be helpful.

berrnd (approved):

ericblade:
Cheers.  If you run into any issues, happy to work together to try to solve

PR #849, opened 2020-05-14
Added route caching

Forceu:
Add a route cache, which can speed up the routing a little bit


kriddles:
Similar... was recently added and then reverted but I don't recall why it was reverted but there are comments found here about it
https://github.com/grocy/grocy/commit/59cd4dbac47c5ceb7a28de0f14cc133390612a72#commitcomment-38477633

Forceu:
Ah thanks, I was 90% sure that I already saw it here, but I couldn't find it in the source, so I added it

Issue #794, opened 2020-04-23
Unable to enter more than 999 quantity in recipes


crazyelectron-io:
When entering an ingredient for a recipe, any amount over 999 is reverted to 1.
For example, I have a product 'beaf' which is bought and used by a quantity 'grams', when entering 1000 or more in the amount field it always reverts to 1.

berrnd:
Strange, whatever that is...

The field has the max value set to `999999`, but however when for example entering `123456`, `123.456` is send in the recipe ingredient API update request, so clearly a frontend / JS problem...

crazyelectron-io:
When checking the database, the field for this ingredient contains the value '1,000', with a comma as decimal point. Looks like language related formatting gets in the way. My Windows system is configured for Dutch number formatting.
Needs some more investigation...

Wordna1:
I am running into the same issue

PR #876, opened 2020-05-29
Make product amount picker locale independent

m-byte:
Since the value of $('#amount') will be written to the database it should not be locale dependent. This code also limits the result to a maximum of 4 digits but always uses a dot as decimal separator.

### DIFF
--- a/public/viewjs/components/productamountpicker.js
+++ b/public/viewjs/components/productamountpicker.js
@@ -83,7 +83,7 @@ $(".input-group-productamountpicker").on("change", function()
 		$("#qu-conversion-info").text(__t("This equals %1$s %2$s in stock", destinationAmount.toLocaleString(), destinationQuName));
 	}
 
-	$("#amount").val(destinationAmount.toLocaleString({ minimumFractionDigits: 0, maximumFractionDigits: 4 }));
+	$("#amount").val(destinationAmount.toFixed(4).replace(/0*$/g,''));
 });
 
 $("#display_amount").on("keyup", function()


berrnd:
Thank you! :)

(Maybe this is related to #794.)

m-byte:
Yes, as far as I understand it, it is. I will try later today to check.

m-byte:
Just checked. It does in fact fix #794 as well.

Wordna1:
#794 is impacting most of our products and recipes because we primarily use metric and thus ml to L and g to kg is not working.

It looks like this bug fix is almost ready for release?  Are we able to temporarily solve this issue by adjusting our locale/language settings?  If so, what would we change and would it be a browser or operating system level change?

berrnd:
Feel free to try it by just applying what this pull request suggests: https://github.com/grocy/grocy/pull/876/files

(Sorry that my work on grocy is currently paused, will hopefully change again in a not so many weeks...)

Wordna1:
> Feel free to try it by just applying what this pull request suggests: https://github.com/grocy/grocy/pull/876/files
> 
> (Sorry that my work on grocy is currently paused, will hopefully change again in a not so many weeks...)

So far, so good!  I will add a comment if we run into any additional challenges, but this pull request seems to resolve the related issues.

berrnd:
Thanks for testing!

> Will we still be able to upgrade normally or has applying this pull request complicated applying the next update?

Will be no problem, the update justs overwrites everything (except the data directory)...

PR #892, opened 2020-06-14
Fix amount in recipes ingridients

migolovanov:
When amount in recipie ingridients is over 1000, it is sent to API as `"amount": "1 000"` and is saved in DB the same way. It displays in UI then as 1 and seems that counts same way when calculated stock items for recipie.
Removing spaces in amount value before sending it to API fixes this issue.


berrnd:
Thank you.

But most probably already fixed by #876 (references #794).

But I'll leave this until reviewed (sorry that my work on grocy is currently paused, hope to change that again soon).

migolovanov:
Thank you, have missed that issue and pull request. I've checked - it works, and seems to be better solution, so this pull request should be closed as duplicate.
I appreciate your work on grocy and hope you will continue it in future.

Issue #863, opened 2020-05-20
How to setup development environment?


TheBird956:
Hi!

I am trying to set up a dev environment, but I can't seem to get it right. I am running on Windows 10.

I first try going the docker route but I quickly got to permission errors (need to assign www-data owner on all files, can't mount aw volume because of that and `chown` is taking a very long time, don't want to troubleshoot docker on windows)

I used Xampp as a fallback but whatever I do I am stuck with this 404 error:
```
404 Not Found
The application could not run because of the following error:

Details
Type: Slim\Exception\HttpNotFoundException
Code: 404
Message: Not found.
File: E:\xampp\htdocs\grocy\vendor\slim\slim\Slim\Middleware\RoutingMiddleware.php
Line: 93
Trace
#0 E:\xampp\htdocs\grocy\vendor\slim\slim\Slim\Middleware\RoutingMiddleware.php(59): Slim\Middleware\RoutingMiddleware->performRouting(Object(Slim\Http\ServerRequest))
#1 E:\xampp\htdocs\grocy\vendor\slim\slim\Slim\MiddlewareDispatcher.php(140): Slim\Middleware\RoutingMiddleware->process(Object(Slim\Http\ServerRequest), Object(Slim\Routing\RouteRunner))
#2 E:\xampp\htdocs\grocy\vendor\slim\slim\Slim\Middleware\ErrorMiddleware.php(107): class@anonymous->handle(Object(Slim\Http\ServerRequest))
#3 E:\xampp\htdocs\grocy\vendor\slim\slim\Slim\MiddlewareDispatcher.php(140): Slim\Middleware\ErrorMiddleware->process(Object(Slim\Http\ServerRequest), Object(class@anonymous))
#4 E:\xampp\htdocs\grocy\vendor\slim\slim\Slim\MiddlewareDispatcher.php(81): class@anonymous->handle(Object(Slim\Http\ServerRequest))
#5 E:\xampp\htdocs\grocy\vendor\slim\slim\Slim\App.php(215): Slim\MiddlewareDispatcher->handle(Object(Slim\Http\ServerRequest))
#6 E:\xampp\htdocs\grocy\vendor\slim\slim\Slim\App.php(199): Slim\App->handle(Object(Slim\Http\ServerRequest))
#7 E:\xampp\htdocs\grocy\app.php(68): Slim\App->run()
#8 E:\xampp\htdocs\grocy\public\index.php(27): require_once('E:\\xampp\\htdocs...')
#9 {main}
```

[...]

PR #904, opened 2020-07-03
Docker development environment and start script

TheBird956:
With this, the only thing needed to launch a development environment is a working Docker setup and be able to run bash scripts.

Starting the environment is done by running `./start-dev.sh`. All the dependencies will be installed and the project directory will be mounted inside the image.

There are two arguments that can be passed to the script:
- -u | --uid        Specify a different UID to pass to the user www-data in the container (defaults to the current user ID)
- -c | --clean    Delete all cache and vendor files

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,77 @@
+FROM php:7.3-apache
+
+# Enable Apache Rewrite + Expires Module
+RUN a2enmod rewrite expires
+
+# Install dependencies
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    unzip \
+    libfreetype6-dev \
+    libjpeg62-turbo-dev \
+    libpng-dev \
+    libyaml-dev \
+    libzip4 \
+    libzip-dev \
+    zlib1g-dev \
+    libicu-dev \
+    g++ \
+    git \
+    cron \
+    vim \
+    curl \
+    unzip \
+    gnupg2 \
+    && docker-php-ext-install opcache \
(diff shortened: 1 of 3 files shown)


berrnd:
Thank you, but I think development (environments) heavily depend on personal preferences ... so I don't know if this would anybode use "out of the box"...

See #811 and #863.

TheBird956:
Having multiple ways to install a development environment isn't really a bad thing, and if potential contributors can have an easy way to install the environment and just start coding without having to install multiple libraries on their system, I think it would encourage them more to contribute. With a dockerfile it is also possible to control a bit more what is in the environment since it is unified and it reduces the amount of "but it works on my machine".

For #863, I'm the one who opened that issue and still cannot make it work on Windows (I must have a setting somewhere that is doing this or a conflict with another piece of software). At that point, it isn't a preference, it's just making it work. I had to switch my OS to Ubuntu to make it work (partially because I hate Docker on Windows) and to build this image.

From my experience, once people are introduced to this kind of environment setup (project files mounted in a development environment built with Docker) it's generally very well received. It also opens the door to more code automation in the future.

berrnd:
Thanks again, but I decided to not merge this.

I'm still thinking that dev environments heavily depend on personal preference, what someone finds useful and easy to set up, others don't. For me this feels much more complicated than just checking out the repo, opening it with VSCode while having some extensions installed which do the rest...

So I think we have enough shared ideas/experiences on how we work on grocy development here and in the referenced issue which others may find useful (or not).

Issue #828, opened 2020-05-04
Auth proxy support


yegle:
I took a quick look at config-dist.php and don't see anything related to auth proxy and thus this FR.

An auth proxy is an HTTP reverse proxy sits between the client and the web service. The proxy will authenticate user (e.g. requiring Google login) and only forward request when the authentication passes. This is a great way to expose your self hosted service on the public Internet while ensuring only whitelisted person can access.

See https://grafana.com/docs/grafana/latest/auth/auth-proxy/ for more context.

I'm currently using https://pomerium.io as my auth proxy but there are other similar auth proxies out there.

berrnd:
Duplicate of #627.

berrnd:
This is now done by #921 

Excerpt from the changelog:
> ### New feature: Reverse proxy authenticaton support
> - New `config.php` settings `AUTH_CLASS` and `REVERSE_PROXY_AUTH_HEADER`
> - If you set `AUTH_CLASS` to `Grocy\Middleware\ReverseProxyAuthMiddleware` and your reverse proxy sends a username in the HTTP header `REMOTE_USER` (header name can be changed by the setting `REVERSE_PROXY_AUTH_HEADER`), the user is automatically authenticated (and will also be created, if not already present)
> - (Thanks @fipwmaqzufheoxq92ebc for the initial work on this)

Issue #305, opened 2019-07-08
LDAP authentication


histal89:
Would be nice to authenticate with ldap users.

berrnd:
Already requested in #207, but I let this open as it is specific about LDAP...

berrnd:
This is now done. 🙂
(Have tested it with a Microsoft Windows Domain / Active Directory.)

Changelog excerpt:
> ### New feature: External authentication support
> - New `config.php` setting `AUTH_CLASS` to change the used authentication provider
> - Via LDAP
>   - New `config.php` settings `LDAP_DOMAIN`, `LDAP_ADDRESS` and `LDAP_BASE_DN`
>   - If you set `AUTH_CLASS` to `Grocy\Middleware\LdapAuthMiddleware`, users will be authenticated against your directory (and will also be created (in grocy), if not already present)

Issue #207, opened 2019-04-10
Feature Request: External Authentication Source


davmin:
I'd like to be able to authenticate my grocy users against an external source.  The method/protocol chosen isn't super important IMO as most people who have central authentication can support any of them with a bit of effort. LDAP(S), OIDC, and SAML are options.

Additionally some user management in the form of groups and permissions would be nice. For instance I may want a subset of users to be able to access the chores but not anything recipe related. Or I may want a subset of users to be able to add recipes to the cart but not edit recipes or ingredients.

berrnd:
Let's split the permisson part into the new request #263 and keep this about adding external authentication source support.

If anyone wants to propose which roles/groups with which permissions would be needed, feel free to do so in #263.

berrnd:
#921 is now merged, which also brings that the authentication class is now abstracts.

Means, any class that implements [`Grocy\Middleware\AuthMiddleware`](https://github.com/grocy/grocy/blob/master/middleware/AuthMiddleware.php) can now be used for (custom) authentication.

https://github.com/grocy/grocy/blob/31c412a28c44ab28ac1183f88a539c42af92c21c/config-dist.php#L69-L71

Permission things are still open and will be tracked in #263, as mentioned above.

PR #921, opened 2020-07-21
Refactor Authentication and add proxy-authentication

### LINKED ISSUE #627: Feature Request: Use username from reverse proxy
Hello,

I want to use grocy behind a reverse nginx proxy, similar to other services that I use. For my other services, I also use the nginx PAM module for authentication to keep the password management for all services in one place.

It would be great if grocy would be able to use the username supplied in a request header.
This request seems related to #246 , but instead of using the first user in the database, it would probably need to create users if they do not exist.

fipwmaqzufheoxq92ebc:
Hi,
I refactored the Authentication-System:
I created a new abstract class AuthMiddleware which does common tasks (e.g. check if auth is disabled or in demo-mode).
Currently, there are four Subclasses:
Session-, ApiKey- and Proxy-, DefaultAuth- Middleware

DefaultAuth checks for an Api-Key or Session.	
Note: The old implementation used Session for UI and Api-Key or Session for the API. Now, there is only one setting for both, so you can access both with Api-Key or Session. I don't think that's a problem.

The Proxy-Authentication uses the username from the reverse Proxy.
e.g. for nginx:
`fastcgi_param  HTTP_X_USERNAME <username>;`
where <username> is the username to use.

A new Setting 'AUTH_CLASS' configures which Middleware to use.
DefaultAuthMiddleware is the default.

When Proxy-Auth is used, I hide the "Logout", "Change password" and "Manage users" buttons.

Fixes #627   (and #828)
Also might help for #305, #207


berrnd:
Cool, thanks a lot! :)

Will review this as soon as possible, but currently I cannot really work on grocy - will hopefully change in the next couple of weeks..

berrnd (approved):
I did some extensive tests with that, everything basically worked - thanks again. :)

I noted and fixed the following:
- Undefined constant warnings when `MODE` was not `production` or when `DISABLE_AUTH` was set to `true`
- `AuthMiddleware` always redirects to `/login` when unauthenticated, this is ok for frontend routes, but not for the API (which should return a 401 response instead) - I tried to distinguish that by the URL for now in `AuthMiddleware`

knnnrd:
Hi @berrnd, when will you tag a new release containing this PR?
Thanks

berrnd:
I never promise release dates, but see: https://www.reddit.com/r/grocy/comments/jvphp8

Issue #898, opened 2020-06-26
"Auto-reload on external changes" doesnt work as expected 


Dinth:
Ive got a large database with several hundreds of products and for past couple of months i've been experiencing serious problems with adding new products to stock - every time i've saved a change, i needed to wait literally a couple of minutes until Grocy becomes responsive again and i could start entering the next product. 
I just found out that disabling "Auto-reload on external changes" actually fixes the issue and i can start adding the next product few seconds after saving the previous one. I understand that Grocy may not know if the change was actually external and does its thing on every change, but in this circumstances maybe "Auto-reload on external changes" option should be renamed to something less confusing?

berrnd:
"Auto-reload on external changes" currently works by just checking the modified time of the database file once every minute.
If that locally saved date is different and if there is unsaved form data and the idle time is > 50 seconds, the page will be reloaded.

Code reference: https://github.com/grocy/grocy/blob/master/public/js/grocy_dbchangedhandling.js

Maybe this has nothing todo with that and is just a problem with loading large tables (which is currently not handled in the best possible way as the whole HTML markup for the whole table is generated server side...)?

Dinth:
How can i troubleshoot it? 
If im opening two tabs with Grocy, adding products in one of them, while the second one is just sitting in the background, with "Auto-reload on external changes" enabled, i can see that every time im saving a change and Grocy freezes for 2 minutes, the other tab also gets reloaded and has a spinning "reload" icon for whole duration of freeze.

berrnd:
If you disable auto reloading, the JavaScript part referenced above does (should do) nothing - so if it's only happening when enabled, strange - have no idea what this could be at the moment...

When it's only when you have a second tab open and that one is currently reloading, maybe that's the thing - depends not only on server power but also webserver/PHP configuration, maybe the (longer running) reload request of the second tab blocks further requests (of the first tab)...

Dinth:
>When it's only when you have a second tab open and that one is currently reloading, maybe that's the thing - depends not only on server power but also on webserver/PHP configuration, maybe the (longer running) reload request of the second tab blocks further requests (of the first tab)...
The thing is that it doesnt only happen when a second tab is open, i just wanted to use a second tab to check if Grocy is accessible in a different session when the first one freezes.

berrnd:
[...]

Issue #848, opened 2020-05-14
API calls to /stock/products/ are slow


Forceu:
I recently noticed that API calls to `/stock/products/` are rather slow. First I thought it might be because the routing is not cached, but adding 

```
$routeCollector = $app->getRouteCollector();
$routeCollector->setCacheFile(GROCY_DATAPATH . '/routecache');
```

to `app.php` did not increase the performance significantly, therefore the problem is probably somewhere else. My current database has 246 products, 261 stock entries and around 3000 stock logs. Deleting the stock logs did not increase the performance. On average a `/stock/products/` API call takes around 650ms, while a call like `/stock/shoppinglist/remove-product` takes only around 100ms.

Unfortunately I did not find any causes for the problem, do you have any idea how to improve the speed?

kriddles:
I assume you mean pulling an individual product detail with the api?
'/stock/products/{productId}' which calls '\Grocy\Controllers\StockApiController:ProductDetails

ProductDetails pulls from quite a few tables to build the product details information including a few calls to the stock_log table using different parameters which could probably be optimized. It is probably the heaviest api call.

If you're looking for dedicated product information then making a call to the product table would be the fastest. /objects/products/{id}

This recent commit looked to improve performance
1a5f3ce92628e81310d17f4ec1d6ad65282d0ea3

The other recent commits just add more information to product details such as the new feature shopping locations.

Forceu:
```
2020-05-14 17:27:46: Executing API call: objects/products
2020-05-14 17:27:46: Executing took 221ms
2020-05-14 17:27:46: Executing API call: stock/products/255/open
2020-05-14 17:27:46: Executing took 196ms
2020-05-14 17:27:46: Executing API call: system/info
2020-05-14 17:27:47: Executing took 167ms
2020-05-14 17:27:47: Executing API call: stock/products/255
2020-05-14 17:27:47: Executing took 764ms
2020-05-14 17:27:47: Executing API call: stock/products/255/add
2020-05-14 17:27:48: Executing took 873ms
2020-05-14 17:27:48: Executing API call: stock/shoppinglist/remove-product
2020-05-14 17:27:48: Executing took 184ms
2020-05-14 17:27:48: Executing API call: stock/shoppinglist/remove-product
2020-05-14 17:27:49: Executing took 179ms
2020-05-14 17:27:49: Executing API call: stock/shoppinglist/add-product
2020-05-14 17:27:49: Executing took 188ms
[...]

PR #927, opened 2020-07-24
Improve Performance

fipwmaqzufheoxq92ebc:
Hi,

I tried to improve the Performance of the "Stock overview" page, and found some improvements:
1) I replaced some of the `FindObjectInArrayByPropertyValue`-calls by getting this values directly in the database query.
This also made the template-code more readable.

2) I optimized the "stock_current"-database-view:
I created 2 indices, and replaced a subquery with a (left-)join.
I'm not sure why, but this combination improved the speed significantly.
stock_current is used often:
I got a speedup from 1.26s to 0.13s on /stock/products #848
The view stock_missing_products, which uses stock_current, is now 25 times faster (from ~1250 ms to ~50ms) on my sample data.

Might be related to #898 (Note that JavaScript ("Datatables") also has some loading time. I think pagination could fix that.)

### DIFF
--- a/controllers/StockController.php
+++ b/controllers/StockController.php
@@ -16,17 +16,13 @@ class StockController extends BaseController
 		$nextXDays = $usersService->GetUserSettings(GROCY_USER_ID)['stock_expring_soon_days'];
 
 		return $this->renderPage($response, 'stockoverview', [
-			'products' => $this->getDatabase()->products()->where('active = 1')->orderBy('name'),
-			'quantityunits' => $this->getDatabase()->quantity_units()->orderBy('name'),
+		    'currentStock' => $this->getStockService()->GetCurrentStockOverview(),
 			'locations' => $this->getDatabase()->locations()->orderBy('name'),
-			'currentStock' => $this->getStockService()->GetCurrentStock(true),
 			'currentStockLocations' => $this->getStockService()->GetCurrentStockLocations(),
-			'missingProducts' => $this->getStockService()->GetMissingProducts(),
 			'nextXDays' => $nextXDays,
 			'productGroups' => $this->getDatabase()->product_groups()->orderBy('name'),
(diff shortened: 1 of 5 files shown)


berrnd:
Great, thanks a lot for that too! :)

Makes sense to have the columns needed for the stock overview table populated by a SQL query - just one thing:
I tried to avoid to have "bigger" SQL queries in-code.
So maybe a view instead of the query in `GetCurrentStockOverview()` would be great, if possible somehow with the dynamic JOIN part based on feature flags...

> Note that JavaScript ("Datatables") also has some loading time. I think pagination could fix that.

The only real enhancement would be to have pagination + load the data dynamically and not render the whole HTML table server side...
The biggest table the stock journal, I have about 10K lines there on my private instances which loads currently in about 10 seconds - great that this page is only needed rarely on daily grocy use cases... :D

fipwmaqzufheoxq92ebc:
> Makes sense to have the columns needed for the stock overview table populated by a SQL query - just one thing:
> I tried to avoid to have "bigger" SQL queries in-code.
> So maybe a view instead of the query in `GetCurrentStockOverview()` would be great, if possible somehow with the dynamic JOIN part based on feature flags...

I changed that: I create now two views and query them.

berrnd:
Thanks again - the page speed improvement on the stock overview page is noticeable. :)

Maybe we should extend this "helper-view-approach" to at least all other overview pages...

berrnd:
I have no explanation, but any of the indexes create in [that migration](https://github.com/grocy/grocy/blob/62e8d88adb321d916a1a61eccb4efaf081f7f354/migrations/0106.sql) made that the `stock_current` view failed (the `is_aggregated_amount` did not work anymore, returned always `0`).

I searched for over an hour, but haven't found out what specific thing the problem is, it seems to be the combination of those indexes - strange...

PR #933, opened 2020-07-27
use last Chore-Log when determining the next assigned person

### LINKED ISSUE #930: Chore rotation alphabetic not looping
I notice that with some chores I make that the alphabetic rotation does not work. For example with this one: 

image

I try to rotate the chore by pressing the exectud buttong. When a chore reached Mirjam, I cannot get it back to Jan. I tried accessing grocy from both accounts..

fipwmaqzufheoxq92ebc:
Fixes #930 
I think I found the problem:
When determining the next assigned person for a chore, the last "tracked_time" is used.
If you "Track date only", and multiple users execute the chore at one day, it is "undefined" which of these executions is the last one. It seems to be always the first tracked on that date, so the next assigned person is whoever is after the person who first executed it.
That happens also with more than 2 persons. (and is not related to the alphabetically last person).

I fix this by using the last row_created_timestamp to identify the last execution.

Quick Fix: Uncheck the "Track date only" checkbox.

### DIFF
--- a/services/ChoresService.php
+++ b/services/ChoresService.php
@@ -133,7 +133,7 @@ class ChoresService extends BaseService
 
 		$chore = $this->getDatabase()->chores($choreId);
 		$choreLastTrackedTime = $this->getDatabase()->chores_log()->where('chore_id = :1 AND undone = 0', $choreId)->max('tracked_time');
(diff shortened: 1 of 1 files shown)


fipwmaqzufheoxq92ebc:
I should clarify this: I use row_created_timestamp when tracked_time is ambiguous, so when you track an execution for today and then for yesterday, the highest tracked_time (today) is used.
If there was already a tracked execution for today, the highest row_created_timestamp (of the executions today) is used.

berrnd:
Ah, I see, sorry and thanks for the clarification. :)

PR #940, opened 2020-08-01
Change nginx guidance in README

jordannewport:
The readme currently advises that nginx users put `try_files $uri
/index.php$is_args$query_string;` in the location block; this doesn't
work (produces 404s) and is also inconsistent with the example nginx
configuration at
https://github.com/grocy/grocy-docker/issues/15#issuecomment-486579894.

This PR changes the README so that it works and is consistent with that
configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Just unpack the [latest release](https://releases.grocy.info/latest) on your PHP
 
 Alternatively clone this repository (the `release` branch always references the latest released version, or checkout the latest tagged revision) and install Composer and Yarn dependencies manually.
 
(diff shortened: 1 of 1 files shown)


berrnd:
That's wrong and will not work, if you have 404 errors without that, it must be something else.

Without `$is_args$query_string`, the rewritten URL does not include query parameters and therefore they will be not visible for the server/backend. At least required for some API routes.

`grocy-docker` actually also uses exactly that ([ref](https://github.com/grocy/grocy-docker/blob/fe9f605fd9d82163709e791b84763e5ba1d5665f/docker_nginx/common.conf#L3-L5)), my mentioned comment is outdated (I updated it just now).

PR #945, opened 2020-08-03
Fix base path computation when running in subdirectory

### LINKED ISSUE #944: Inconsistent computation of $BasePath in Grocy\Helpers\UrlManager
I observed that the value of `Grocy.BaseUrl` is different when Grocy is running in the web root or in a subdirectory.

This is leads to a wrong computation of the `&returnto=` value in the product card edit buttion and hence later to an invalid redirect when Grocy runs in a subdirectory.

**Grocy in web root:**
```
[config.php] Setting('BASE_PATH', '');
[config.php] Setting('BASE_URL', '/');
[JS] Grocy.BaseUrl === 'https://grocy.example'
```

**Grocy in subdirectory:**
```
[config.php] Setting('BASE_PATH', '/subdir');
[config.php] Setting('BASE_URL', '/subdir/');
[JS] Grocy.BaseUrl === '/subdir'
```

**Problem caused:**
This is problematic because `Grocy.CurrentRelativeUrl` is computed like [this](https://github.com/grocy/grocy/blob/f6852e82b278d4fd8836346c2603f2852f760540/views/layout/default.blade.php#L53): `Grocy.CurrentUrlRelative = "/" + window.location.href.split('?')[0].replace(Grocy.BaseUrl, "");`

The `.replace` obviously only works if the base url starts with `https://` and the full hostname due to the usage of `window.location.href`.

If you run grocy in a subdirectory you get redirect urls like `https://grocy.example/subdir/product/14?returnto=/http://grocy.examplepurchase` which obviously breaks.

**The root cause:**
The `U` function uses [`constructUrl`](https://github.com/grocy/grocy/blob/f6852e82b278d4fd8836346c2603f2852f760540/helpers/UrlManager.php#L21) of the `UrlManager`. However the `$BasePath` is only set to the full `https://hostname` URL if the `BASE_PATH` constant is set to '/'. In any other case it is set to something that the frontend is unable to handle correctly.

I do not now why the special handling was introduced. I replaced the constructor by a simple `$this->BasePath = $this->GetBaseUrl() . $basePath;` and it works for all cases (grocy in web root and in subdirectory). I did not find any other downstream users of the `UrlManager`, so I think the change does not break other behaviour.

fgrsnau:
This is my attempt to fix #944. I tested it with a version of grocy running in the web root and a version running in a subdirectory.

### DIFF
--- a/helpers/UrlManager.php
+++ b/helpers/UrlManager.php
@@ -6,14 +6,7 @@ class UrlManager
 {
 	public function __construct(string $basePath)
 	{
-		if ($basePath === '/')
-		{
-			$this->BasePath = $this->GetBaseUrl();
-		}
-		else
-		{
-			$this->BasePath = $basePath;
-		}
+		$this->BasePath = $this->GetBaseUrl() . $basePath;
 	}
 
 	protected $BasePath;


berrnd:
See https://github.com/grocy/grocy/issues/944#issuecomment-675570413.

PR #954, opened 2020-08-18
Revert "Fix base path computation when running in subdirectory"

berrnd:
Reverts grocy/grocy#945


berrnd:
See https://github.com/grocy/grocy/issues/944#issuecomment-675570413.

PR #960, opened 2020-08-22
[WIP] Implemented basic permissions

fipwmaqzufheoxq92ebc:
I implemented a basic permission system.
Currently, I only check permissions in Users(Api-)Controller.

For that, I added an abstract class "User" and three implementations:
- AllowedUser  - always returns true on permission-check (= the old behavior)
- DefaultUser   - checks the database
- LockedUser   - for debugging, no permissions

A new Setting "USER_CLASS", controls which user-implementation to use.

TODOs:

- [x] Implement a UI to add/view permissions
- [x] Add a nice view for 403-Exceptions (thrown if permission is missing)
     - [x] Other exceptions like 404/500 could also get a nice view.
     - [ ] Hide buttons, e.g. "manage users", if the user is not allowed to access them?

- [x] Add more permissions
- [x] More complex permission-relations (e.g. ADMIN should implicitly have all other permissions, or someone who can edit something can also read/view that.)


fipwmaqzufheoxq92ebc:
I implemented the following Permissions:

https://github.com/fipwmaqzufheoxq92ebc/grocy/blob/b24267648834c174e6124d4d130d7dc680100de5/controllers/Users/User.php#L7-L31

I also added a new error page for exceptions, which redirects after 5 seconds to "/":
Screenshot_2020-08-26 You are not allowed to view this page grocy


I'm unsure if I should hide buttons which the user isn't allowed to use. I think it's confusing if there are buttons missing.

#263

berrnd:
Thanks a lot for starting this! :)
Will have a look as soon as possible.

I think it's ok to not hide menus.
But for example if I not have the permission to delete something, it would make more sense to grey out the corresponding delete button I would say, but sure it's technically more convenient to just check the permissions in the API...

berrnd (changes requested):
I gave it a quick try, this is what I noticed so far:

- `DefaultUser` didn't work, I removed the permission `PRODUCT_ADD`, adding products was still possible
- `LockedUser` worked, the API returns the correct HTTP status code, but HTML as content, I think this should be empty or JSON
- Do we really need the configurable implementation (beside debugging)? If we just give all users all permissions during migration, nothing would change if nobody cares about permissions (also for new users created later)
- After removing a permission, this seems to be saved/removed in the database, but when reloading the permission edit page the corresponding checkbox is still checked
- I don't think it's a good idea to redirect every error page after 5 seconds, maybe ok for 404, but for any other error I might want to have a longer look on the error message...
- Move permissions constants in a separate .pot file and translate them manually for english (like `chore_assignment_types`)

fipwmaqzufheoxq92ebc:
>     * `DefaultUser` didn't work, I removed the permission `PRODUCT_ADD`, adding products was still possible
PRODUCT_ADD means purchasing new products - I renamed the permission to PRODUCT_PURCHASE. For adding new Products, you need MASTER_DATA_EDIT.
 
>     * `LockedUser` worked, the API returns the correct HTTP status code, but HTML as content, I think this should be empty or JSON
I changed that: The API now returns JSON with "error_message" and "error_details"

>     * After removing a permission, this seems to be saved/removed in the database, but when reloading the permission edit page the corresponding checkbox is still checked
Seems to be a weird caching behavior on Firefox - should be fixed now. 

>     * I don't think it's a good idea to redirect every error page after 5 seconds, maybe ok for 404, but for any other error I might want to have a longer look on the error message...
I changed that: now only 404 redirects.

>     * Move permissions constants in a separate .pot file and "translate" them manually for english (like `chore_assignment_types`)
Done.

>     * Do we really need the configurable implementation (beside debugging)? If we just give all users all permissions during migration, nothing would change if nobody cares about permissions (also for new users created later)
I thought that it would be useful when I started implementing. 
For debugging, I instead change the permissions directly in the database.
I can remove the configuration if you want.

But you bring up another point: Default permissions for current & new users.
Currently, the first user (normally "admin") gets the ADMIN-Permission

fipwmaqzufheoxq92ebc:
I removed the configurable implementation and added a new DEFAULT_PERMISSIONS Setting.

I also started to disable buttons which the user isn't allowed to use.

berrnd:
Cool, thanks for all your work! :)

Everything looks good, will do another test later or tomorrow and then merge it - "detail polish" can be done later... :)

berrnd (approved):

berrnd:
Works! :)
I did some little naming and UI style changes in https://github.com/grocy/grocy/pull/960/commits/4f31627f8e06492826f771592adfbf03009df702.

And created some sub-tasks in new issues:
- #970
- #969
- #971

PR #965, opened 2020-08-26
Include location and shopping_location_id when opening products

kriddles:
Using Mark as Open in the stock entry page loses some information if new stock entries are created.


kriddles:
I just noticed that when marking open it also does not refresh the stock entries.

PR #973, opened 2020-08-29
Run database-Migrations in right order

### LINKED ISSUE #970: PHP/SQL migrations are not executed in right order
[`DatabaseMigrationService.php`](https://github.com/grocy/grocy/blob/master/services/DatabaseMigrationService.php) first executes all `.sql` migrations, then all `.php` migrations.

This should of course be in the right order regardless of the migration type.

Currently leads to that when starting from scratch, the default user (admin) has no permissions (as the default permissions are set in [`0111.sql`](https://github.com/grocy/grocy/blob/master/migrations/0111.sql) and the default user is created in [`0027.php`](https://github.com/grocy/grocy/blob/master/migrations/0027.php)).

References #960.

fipwmaqzufheoxq92ebc:
Fixes #970

### DIFF
--- a/services/DatabaseMigrationService.php
+++ b/services/DatabaseMigrationService.php
@@ -6,37 +6,28 @@ class DatabaseMigrationService extends BaseService
 {
 	public function MigrateDatabase()
 	{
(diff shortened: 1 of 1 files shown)


berrnd (approved):
Thanks! :)

Issue #441, opened 2019-11-13
Feature Request: Add attachments to equipment.


CorneliousJD:
I know we already have the ability to attach user manuals (which is fantastic) but sometimes equipment has other related documents. 

For one, I'd like to keep the following.
Sales receipt 
Warranty/extended warranty info
repair person contact info/receipts/service history, etc.

Thanks for the consideration!

CorneliousJD:
To add some additional context to an easy implementation, two good options I think would be.

1. Just allowing an attachment in notes, right now it's only photo or videos, but you can't add/link other PDFs or anything like that. 

2. Add a tab for attachments, so have user manual / notes / attachments and then you can just upload whatever file (within reason) and give it a friendly name maybe?

berrnd:
More or less the same as #262.

HydrelioxGitHub:
If this feature is implemented, it could be very useful for my needs. Especially for warranty files. I hope it will be integrated.

HydrelioxGitHub:
Any news on this enhancement ?

berrnd:
https://github.com/grocy/grocy#roadmap

berrnd:
This is now basically possible, there is new Userfield type "File" which can be used to attach any file, will be rendered as a link in tables - see #977.

Based on that, I created a follow-up issue to improve that even more for Equipment - I think a tab per file would be handy here - see #978.

berrnd:
Cross-referencing #978 here - it's now possible to add multiple files / PDFs to each equipment by using Userfields, changelog excerpt:

> - It's now possible to add multiple files (PDFs / manuals) to each equipment
>    - Define as many Userfields for the entity `equipment` and use the type `File`
>    - => Each of those File-Userfields will be shown as a separate tab on the equipment page

Issue #262, opened 2019-05-23
Feature Request: Ability to Add Multiple Manuals to Equipment


CyberRad:
Some of the items that I would like to add to the equipment portion of this product have multiple manuals. Repair and Owners manuals for example.

chadmccune:
This would also be nice for keeping receipts and proofs of purchase

berrnd:
This is now basically possible, there is new Userfield type "File" which can be used to attach any file, will be rendered as a link in tables - see #977.

Based on that, I created a follow-up issue to improve that even more for Equipment - I think a tab per file would be handy here - see #978.

berrnd:
Cross-referencing #978 here - it's now possible to add multiple files / PDFs to each equipment by using Userfields, changelog excerpt:

> - It's now possible to add multiple files (PDFs / manuals) to each equipment
>    - Define as many Userfields for the entity `equipment` and use the type `File`
>    - => Each of those File-Userfields will be shown as a separate tab on the equipment page

Issue #282, opened 2019-06-11
New Userfield type "picture"


berrnd:
References #280.

utrost:
Hello berrnd,
are there currently any plans to implement this new userfield of type "picture"?

berrnd:
Sorry, no schedules from my side - grocy is only my hobby, one I like, but not the only one, and because of that, there are no release dates, it's done when it's done, maybe tomorrow, maybe tomorrow next year, everyone is invited to contribute, I appreciate all ideas and contributions.

berrnd:
Was done by #977,

PR #977, opened 2020-08-30
Add user-field-type "file"

fipwmaqzufheoxq92ebc:
I added a new userfield-type "file", which allows to add a file to an object.
Files are uploaded via FilesApiController and shown as a link in tables.
I also added a new non-api-route to view the file, to show a human readable name, and to get a "normal" error page (instead of JSON).

Related to #262 and #441

### DIFF
--- a/controllers/FilesApiController.php
+++ b/controllers/FilesApiController.php
@@ -2,8 +2,8 @@
 
 namespace Grocy\Controllers;
 
-use Grocy\Controllers\Users\User;
 use \Grocy\Services\FilesService;
+use Slim\Exception\HttpNotFoundException;
 
 class FilesApiController extends BaseApiController
 {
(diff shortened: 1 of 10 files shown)


berrnd:
Should also fit for #282, just that the uploaded file shold be checked to be a picture and then also rendered as one (size?).

fipwmaqzufheoxq92ebc:
> Should also fit for #282, just that the uploaded file shold be checked to be a picture and then also rendered as one (size?).

I implemented a new field type "image", which is always served with force_serve_type=picture.
For tables I use a default size of 32x32 which is small, but there is not much space.

berrnd (approved):
Works in my tests, thanks! :)

I just added to render files and images in dialogs, when clicked on them in tables.

Also noted that empty userfields are never deleted from the `userfield_values` tables and that caused warnings as an empty file name was used on the pages - just added triggers to prevent empty fields in general.

berrnd:
I created a follow-up issue (#978) to fulfill #262 and #441 even better specifically for Equipment.

Issue #969, opened 2020-08-29
Document new API routes/error output/filter options


berrnd:
References #960 and #985.

Document in [`grocy.openapi.json`](https://github.com/grocy/grocy/blob/master/grocy.openapi.json):
- New user permission routes
- Changed error output
- New filter/order/pagination options, see https://github.com/grocy/grocy/issues/577#issuecomment-685048875
- New stock journal endpoints (see #989)
- New option `exact_amount` for the `/stock/consume` endpoint (see #1049 and https://github.com/grocy/grocy/pull/1049/commits/165abdeccb035b1bbe07339027cb84170bc90a12)

PR #985, opened 2020-09-01
Filtering of API-Results

### LINKED ISSUE #577: Feature request (API): search object by other fields

fipwmaqzufheoxq92ebc:
I've added a simple generic function for filtering, ordering and pagination for API-Results.
It can be used (only) on a LessQL\Result.

For now, I use it in '/objects/{entity}/search/' and '/recipes/fulfillment'.

Fixes #577 

It uses 4 (optional) Query-Params:

- order   The column to order by
- limit     The maximum count of objects to return
- offset     Number of elements to skip
- query[]    An array of conditions, each of them is a string with:
     - a column name
     - a comparison operator
     - a value to search for

The available comparison operators:
    - `=`    equal
    - `~`   LIKE
    - `<`   less
    - `>` greater
    - `>=` greater or equal
    - `<=` less or equal

All operators can be negated by prepending it with a `!`

Example:
`/api/objects/products/search/?order=name&limit=5&offset=1&query[]=name%3DCho`
(search for Products, order by name, max 5 items, skip the first, where the name = "Cho")
Has an empty result on demo data (english).

`/api/objects/products/search/?order=name&query[]=name~Cho`
(search for Products, order by name, where the is like Cho)
On demo data, returns 3 Products: "Chocolate", "Dark Chocolate", "Milk Chocolate"

Edit:
Also implemented for:
`/users`
`/tasks`
`/stock/products/{productId}/entries`
`/stock/products/{productId}/locations`
`/chores`
`/batteries`

### DIFF
--- a/controllers/BaseApiController.php
+++ b/controllers/BaseApiController.php
@@ -2,9 +2,15 @@
 
 namespace Grocy\Controllers;
 
+use LessQL\Result;
+
 class BaseApiController extends BaseController
 {
 	protected $OpenApiSpec = null;
+
+	const PATTERN_FIELD = '[A-Za-z_][A-Za-z0-9_]+';
+	const PATTERN_OPERATOR = '!?(=|~|<|>|(>=)|(<=))';
+	const PATTERN_VALUE = '[A-Za-z_0-9.]+';
 
 	public function __construct(\DI\Container $container)
 	{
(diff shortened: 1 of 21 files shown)


berrnd (approved):
Wow, thanks a lot again! :)

berrnd:
I have added this to #969 to track adding all the new things to [`grocy.openapi.json`](https://github.com/grocy/grocy/blob/master/grocy.openapi.json).

PR #986, opened 2020-09-02
Apikeys

### LINKED ISSUE #829: Display a QRCode for API Keys & API URL
For clients which use the API, it's required to somewhere type in the grocy API URL & an API Key, which is maybe somehow tedious...

So let's display a QRCode per API Key on the `/manageapikeys` page which contain the API URL and the Key (so use a separator, something like `https://grocy.tld/api|xxyAPIKEYxxy` which could then be scanned by the client (app).

//CC @arjes

fipwmaqzufheoxq92ebc:
Fixes #829.
I've added a QR-Code for the API-Key and API-Url:
For "default"-API-Keys, the QR-Data is:
`<API-Url>`|`<API-Key>`
(the url, a `|` as separator, then the API-Key)
For ical-API-Keys, the QR-Data is the URL including the "secret"-URL-Parameter

The QR-Code can be accessed on API-Key-List.
For ical, the QR-Code is also displayed when clicking on "Share/Integrate..." on the calendar page.

I used the Library [https://github.com/kazuhikoarase/qrcode-generator](url) mentioned by @Forceu in #829 

While implementing, I noticed two issues and fixed them:
- All users could see all API-Keys on the API-Keys page. Now only ADMIN can see all, while others can only see their API-Keys.

- All users with MASTER_DATA_EDIT could POST to `objects/api_keys` to create a new Key - for any user. Now only ADMIN can do that.


berrnd (commented):
Works, thanks! :)

Issue #963, opened 2020-08-26
Get stock journal with API?


MrJohnZoidberg:
Is there a way to get the stock journal with the API?

I can access details of a specific booking, but I don't see a possibility to access the full list.

berrnd:
Currently not, let's add it.

berrnd:
There is now the new endpoint `/stock/journal` to get the stock journal.

Issue #949, opened 2020-08-05
Feature Request: See user who made change in Stock Journal


jonathancaruso:
It would be extremely useful to see which logged in user made changes in the Stock journal.

Right now the columns listed are Product, Amount, Booking time, Booking type, and Location.

Adding a User column which lists which logged in user purchased, consumed, transferred, or corrected inventory, would be very nice.

Issue #64, opened 2018-09-24
Ideas/Needs for stock analysis


berrnd:
What do we want to have for stock analysis:

- An overview page with
  - Most bought/used products
  - The value of the current stock (based on tracked prices)
  - Percentage of spoiled products

- A detail page per product with
  - A list of the last bookings
  - How often the product was bought/used on average per week/month/year

On both pages the period (date from - date to) should be selectable.

BlizzWave:
One I thought of was "what recipe has consumed the most"

oakbrad:
How about analysis of spoiled products vs Product Group?

- Percentage of spoiled products vs Product Group, like a pie chart
- Historical data of the value (based on tracked prices) of spoiled products, filterable by Product Group
    - "What is the total cost of spoiled food over the last week"
    - "Is the cost of Vegetable spoilage going down or up" by Product Group
- List of most often spoiled products, or most spoiled per Product Group  ("buy less of these"). Maybe flag these visually somehow on product page/shopping list as 'Often Spoils'

berrnd:
Was accidentaly closed by a wrong commit message in https://github.com/grocy/grocy/commit/8020f92d6b18a720823e47e7cc2dc6daba551807...

hannahswain:
How about a list of the most recently used up/consumed products? 

You can set a minimum stock per product, but that's for well-known staples. It can be useful to go back and look at what has just been used up to see if it should be purchased again.

berrnd:
_Originally posted by @teitrand in https://github.com/grocy/grocy/issues/504#issue-551930583_

> I'd love to se some statistics on how much food goes to waste, and some functions to help reducing waste. Maybe get grocy to propose meals from the list of soon expiring food.

Wordna1:
What about leftover tracking?  I'm not exactly sure what this might look like, but such an analytic could certainly help confirm/adjust a recipe's servings.  In other words, if only half the recipe was consumed, perhaps it makes double the servings you thought.

raupie:
I think this is the feature I need the most and the only one I really care about. I don't think I can get my family on board to contribute to adding and removing inventory from stock. I care most about how much we spend a month/year on certain products. The data analytics side of things for budgeting purposes. Spotting trends or ways to save on buying things in bulk when there is a sale or coupons.

As in, I spent x amount on milk this year or x amount on toilet paper. 

[...]

PR #989, opened 2020-09-03
Stock-Journal: API, Summary, Done By

### LINKED ISSUE #827: Product ownership / roommate setup
While considering this for a roommate setup, some groceries would be shared but some are not. If my roommate and I both like to eat apples, it would be useful to track who is eating what. Perhaps there is a better way to denote shared vs. non-shared resources though.

fipwmaqzufheoxq92ebc:
Fixes #827 , #963 , #949 

I've implemented an API for the stock journal, available at `/api/stock/journal`.

For #827 and #949, I added the user to stock_log. The default for the migrations is 'admin'.

I also added a new Page (and API) "stock journal summary", which sums up the amounts grouped by product, user and transaction-type.

### DIFF
--- a/migrations/0115.sql
+++ b/migrations/0115.sql
@@ -1,53 +1,5 @@
-ALTER TABLE stock_log
-    RENAME TO stock_log_old;
-
-CREATE TABLE stock_log(
-    id                          INTEGER        not null        primary key autoincrement unique,
-    product_id                  INTEGER        not null,
-    amount                      DECIMAL(15, 2) not null,
-    best_before_date            DATE,
-    purchased_date              DATE,
-    used_date                   DATE,
-    spoiled                     INTEGER  default 0 not null,
-    stock_id                    TEXT           not null,
-    transaction_type            TEXT           not null,
-    price                       DECIMAL(15, 2),
-    undone                      TINYINT  default 0 not null     CHECK(undone IN (0, 1)),
-    undone_timestamp            DATETIME,
-    opened_date                 DATETIME,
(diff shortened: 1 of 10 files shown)


migrations/0115.sql
```diff
@@ -0,0 +1,89 @@
+ALTER TABLE stock_log
+    RENAME TO stock_log_old;
+
+CREATE TABLE stock_log(
```
berrnd, line 4: Was this only about adding the new column `user_id`? Then just `ALTER TABLE stock_log ADD user_id INTEGER NOT NULL` would be enough...

berrnd (approved):
Thank you! :)

berrnd:
Everything works, so I merged it - but what's the point of the new stock journal summary page?
The entries are grouped, ok - but without any date/time and/or undo possibility, I don't see any benefit of what this should say to me...
(For all the rest "analysis" things there is #64...)

fipwmaqzufheoxq92ebc:
> Everything works, so I merged it - but what's the point of the new stock journal summary page?
> The entries are grouped, ok - but without any date/time and/or undo possibility, I don't see any benefit of what this should say to me...
> (For all the rest "analysis" things there is #64...)

For #827  "reports of 'consumed by'": My understanding is that he means a summary, e.g. how many apples he eats, and how many apples his roommate eats. It calculates the SUM(amount) per Product,User,Booking type. yes - that may be related to #64 .

berrnd:
Ah, ok, thanks for the explanation - makes sense for this...

PR #995, opened 2020-09-07
Adds a few small frontend fixes

Ma27:
Detailed descriptions can be found in the corresponding commit messages. To summarize, the following things changed with this PR:

* 6ffad1d  adds instructions how to bootstrap a dev environment
* 2b0208e  fixes the checkbox-behavior for enabling/disabling the header clock
* 007b427: Make sure it's possible to delete and update a product's picture in one go.
* 248852b: Add a `Save & continue` button to product form.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,24 +40,6 @@ Just overwrite everything with the latest release while keeping the `data` direc
 
 If you run grocy on Linux, there is also `update.sh` (remember to make the script executable (`chmod +x update.sh`) and ensure that you have `unzip` installed) which does exactly this and additionally creates a backup (`.tgz` archive) of the current installation in `data/backups` (backups older than 60 days will be deleted during the update).
 
-## How to bootstrap a development environment
-
-Initialize:
-
-```
-$ npm install
-$ composer install
-$ cp config-dist.php data/config.php
-```
-
-Start the webserver:
-
-```
-$ cd public/
-$ ln -sf ../node_modules node_modules
-$ env GROCY_DATAPATH=$(pwd)/../data php -S localhost:3000
-```
-
 ## Localization
 grocy is fully localizable - the default language is English (integrated into code), a German localization is always maintained by me.
 You can easily help translating grocy at https://www.transifex.com/grocy/grocy, if your language is incomplete or not available yet.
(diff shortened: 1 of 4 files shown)


berrnd:
Thank you! :)

About the "how to dev env"-notes:
We had that multiple times before (#904, #811, #863), I still think this heavily depends on personal preference and there will be more questions if the described steps do not work, as this helps. Anyone who is familiar with PHP development will be able to set this up in his prefered way, I would say.
(Your example does not work on Windows, expects that PHP/composer/npm is available in `PATH`, that required PHP extensions are available, uses `npm` instead of `yarn`...)

Ma27:
> Anyone who is familiar with PHP development will be able to set this up in his prefered way, I would say.

I have to disagree unfortunately: I'm regularly doing PHP stuff and I had to fiddle around for a while to get everything running because this project has (just as every other bigger project) its own customizations in the structure I had to understand first. When I contribute to a project, I prefer to have some straight-forward instructions for a *minimal* setup.

In contrast to the other PRs this is a minimal approach (right?) which doesn't depend on additional tools such as `docker`.

>  (Your example does not work on Windows, expects that PHP/composer/npm is available in PATH, that required PHP extensions are available, uses npm instead of yarn...)

re/devtools: I thought that would be fine as those things are noted in the section above (I mean, you need `composer`/`npm` - it's available on every system with `node` in contrast to `yarn` - in any case, right?). Also for a misconfigured PHP we have the requirements checker, right?
re/windows: I could find a different way to install `node_modules` into `public/` and the `env GROCY_DATAPATH php -S` thing should be usable on Windows 10 with WSL, right? (correct me if I'm wrong though,I'm not a Windows user :))

However I understand if you don't want to have this section in the README atm, I could also move this into another PR.

berrnd:
If any, I prefer to just add some links in README (or on [grocy.info/links](https://grocy.info/links)) to this and other comments in the referenced issues. Multiple manual setups, the Docker one and the (my personally prefered) VSCode one is described, I think that's enough. I'm not against adding this in general, I just don't see the point if it's incomplete or describes (maybe too complex) how _I_ like to do it. It's more something for a blog article somewhere...

> I could find a different way to install `node_modules` into `public/` 

`yarn` does this automatically because of [`.yarnrc`](https://github.com/grocy/grocy/blob/master/.yarnrc)

> the `env GROCY_DATAPATH php -S` thing should be usable on Windows 10 with WSL

Probably. I'm a big fan of Windows (on workstations), but would never use WSL (personal pereferences everywhere :D)...

Everything needed, and that the `public` directory is the main webroot, is also already mentioned in the ["How to install"-section](https://github.com/grocy/grocy#how-to-install)...

berrnd (approved):

PR #1000, opened 2020-09-12
Some bug fixes.

### LINKED ISSUE #997: FEATURE_FLAG_STOCK_BEST_BEFORE_DATE_TRACKING not removing element from form
Hi,
when the FEATURE_FLAG_STOCK_BEST_BEFORE_DATE_TRACKING is set to false the purchase page is failing the form.checkValidity() call and is not enabling the OK button.
It looks like the best_before_date element is still in the form even is not visible.

fipwmaqzufheoxq92ebc:
- 515da5d The datetimepicker is now removed if FEATURE_FLAG_STOCK_BEST_BEFORE_DATE_TRACKING is not set (fixes #997)
- 9f9acb7 `DateTimePicker.SetValue()` did not set the value if the shortcut-checkbox was checked (e.g. "Never expires") - it only cleared the field and the checkbox (caused #997: without FEATURE_FLAG_STOCK_BEST_BEFORE_DATE_TRACKING, the field is set to "Never Expires". When you choose a product with default-best-before-date, the field is cleared/empty and hidden, and the validation failed.)

- ef2e5dd Clicking the whole (highlighted) area in drop downs with checkboxes now toggles the checkbox.
- 50ee248 Night-mode enabling does not need a reload
- Some undefined variables

### DIFF
--- a/controllers/GenericEntityApiController.php
+++ b/controllers/GenericEntityApiController.php
@@ -64,7 +64,7 @@ class GenericEntityApiController extends BaseApiController
 		}
 		else
 		{
-			return $this->GenericErrorResponse($response, $ex->getMessage());
+			return $this->GenericErrorResponse($response, "Invalid entity");
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


berrnd (approved):
Thanks! :)

PR #1018, opened 2020-09-26
Ui fixes

### LINKED ISSUE #1005: Recipes -> filters (engouh in stocks,..) doesn't work in Gallery label
Hi,

when I am searching recipe threw a filters in Gallery label with choosing "Enough in stock", "Not enough  in stock but on shopping list" and "Not enough in stock". I won't get any resaults for that. Filters are applied only in List label.

Can you also change behaviour of selecting recipes? List&Gallery. If I am searching for ingredients or whatever then I click on first recipe and page is refreshed and search box is empty. So if I want to see next recipe with searched ingredience I have to type it again. 

v2.7.1

### LINKED ISSUE #1017: undefined variable on product page (when FEATURE_FLAG_STOCK_PRICE_TRACKING is disabled)
I'm using master state from grocy. I can't delete custom conversions for products. Sadly this is not reproducible in prerelease demo (or at least I don't know what I did to break it 😅)  The following error occurs on load.

```
Uncaught TypeError: i is undefined
    jQuery 15
    <anonymous> http://server/viewjs/productform.js?v=2.7.1:395
jquery.dataTables.min.js:24:465
```

Clicking on the delete icon does not do anything. I can delete conversions using postman, so it's not the API. Any idea about this?

fipwmaqzufheoxq92ebc:
Fixes #1005.

Fixes #1017. If FEATURE_FLAG_STOCK_PRICE_TRACKING is not set, there is one more column in the header than in the body (in the barcode-table), which causes the Datatables-constructor in Line 395 to throw an error.
The handlers for deleting an quantity units are created after the table, and don't run because of the error.

### DIFF
--- a/public/viewjs/recipes.js
+++ b/public/viewjs/recipes.js
@@ -77,14 +77,20 @@ $("#status-filter").on("change", function()
 	recipesTables.column(5).search(value).draw();
 
 	$('.recipe-gallery-item').removeClass('d-none');
-	if(value !== "")
+	if (value !== "")
 	{
-		if(value === 'enoughtinstock')
-			$('.recipe-gallery-item').not('.recipe-enoughtinstock').addClass('d-none');
-		if(value === 'enoughinstockwithshoppinglist')
+		if (value === 'enoughinstock')
+		{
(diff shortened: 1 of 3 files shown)


berrnd (approved):
Thank you! :)

Issue #1024, opened 2020-10-01
Duplicate ingredients since migration 104


m-byte:
Migration 104 introduces a change to `recipes_pos_resolved` that causes some of the ingredients from the demo database to be listed multiple times in recipes.

berrnd:
Thanks for finding that, just to note here:
Migration 104 is not yet part of a release, so only affects master.

m-byte:
Yes, no release is affected by this.

berrnd:
Fixed in https://github.com/grocy/grocy/commit/e91bc0b01b804bde496a9812733e87583aeab679.

PR #1025, opened 2020-10-01
Fix duplicate ingredients

m-byte:
See #1024

### DIFF
--- a/migrations/0116.sql
+++ b/migrations/0116.sql
@@ -1,0 +1,102 @@
+-- Fix duplicate ingredients introduced in migration 104
+DROP VIEW recipes_pos_resolved;
+CREATE VIEW recipes_pos_resolved
+AS
+
+-- Multiplication by 1.0 to force conversion to float (REAL)
+
+SELECT
+	r.id AS recipe_id,
+	rp.id AS recipe_pos_id,
+	rp.product_id AS product_id,
+	rp.amount * (r.desired_servings*1.0 / r.base_servings*1.0) * (rnr.includes_servings*1.0 / CASE WHEN rnr.recipe_id != rnr.includes_recipe_id THEN rnrr.base_servings*1.0 ELSE 1 END) AS recipe_amount,
+	IFNULL(sc.amount_aggregated, 0) AS stock_amount,
+	CASE WHEN IFNULL(sc.amount_aggregated, 0) >= CASE WHEN rp.only_check_single_unit_in_stock = 1 THEN 1 ELSE rp.amount * (r.desired_servings*1.0 / r.base_servings*1.0) * (rnr.includes_servings*1.0 / CASE WHEN rnr.recipe_id != rnr.includes_recipe_id THEN rnrr.base_servings*1.0 ELSE 1 END) END THEN 1 ELSE 0 END AS need_fulfilled,
+	CASE WHEN IFNULL(sc.amount_aggregated, 0) - CASE WHEN rp.only_check_single_unit_in_stock = 1 THEN 1 ELSE rp.amount * (r.desired_servings*1.0 / r.base_servings*1.0) * (rnr.includes_servings*1.0 / CASE WHEN rnr.recipe_id != rnr.includes_recipe_id THEN rnrr.base_servings*1.0 ELSE 1 END) END < 0 THEN ABS(IFNULL(sc.amount_aggregated, 0) - (CASE WHEN rp.only_check_single_unit_in_stock = 1 THEN 1 ELSE rp.amount * (r.desired_servings*1.0 / r.base_servings*1.0) * (rnr.includes_servings*1.0 / CASE WHEN rnr.recipe_id != rnr.includes_recipe_id THEN rnrr.base_servings*1.0 ELSE 1 END) END)) ELSE 0 END AS missing_amount,
+	IFNULL(sl.amount, 0) * p.qu_factor_purchase_to_stock AS amount_on_shopping_list,
+	CASE WHEN IFNULL(sc.amount_aggregated, 0) + (CASE WHEN r.not_check_shoppinglist = 1 THEN 0 ELSE IFNULL(sl.amount, 0) END * p.qu_factor_purchase_to_stock) >= CASE WHEN rp.only_check_single_unit_in_stock = 1 THEN 1 ELSE rp.amount * (r.desired_servings*1.0 / r.base_servings*1.0) * (rnr.includes_servings*1.0 / CASE WHEN rnr.recipe_id != rnr.includes_recipe_id THEN rnrr.base_servings*1.0 ELSE 1 END) END THEN 1 ELSE 0 END AS need_fulfilled_with_shopping_list,
+	rp.qu_id,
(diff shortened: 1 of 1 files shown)


berrnd:
The problematic part is the join on the view (new, also introduced in migration 104) `products_oldest_stock_unit_price`.

@kriddles:
Do we need `products_oldest_stock_unit_price`? I only see one usage of that in StockService/GetProductDetails - but seems that this property is then never used anywhere...

kriddles:
> The problematic part is the join on the view (new, also introduced in migration 104) `products_oldest_stock_unit_price`.
> 
> @kriddles:
> Do we need `products_oldest_stock_unit_price`? I only see one usage of that in StockService/GetProductDetails - but seems that this property is then never used anywhere...

From what I remember, the products_oldest_stock_unit_price was written the way it is in an attempt to make sure to get the correct price even if stock information is edited through the UI.

It is necessary to calculate the cost column correctly in the recipes_pos_resolved.

The comments suggest only one stock row is returned using max but... I don't see it so that comment may have been written in a different iteration of the query or I forgot to put the max in...

And it looks like I added it to the GetProductDetails but never put it on the card to display.

kriddles:
It looks like this commit changed the query to due to errors on the product detail page for not in stock items but began to return multiple rows per product.
[https://github.com/grocy/grocy/commit/939b98e470a5b65786ec82bbb342edc9c8715dc5](url)

I'll take a look further

kriddles:
Hi @berrnd, 
I would recommend reverting these two commits 939b98e470a5b65786ec82bbb342edc9c8715dc5 and 939b98e470a5b65786ec82bbb342edc9c8715dc5 to restore the original queries and then will you take a look and let me know what you meant in these commit comments "Otherwise there are errors when getting product details for currently not in stock items" when using the original queries? I do not see any errors in my testing with the original queries using the demo database. Using the original queries also resolves the recipe duplicated ingredients issue.

berrnd:
When reverting the two mentioned commits the productcard fails to load for not in stock products (default demo data, product "Cookies" for example) and there will be a lot warnings about missing object properties when populating the demo data. Can be reproduced by checking out https://github.com/grocy/grocy/commit/ffec1134a3866e336fefaea74964ad439976039d.

kriddles:
> When reverting the two mentioned commits the productcard fails to load for not in stock products (default demo data, product "Cookies" for example) and there will be a lot warnings about missing object properties when populating the demo data. Can be reproduced by checking out [ffec113](https://github.com/grocy/grocy/commit/ffec1134a3866e336fefaea74964ad439976039d) (the commit before these two).

I'm not sure where to look for the missing object properties warnings for populating demo data but that should be unrelated to updating these views. It looks like you have a commit that resolves demo data issues fdb419  

I checked out ffec113 and also created a branch on master with the changes but in both cases I am able to see the "Cookies" product card from the Stock Overview page. I also see it on the purchase page. Is there a different location that I should try to launch it from to see it fail loading?

image

berrnd:
If you only checkout that commit, it's not reproducible. You also have to delete/recreate the database (otherwise it already contains migrations which not existed at that point), then it's definitely reproducible, also with the productcard from the stock overview page...

kriddles:
> If you only checkout that commit, it's not reproducible. You also have to delete/recreate the database (otherwise it already contains migrations which not existed at that point), then it's definitely reproducible, also with the productcard from the stock overview page...

Yes, I deleted the database so that it would rebuild from the migrations but did not see the product card issue.

berrnd:
I've checked it: It's all about the missing object properties. When you don't see it, you probably have disabled them - `php.ini` option `error_reporting`, `E_ALL` is the default in the distributed `php.ini-development`.

So let's revert the two mentioned commits to fix this problem and take care of the missing properties "problem" later.

There are a lot of those notices when executing migrations/populating demo data, so when visiting the root route, before the redirect to `/stockoverview` happens.

And the productcard problem is caused by that for example the response of `/api/stock/products/1` is the following, which is not valid JSON at the end and parsing fails:
```
<br />
<b>Notice</b>:  Trying to get property 'price' of non-object in <b>C:\Users\beb\WorkspaceP\grocy\services\StockService.php</b> on line <b>134</b><br />
<br />
<b>Notice</b>:  Trying to get property 'qu_factor_purchase_to_stock' of non-object in <b>C:\Users\beb\WorkspaceP\grocy\services\StockService.php</b> on line <b>135</b><br />
<br />
<b>Notice</b>:  Trying to get property 'shopping_location_id' of non-object in <b>C:\Users\beb\WorkspaceP\grocy\services\StockService.php</b> on line <b>139</b><br />
<br />
<b>Notice</b>:  Trying to get property 'purchased_date' of non-object in <b>C:\Users\beb\WorkspaceP\grocy\services\StockService.php</b> on line <b>152</b><br />
<br />
<b>Notice</b>:  Trying to get property 'price' of non-object in <b>C:\Users\beb\WorkspaceP\grocy\services\StockService.php</b> on line <b>164</b><br />
<br />
<b>Notice</b>:  Trying to get property 'price' of non-object in <b>C:\Users\beb\WorkspaceP\grocy\services\StockService.php</b> on line <b>165</b><br />
{"product":{"id":"1","name":"Cookies","description":null,"product_group_id":"1","active":"1","location_id":"4","shopping_location_id":null,"qu_id_purchase":"3","qu_id_stock":"3","qu_factor_purchase_to_stock":"1.0","min_stock_amount":"8","default_best_before_days":"0","default_best_before_days_after_open":"0","default_best_before_days_after_freezing":"0","default_best_before_days_after_thawing":"0","picture_file_name":"cookies.jpg","allow_partial_units_in_stock":"0","enable_tare_weight_handling":"0","tare_weight":"0.0","not_check_stock_fulfillment_for_recipes":"0","parent_product_id":null,"calories":"123","cumulate_min_stock_amount_of_sub_products":"0","row_created_timestamp":"2020-10-12 17:05:00"},"product_barcodes":[],"last_purchased":null,"last_used":null,"stock_amount":0,"stock_factor_purchase_amount":0,"stock_value":0,"stock_amount_opened":0,"stock_amount_aggregated":0,"stock_amount_opened_aggregated":0,"quantity_unit_purchase":{"id":"3","name":"Pack","description":null,"row_created_timestamp":"2020-10-12 17:04:58","name_plural":"Packs","plural_forms":null},"quantity_unit_stock":{"id":"3","name":"Pack","description":null,"row_created_timestamp":"2020-10-12 17:04:58","name_plural":"Packs","plural_forms":null},"last_price":null,"last_qu_factor_purchase_to_stock":null,"avg_price":null,"oldest_price":null,"last_shopping_location_id":null,"default_shopping_location_id":null,"next_best_before_date":null,"location":{"id":"4","name":"Candy cupboard","description":null,"row_created_timestamp":"2020-10-12 17:05:00","is_freezer":"0"},"average_shelf_life_days":-1,"spoil_rate_percent":0,"is_aggregated_amount":0}
```

berrnd:
This is now fixed by https://github.com/grocy/grocy/commit/e91bc0b01b804bde496a9812733e87583aeab679 like proposed by @kriddles above.

berrnd:
Missing property warnings are now also fixed in https://github.com/grocy/grocy/commit/a66a4d0c225ed45083940855f577bc0633654f04.

Issue #928, opened 2020-07-26
Shopping list "Add expired products"


cwoomi:
Is it possible to add a function for the shopping list to add all expired products so you can replace them easily?

berrnd:
Currently not, but thanks for the idea! :)

PR #1026, opened 2020-10-02
added button to add expired products to shoppinglist

sourcefactory-at:
I added a new button to add all expired products to a shoppinglist. This will add 1 unit for each expired product which is not yet in the list.

The change also includes a new API endpoint /api/stock/shoppinglist/add-expired-products which works like /api/stock/shoppinglist/add-missing-products.

As I saw there is also an enhancement issue for this: #928

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -31,6 +31,30 @@ class StockApiController extends BaseApiController
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


berrnd (approved):
Thank you! :)

Works, I cleaned up some little things - see my commits here.

PR #1029, opened 2020-10-04
Recipe templates

m-byte:
Fractions of the total ingredients required can be mentioned in recipes. This can be helpful when, for example, only half of some ingredient is needed at the start of the recipe. The proper values are automatically calculated when the number of servings is adjusted.

### DIFF
--- a/public/viewjs/recipeform.js
+++ b/public/viewjs/recipeform.js
@@ -28,6 +28,9 @@ $('.save-recipe').on('click', function(e)
 {
 	e.preventDefault();
 
+	$(".note-editable span.ingredient").removeClass(["badge", "btn", "btn-secondary"]);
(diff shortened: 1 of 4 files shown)


berrnd (changes requested):
Thank you.

I see the point about this, but I don't think this is really intuitive to use that way - kind of unexpected that when using the button some text is inserted at the current cursor position of the text editor (or nowhere, when the cursor position is undefined). Spontaneously also have no idea how to do that "better"...

For me personally a text like "take half of the amount of ingredient xy" does the same job...

berrnd:
Closing this for now, feel free to reopen it if you want to continue on this / have an idea how to do this in an kind of more user friednly/predictable way.

PR #1043, opened 2020-10-12
productCard Fix

kriddles:
I have not tested this fix but I believe it will resolve the issue you mentioned in https://github.com/grocy/grocy/pull/1025#issuecomment-707188338


berrnd:
Thank you - but already did something similar just now - so pretty the same already done in https://github.com/grocy/grocy/commit/a66a4d0c225ed45083940855f577bc0633654f04.

PR #1045, opened 2020-10-12
Value information

kriddles:
I find it interesting for some of my products to know how much I've spent on the current stock and statistics for total value.
This branch adds the total value and also a value column to the stock overview page if stock price tracking is enabled.


berrnd (approved):
Thank you, works! :)

I removed the total units info, was already removed before - I think this number is totally senseless (if you store flour and some other similar product for example in grams, this number is always really high but means nothing).

As maybe this is not interesting for everybody, I think we should find a way to be able to hide on every table - created #1058 to track that.

Issue #888, opened 2020-06-09
Feature request: Date field for purchases


towo:
Occasionally, I'm a bit of a slacker and just enter purchases a day or two later by using the receipt. 
For these cases (and "bought it a day earlier" and somesuch), it'd be great if there were a date field in the purchase dialogue, preseeded with the current date.

Ideally, it would retain the shopping date during the current session, so one can rapidly add things bought at a specific date.

Otherwise, you manually have to go back and move entries one by one, by hand, and I'm not sure it doesn't mess up the store pricing, either.

PR #1056, opened 2020-10-15
add purchased date to purchase

kriddles:
feature #888

Initial feature request checkin
I did not limit this to GROCY_FEATURE_FLAG_STOCK_BEST_BEFORE_DATE_TRACKING but perhaps it should be?

This might be made a little better by updating the best before date whenever the purchase date is changed?

### DIFF
--- a/config-dist.php
+++ b/config-dist.php
@@ -114,6 +114,7 @@ DefaultUserSetting('stock_default_consume_amount', 1);
 DefaultUserSetting('scan_mode_consume_enabled', false);
 DefaultUserSetting('scan_mode_purchase_enabled', false);
 DefaultUserSetting('show_icon_on_stock_overview_page_when_product_is_on_shopping_list', true);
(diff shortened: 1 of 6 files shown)


berrnd:
I think this is not related to `FEATURE_FLAG_STOCK_BEST_BEFORE_DATE_TRACKING`, it's just about to provide a manual purchase date, instead of always using today (don't know if we should always show this field on the purchase page, maybe best hide it by default and add a stock setting (top right corner menu) for this).

kriddles:
> I think this is not related to `FEATURE_FLAG_STOCK_BEST_BEFORE_DATE_TRACKING`, it's just about to provide a manual purchase date, instead of always using today (don't know if we should always show this field on the purchase page, maybe best hide it by default and add a stock setting (top right corner menu) for this).

Sounds good, I added a stock setting checkbox for this feature to hide it by default.

berrnd (approved):

PR #1057, opened 2020-10-15
fix for price and total price

kriddles:
Price calculations were not correct in Purchase page and last price was not initialized correctly which was most likely resolved in my other wip branch.
This fixes it for master.


berrnd (approved):
Ok, thanks - don't know what was wrong, but looks good. :)

Also wanted to give all the changed price handling things a more thorough test, hopefully some time soon...

PR #1067, opened 2020-10-18
Fix night mode

JohnMH:
This fixes a few lines of white when in night mode.

This change was done from a tablet, I can redo this commit properly on a computer when I get back home if needed. This just really annoyed me ;)


berrnd (approved):
Thanks! 🙂

PR #1077, opened 2020-10-20
Allow price up to 4 decimals instead of 2

kriddles:
Some stock units may resolve to a price less than .01 so should be allowed to be max 4 decimal places .0001
For example, buying 1 bag (purchase unit) of flour stock units 500 grams for $4 a bag would set a price per gram of $0.008.

This commit also resolves a bug setting the price correctly for purchase. using the qu_factor


kriddles:
We also could just allow the price decimal to be a stock user setting?

berrnd:
> We also could just allow the price decimal to be a stock user setting?

Yes, also thought about it as this has been controversial often in the past.

Let's add two settings:
- decimals for amount inputs
- decimals for price inputs

kriddles:
Updated to include decimals for price inputs as a user setting.

Will you take care of the amount setting after reviewing the price input settings?

berrnd (approved):
Thanks!

> Will you take care of the amount setting after reviewing the price input settings?

Have finished it (hopefully) in https://github.com/grocy/grocy/pull/1077/commits/4648409e03f20a30cf238a1d5e6d18047d96cd22.

PR #1079, opened 2020-10-20
stock entries button on product card

kriddles:
Add button on product card to go to the stock entries for the product


berrnd (approved):
👍🙂

PR #1081, opened 2020-10-20
inventory fix for quFactorPurchaseToStock - defaults to product details

kriddles:
missing parameter - if null just uses product detail information for AddProduct


kriddles:
bug occurs if you adjust inventory adding more - the qu factor is incorrectly being set using the location id

berrnd (approved):
👍🙂

PR #1102, opened 2020-10-27
Add purchased date to inventory

kriddles:
Allow added inventory to set purchased date with same permissions as on the purchase tab.


berrnd (approved):
Thanks!

PR #1103, opened 2020-10-27
dialogs for shopping list

### LINKED ISSUE #1091: Use dialogs instead of jumping between pages where appropriate
Follow up to #1089.

Like already done for adding recipe ingredients, use dialogs instead of jumping between pages everywhere (e. g. when adding shopping list items).

kriddles:
This adds an edit button for shopping lists and attempts to use dialogs.
fixes #1091

One note: if you could look into the add product to shopping list it is not refreshing the item list after saving the dialog box.
I'm not sure if you want to just refresh the whole screen or refresh the datatable.

### DIFF
--- a/public/viewjs/shoppinglist.js
+++ b/public/viewjs/shoppinglist.js
@@ -447,3 +447,13 @@ if ($(window).width() < 768 & window.location.hash !== "#compact" && !BoolVal(Gr
 {
 	$("#shopping-list-compact-view-button").click();
 }
+
+$(window).on("message", function(e)
+{
+	var data = e.originalEvent.data;
+
+	if (data.Message === "ShoppingListChanged")
+	{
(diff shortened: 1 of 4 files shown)


kriddles:
fix #1091

berrnd:
>  I'm not sure if you want to just refresh the whole screen or refresh the datatable.

Currently the tables are nowhere refreshed by JavaScript, would be of course nicer - for now just a page refresh is ok I think...

kriddles:
> > I'm not sure if you want to just refresh the whole screen or refresh the datatable.
> 
> Currently the tables are nowhere refreshed by JavaScript, would be of course nicer - for now just a page refresh is ok I think...

Sounds good, I fixed it to now just refresh the page when adding or editing the shopping list items

berrnd (approved):
Thanks!

I'll leave #1091 open, I think there are still more places were dialogs would be better...

Issue #1100, opened 2020-10-26
Group stock transaction together by "action"


berrnd:
_Follow up for #1061._

_Mentioned by @berrnd in https://github.com/grocy/grocy/issues/1061#issuecomment-716139435:_
> - Each stock entry (table `stock`), has an `stock_id`, this id is generated when the entry was first added (normally via purchase)
> - Stock entries can get split, this `stock_id` remains the same and groups rows in the `stock` table together
> - Stock journal entries (table `stock_log`) have also that `stock_id` and
>   - a column `transaction_id` to group all used stock entries by one transaction together (e. g. when you consume an amount of 5 and this transaction used 5 different rows/entries of the `stock` table)
>   - a column `correlation_id` to group journal entries together which are dependent on each other (currently only used when editing stock entries, there is one journal entry for the values before and one for the values after the edit action)
> 
> So when we need this (?), this would be maybe more a "action_id". If using the normal purchase page, the question would be also when such an action starts and when does it end...

_Mentioned by @white-gecko in https://github.com/grocy/grocy/issues/1061#issuecomment-716502921:_
> I think per default everything is an individual action, i.e. the transactions are not part of any action (`null`). For actions like "add all items to stock" it could be possible to group the transaction automatically. But in other cases an action has to be initiated and finished manually. Also it should be possible to manually associate transactions with an action, or remove transactions from the action.

So let's add such an `action_id` for stock journal entries.

berrnd:
Closing this. Unless we need such an `action_id` for a real practical use case, there is no need to add that.

PR #1107, opened 2020-10-29
StockApiController transactionId

kriddles:
I do not think this affects anything... it's just for completion sake most likely for an api user

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -276,7 +276,13 @@ class StockApiController extends BaseApiController
 			{
 				$consumeExact = $requestBody['exact_amount'];
 			}
+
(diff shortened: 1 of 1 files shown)


berrnd:
I don't think we should expose / make it possible to set the `transaction_id` externally.

See #1100 - to group stock actions together (and yet don't really know what to do with it) it's more about a separate id...

kriddles:
> I don't think we should expose / make it possible to set the `transaction_id` externally.
> 
> See #1100 - to group stock actions together (and yet don't really know what to do with it) it's more about a separate id...

Ah, yes that makes sense. I'm not yet sure about the stock actions yet either but I'll think about it a bit.

PR #1109, opened 2020-10-29
Unit stock name in the mealplan

BenoitAnastay:
Stock name instead of purchase name

I'm not sure if it's the right variable, because there is still a convertion function but this seem better than wrongly displaying the purchase unit name


berrnd:
Good find, thank you! :)

Will also need to test this, maybe that's not enough as when adding a product to the meal plan a different quantity unit (which has a related conversion) can be choosen...

BenoitAnastay:
Currently it's displaying the qty in the stock value with the decimal (250ml will display 0.25Litres), `data-meal-plan-entry` do not return the converted value and name, thus to fix it we should edit the PHP API side

berrnd (approved):

PR #1111, opened 2020-10-31
limit to normal recipes

kriddles:
- The recipe dropdown for consume should be limited to normal recipes. Where else should this be considered?
- The transfer page does not need to load recipes.

### DIFF
--- a/controllers/StockController.php
+++ b/controllers/StockController.php
@@ -2,6 +2,8 @@
 
 namespace Grocy\Controllers;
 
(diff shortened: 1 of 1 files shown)


berrnd:
Thanks.

> The recipe dropdown for consume should be limited to normal recipes. Where else should this be considered?

Everywhere where a recipe can be selected I think, but I don't see any other places than this and the recipe page / add included recipe where already only normal recipes are shown.

berrnd (approved):

PR #1112, opened 2020-10-31
Polish localization

RafalDadas:


berrnd:
Thank you, but localizations are manage via [Transifex](https://www.transifex.com/grocy/grocy) as [mentioned in README](https://github.com/grocy/grocy#localization) - feel free to do the changes there.

PR #1115, opened 2020-11-01
[WIP] Simplified overviews on mobile

4lloyd:
With this PR I want to add some improvements to the UI so that on mobile devices the default view is cleaner and simpler.

I made the following changes:
1. The related links buttons are hidden and can be accessed via the 3-dot button in the top right corner.
2. The expiring, expired and missing product messages are slimmed down and the text is changed to an icon.
3. The filters are hidden and can be accessed via the filter button next to the "clear filter" button.

All changes are only visible on XS and SM screens (according to the Bootstrap grid system). The UI for larger screens has not been changed visually.

Please let me know if you approve the changes or if you have some suggestions to improve them.

If you approve the changes then I also want to change the following pages in the same style:
- Chores overview
- Tasks
- Batteries overview
- Calendar (only the iCal button)

Below I have added some screenshots. The left one shows the UI when the user accesses the page. The right one shows both hidden menus opened.
<p float="left">

</p>


berrnd:
Cool, thanks a lot, great idea! :)

Currently we only had a "special handling" for the shopping list page (button "Compact mode").

If improving that, maybe another point would be the tables ([DataTables](https://datatables.net/) specific) - as of now we use the [responsive plugin](https://datatables.net/extensions/responsive/) which adds those "+" buttons when not columns fit. Maybe to make the tables horizontally scrollable somehow would be better...

4lloyd:
I like your idea of making the table scroll horizontally, so I have implemented it. For now, it only works on the stock overview page, but I will check if I can enable it on other pages.

Some small remarks:
- The datatables dependencies are updated because of a bug which misaligned the header and the body of the table.
- The dragscroll dependency is needed for when the user has a small screen on a mouse operated device. This allows to horizontally drag through the table. This extra dependency is not needed on a mobile device.
- The cursor is changed to a hand (grab) when it's possible to scroll horizontally. When clicked the hand will change in a grabbing hand. For some reason, the change from a hand to a grabbing hand doesn't work on Firefox. I have tried multiple possible solutions, but none worked. On Chrome and Edge, it works fine.

berrnd:
Great, thanks for your work! :)

Will give it a test in the next days and can also take care about the other pages then...

berrnd (approved):

4lloyd:
I tried out the changes and they are looking good. Thank you and great work! 👍 

I found two small things. I'm not sure if you already know them, so to be sure I mention them here.
1. Because of removing `.dataTables_scrollBody { overflow: visible !important; }` from grocy.css, the overflow of the dropdown menu in the stock overview table is partially hidden. The CSS was removed because it blocked the horizontal scrolling. I missed the problem during testing because it only occurs when there are a few rows in the table. The solution may be to use a pop-up instead or I have a possible solution/workaround at the end. The same problem occurs in the chores overview table and batteries overview table because they have the same kind of dropdown menu.
2.  Grabbing and scrolling through the recipe overview table also sends a click event which changes the selected visible recipe.

Possible solution/workaround for 1:
This adds the "force-overflow-visible" class to the table when a dropdown is active. A side effect is that when the class is added the table can't be scrolled horizontally. The code is based on my last commit, so you probably need to change it so it fits within the current code.

Add to "grocy.css":
```
.force-overflow-visible {
	overflow: visible !important;
}
```

Change line [205](https://github.com/grocy/grocy/blob/9949f30c2b429d0ea6310190abcff277a6852e3f/views/stockoverview.blade.php#L205) of "stockoverview.blade.php" to (adding the class "dropdown-helper"):
`<div class="dropdown d-inline-block dropdown-helper">`

Change the "column-sizing.dt" event in "stockoverview.js" to (adding two lines):
```
$('#stock-overview-table').on('column-sizing.dt', function(e, settings)
{
	var dtScrollWidth = $('.dataTables_scroll').width();
	var stockTableWidth = $('#stock-overview-table').width();

	if (dtScrollWidth < stockTableWidth)
	{
		$('.dataTables_scrollBody').addClass("grab-cursor");
		$('.dataTables_scrollBody').removeClass("force-overflow-visible");
	} else
	{
		$('.dataTables_scrollBody').removeClass("grab-cursor");
		$('.dataTables_scrollBody').addClass("force-overflow-visible");
	}
});
```

And add below that:
```
$('.dropdown-helper').on('show.bs.dropdown', function(e)
{
	if ($('.dataTables_scrollBody').hasClass("grab-cursor"))
	{
		$('.dataTables_scrollBody').addClass("force-overflow-visible");
	}
});

$(".dropdown-helper").on('hide.bs.dropdown', function(e)
{
	if ($('.dataTables_scrollBody').hasClass("grab-cursor"))
	{
		$('.dataTables_scrollBody').removeClass("force-overflow-visible");
	}
});
```

berrnd:
Yes, already noticed that the dropdown is hidden when the table is too small, thanks for the suggestions - will try them.

I'm unfortunately not so into CSS, maybe only making `overflow-y` visible on `.dataTables_scrollBod` may help for a simpler soluation...

4lloyd:
I tried adding `overflow-y: visible !important` to `.dataTables_scrollBody`, but for some reason it doesn't work. Maybe you can find a way to make it work.

berrnd:
Had no luck, your suggestion works, so just (re)used generic selectors to not have the function on every page and adapted it in https://github.com/grocy/grocy/commit/b15740bdede5a799bcf89d11433184554e46a663. Thanks! :)

berrnd:
I've removed the dragscroll dependency in https://github.com/grocy/grocy/commit/d2d09cf92806bb19a42885cfec84087b44ca04af due to #1135 (text can't be selected, I think this isn't really needed as it works on touch devices without that).

PR #1127, opened 2020-11-14
Fix is_aggregate_amount always 0 on stock_current

fipwmaqzufheoxq92ebc:
> I have no explanation, but any of the indexes created in that migration made that the stock_current view failed (the column is_aggregated_amount did not work anymore, returned always 0).
I searched for over an hour, but haven't found out what specific thing the problem is, it seems to be the combination of those indexes - strange...

I've found the issue:

is_aggregated_amount is not an aggregated column (not min(), avg(), sum(), etc.),
so sqlite picks a value from one of the rows in the group - from which row is undefined.

The indexes change how the query is executed, which causes sqlite to pick another row.

### DIFF
--- a/migrations/0106.sql
+++ b/migrations/0106.sql
@@ -1,3 +1,33 @@
+CREATE INDEX ix_products_performance1 ON products (
+    parent_product_id
+);
+
+CREATE INDEX ix_products_performance2 ON products (
+    CASE WHEN parent_product_id IS NULL THEN id ELSE parent_product_id END,
+    active
+);
+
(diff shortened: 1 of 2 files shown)


berrnd:
Thanks for looking into this, makes sense.

Will test it and also restore your added performance indexes from [migration 106](https://github.com/grocy/grocy/blob/62e8d88adb321d916a1a61eccb4efaf081f7f354/migrations/0106.sql) which I had removed for now.

berrnd (approved):
Thanks, that fixed it. :)

PR #1139, opened 2020-11-20
Typo for consumed

lexitus:
Fix typo for consumed. This also changes the msgid so that it is correct, but I don't know if this will screw up something with Transifex? (I know nothing about Transifex). If it will then I guess we would be better off changing just the msgstr in localization/en_GB/strings.po.

### DIFF
--- a/localization/cs/strings.po
+++ b/localization/cs/strings.po
@@ -1443,7 +1443,7 @@ msgstr ""
 msgid "this location"
 msgstr "toto umístění"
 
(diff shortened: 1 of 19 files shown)


berrnd:
Thank you, but this has already been fixed, the translations have just not been pulled from Transifex for quite a time.

The only files edited manually in the repo are the `*.pot` files on the first level of the [localization directory](https://github.com/grocy/grocy/tree/master/localization), everything else comes from Transifex.

Issue #1148, opened 2020-12-01
Feature request: On Purchase - prefill the Amount or/and Standard Amount per Produkt


sleeplessnight2:
It makes sense to fill in the consumption amount with the Quick Consume Amount instead of a standard amount.

A _global_ standard amount make really no sense... in 90% the usecases i changed this, because the consumption is individual or it ignores "the factor purchase to stock quantity" unit when consuming. 

For Example: 
The Factor is: 1 Piece is 500gr. But when I consume it, I have to manually enter the factor (i.e. 500), otherwise the 1gr will be booked out. The Quick Consume Amount is a solution for this - but the process is not complete finished yet.

A suggestion:
Rename at the Product Page the Field "Quick Consume Amount" into "Standard Consume Amount" and use that accordingly throughout the program.

It saves a lot of your time dealing with grocy.

berrnd:
The easiest thing is that the default consume amount is 1 (QU stock) for all products, so depends on how you organize your products and of course never works for something like Flour tracked in grams - I personally do it like that.

But of course this is a valid feature request to also use the "Quick consume amount" also as a default on the consume page.

PR #1149, opened 2020-12-01
prefill the Amount or/and Standard Amount per Produkt

sleeplessnight2:
Reference to #1148

I'm trying something. Just correct the Line 383.. 
Please do not forget, I program badly. That's why I don't understand all the dependencies ... I've only known the project for about 1 month

### DIFF
--- a/public/viewjs/consume.js
+++ b/public/viewjs/consume.js
@@ -94,7 +94,7 @@
 						Grocy.Components.ProductAmountPicker.Reset();
 						$("#display_amount").attr("min", "0." + "0".repeat(parseInt(Grocy.UserSettings.stock_decimal_places_amounts) - 1) + "1");
 						$("#display_amount").removeAttr("max");
-						$('#display_amount').val(parseFloat(Grocy.UserSettings.stock_default_consume_amount));
+						$('#display_amount').val(parseFloat(productDetails.product.quick_consume_amount));
 						RefreshLocaleNumberInput();
 						$(".input-group-productamountpicker").trigger("change");
 						$("#tare-weight-handling-info").addClass("d-none");
@@ -160,7 +160,7 @@ $('#save-mark-as-open-button').on('click', function(e)
 					Grocy.FrontendHelpers.EndUiBusy("consume-form");
 					toastr.success(__t('Marked %1$s of %2$s as opened', jsonForm.amount + " " + __n(jsonForm.amount, productDetails.quantity_unit_stock.name, productDetails.quantity_unit_stock.name_plural), productDetails.product.name) + '<br><a class="btn btn-secondary btn-sm mt-2" href="#" onclick="UndoStockTransaction(\'' + result.transaction_id + '\')"><i class="fas fa-undo"></i> ' + __t("Undo") + '</a>');
 
-					$('#display_amount').val(parseFloat(Grocy.UserSettings.stock_default_consume_amount));
+					$('#display_amount').val(parseFloat(productDetails.product.quick_consume_amount));
 					RefreshLocaleNumberInput();
 					$(".input-group-productamountpicker").trigger("change");
 					Grocy.Components.ProductPicker.Clear();
@@ -276,7 +276,7 @@ Grocy.Components.ProductPicker.GetPicker().on('change', function(e)
 
 				Grocy.Components.ProductAmountPicker.Reload(productDetails.product.id, productDetails.quantity_unit_stock.id);
 				Grocy.Components.ProductAmountPicker.SetQuantityUnit(productDetails.quantity_unit_stock.id);
-				$('#display_amount').val(parseFloat(Grocy.UserSettings.stock_default_consume_amount));
+				$('#display_amount').val(parseFloat(productDetails.product.quick_consume_amount));
 				RefreshLocaleNumberInput();
 				$(".input-group-productamountpicker").trigger("change");
 
@@ -380,7 +380,7 @@ Grocy.Components.ProductPicker.GetPicker().on('change', function(e)
 	}
 });
 
-$('#display_amount').val(parseFloat(Grocy.UserSettings.stock_default_consume_amount));
+$('#display_amount').val(parseFloat(0)); //ergibt leider aerger mit popper.js beim oeffnen des dropdown... warum auch immer es funktioniert dennoch...
 $(".input-group-productamountpicker").trigger("change");
 Grocy.FrontendHelpers.ValidateForm('consume-form');
 
@@ -400,9 +400,11 @@ $('#consume-form input').keyup(function(event)
 	Grocy.FrontendHelpers.ValidateForm('consume-form');
 });
 
+//das Timeout wurde eingepflegt damit bei valider vorausgewaehlter Menge die Buttons von selbst aktiviert werden
 $('#consume-form select').change(function(event)
-{
+	{setTimeout(function(){
 	Grocy.FrontendHelpers.ValidateForm('consume-form');
+	}, 150)
 });
 
 $('#consume-form input').keydown(function(event)
--- a/views/productform.blade.php
+++ b/views/productform.blade.php
@@ -399,11 +399,11 @@
 			@php if($mode == 'edit') { $value = $product->quick_consume_amount; } else { $value = 1; } @endphp
 			@include('components.numberpicker', array(
 			'id' => 'quick_consume_amount',
-			'label' => 'Quick consume amount',
+			'label' => 'Standard Consume-Amount for this Product',
 			'min' => '0.' . str_repeat('0', $userSettings['stock_decimal_places_amounts'] - 1) . '1',
 			'decimals' => $userSettings['stock_decimal_places_amounts'],
 			'value' => $value,
-			'hint' => $__t('This amount is used for the "quick consume/open buttons" on the stock overview page (related to quantity unit stock)'),
+			'hint' => $__t('This amount is used as the Standard Amount for Consuming'),
 			'contextInfoId' => 'quick_consume_qu_info',
 			'additionalCssClasses' => 'locale-number-input locale-number-quantity-amount'
 			))
--- a/views/stocksettings.blade.php
+++ b/views/stocksettings.blade.php
@@ -114,7 +114,7 @@
 				</label>
 			</div>
 		</div>
-
+		<?php /*
 		<h4 class="mt-2">{{ $__t('Consume') }}</h4>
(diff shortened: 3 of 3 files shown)


berrnd:
Thank you, but I would rather add a checkbox "Use product default consume amount" below the existing stock setting "Default amount for purchase" and then only use the default consume amount if that option is checked - to keep both variants.

The existing handling makes at least practical sense for myself: I have the "Default amount for purchase" set to 0, because when I use the consume page, this enforces to enter an amount - of course this doesn't fit for the quick consume buttons on the stock overview page.

sleeplessnight2:
I had used this function in the same way - a 0 to make the field "necessary".
But that doesn't really have much to do with a "standard amount" anymore. That would be more like a function like "input required" or "Amount required" in the stock setting rather than a standard quantity. 0 is "nothing" and not a Amount ..
So we both "misused" the 0 so that the function does something different than intended ...

That's why I'm still looking for the sense of a _global_ standard set - except that it creates unnecessary steps in the input.
It is simply very unlikely that a 1 or 2 etc. will always fit most products. There are simply too many different amounts for that ...
It is not without reason that you have built in the function that a Amount can be linked to a barcode...

sleeplessnight2:
I have built in the mentioned function.

That would actually correspond to what you ask for and what I mean.

localization/de/strings.po
```diff
@@ -51,6 +51,15 @@ msgid_plural "%s Products"
 msgstr[0] "%s Produkt"
 msgstr[1] "%s Produkte"
 
+msgid "Standard Consume-Amount for this Product"
```
berrnd, line 54: Localizations are managed via Transifex as [mentioned in README](https://github.com/grocy/grocy#localization), the only (source) file which need to be updated with new strings is `/localization/strings.pot`

public/viewjs/consume.js
```diff
@@ -400,9 +415,11 @@ $('#consume-form input').keyup(function(event)
 	Grocy.FrontendHelpers.ValidateForm('consume-form');
 });
 
+//das Timeout wurde eingepflegt damit bei valider vorausgewaehlter Menge die Buttons von selbst aktiviert werden
```
berrnd, line 418: For the same reason we keep the public conversations in English, I think we don't need German code or comments...

berrnd (changes requested):

berrnd:
> That would actually correspond to what you ask for and what I mean.

Then I think you maybe misunderstood.
I thought about a checkbox below the current stock setting `Default amount for purchase` like `Use the products quick consume amount`. If checked, use the products quick consume amount, if not, use the existing `Default amount for purchase`.
This way it's possible to have both variants.

Currently it forces `0` if the new option is not checked...

sleeplessnight2:
Yes, I understood you correctly. But again I quote myself:
>..... "misused" the 0 so that the function does something different than intended ...
> 
> That's why I'm still looking for the sense of a _global_ standard set - except that it creates unnecessary steps in the input.
> It is simply very unlikely that a 1 or 2 etc. will always fit most products. There are simply too many different amounts for that ...
> It is not without reason that you have built in the function that a Amount can be linked to a barcode...

If this pointless function of a global set is to be used further, only so that it does the same afterwards as it already happens with my "0", it must be entered by someone else.

They are additional unnecessary lines of code.
Even you don't even use it ...

One should consider:
It has to be kept simple in your own home. Little effort is often worth a lot of money.

So if I spend 30 minutes with Grocy before I cook a dish - how long do you think Grocy will be used in the future?

Therefore: a standard amount makes no sense. It forces me to take 30 seconds longer per booking than necessary. Depending on the age of the client device and the hosting server, even longer ...

Scan - done. And the next. This is how it has to work if it is to stay interesting. We are not at work where it is normal to need 1 hour for a single booking in SAP ...

berrnd:
It's getting off-topic...

> If this pointless function of a global amount is to be used further, only so that it does the same afterwards as it already happens with my "0", it must be entered by someone else.

Nothing different when you prefill `0` automatically if the new option is not enabled, but you cannot change the prefilled value. However, we had that since a long time, why remove it - at least I find it useful, I'm sure someone else to. If you want to customize grocy to 100 % fit your likeings, you maybe need to maintain your own fork - I do that with other software too.

> They are additional unnecessary lines of code.

Sure, feel free to clean it up. I don't think too much about code style for grocy, because it's relaxing to not do do so for a hobby project.

> So if I spend 30 minutes with Grocy before I cook a dish - how long do you think Grocy will be used in the future?

I don't care about that at all or numbers (users of grocy) in general. First, grocy has features I need. Second, I create things I don't need, but I have fun to build it. I don't need to convince users. If grocy doesn't fit you, don't use it. grocy is not my job.

> Therefore: a standard amount makes no sense.

Then just use the new option and fill in a "Quick consume amount" for each product. I really don't understand what's your problem about the new option (thanks again for implementing it) while keeping optionally the old behaviour...

> Scan - done. And the next.

That's also possible, I use it directly in my kitchen (with a different grocy user on a alwas-on tablet which has the default consume amount set to `1` and "Scan mode" on). So I just scan an item once to remove it, or twice to remove 2. Maybe less than a second per product. I organize everything in my home with grocy and maybe spent 10 minutes per week for everything. Of course the main thing is about how you organize your products and quantity units and the rest. Many sub products, a lot of conversions makes things more complex and therefore needs more time to maintain it. If you need more detail-data, you need more time, as always.

And when we talk about time: Thanks again for building that, but I maybe would have need 5 to 10 minutes to built that myself. Now I spent a lot more minutes with that, for review and discussion. I'm glad for every pull request. But that's the same for more or less every pull request, however...

sleeplessnight2:
> Sure, feel free to clean it up. I don't think too much about code style for grocy, because it's relaxing to not do do so for a hobby project.

Das ist der Grund, warum ich keine Lust mehr auf Englishe Diskussionen habe. Google Translator verursacht unnötige Arbeit und hat hier ein dickes Missverständnis verursacht.

Ich meinte, das diese zusätzliche Funktion unnützen Code erzeugt. Ich meinte NICHT, das überall unnützer Code enthalten ist. Das würde ich nie wagen zu sagen, da ich gar nicht alle zusammenhänge kenne.

> Then just use the new option and fill in a "Quick consume amount" for each product. I really don't understand what's your problem about the new option (thanks again for implementing it) while keeping optionally the old behaviour...

Das Problem ist, das dass Buchungssystem beim Einkauf super aufgrund der Umrechnung Funktioniert.
Beim ausbuchen allerdings muss man die Bestandsmenge angeben, die nie wie beim EInkauf passt.

Sprich: ich buche mit standard 1 ein. Dies Funktioniert, weil ich angeben kann, was 1 Packung im Bestand ist. z.B. 500 gramm.
Beim Ausbuchen muss ich aber 500 gramm angeben, anstatt 1 Packung.
Deshalb macht beim ausbuchen eine globale standard Menge wie beim Einkauf kein Sinn. 

> If you want to customize grocy to 100 % fit your likeings, you maybe need to maintain your own fork - I do that with other software too.

Hab ich ja schon - ich wollte nur als Dank die Funktionen, die ich eingebaut und für nützlich in der Praxis empfunden habe, "zurück" geben.

"Pull request" ist dafür eigentlich völlig übertrieben. Es wäre besser, man könnte hier "vorschläge" machen, die du dir selbst kopieren musst, wenn daran interesse besteht anstatt gleich ein Merge zu machen zu müssen.

Mehr wollte ich nicht - lediglich sagen "hey, so habe ich mein eigenen Feature Request gelöst - wenn es etwas ist, was du gebrauchen kannst, übernehme es einfach und wenn es nichts ist, dann ignoriere es einfach".

Sorry, das ich deine Zeit geklaut hab.

berrnd (commented):
> I thought about a checkbox below the current stock setting `Default amount for purchase` (like `Use the products quick consume amount`). If checked, use the products quick consume amount, if not, use the existing `Default amount for purchase`.
This way it's possible to have both variants.

That didn't work, so I did this now myself in https://github.com/grocy/grocy/commit/2bdb6ab2d4aeb5df9dae4e8f8de9ab519958060d.

berrnd:
Alles gut. :)
Wenn Englisch das Problem ist, na dann sowas zukünftig doch besser auf Deutsch - aber Englisch ist ja halt how the Welt (und vor allem das Internet) spricht - und da dies hier (wie bei Open Source Projekten üblich) alles öffentlich ist, ist das, nicht nur meiner Meinung nach, definitiv besser. Und nein, grocy ist immer noch kein überwiegend deutsches Projekt von den Nutzern her.

Ich gehe mal davon aus, wie in https://github.com/grocy/grocy/commit/2bdb6ab2d4aeb5df9dae4e8f8de9ab519958060d gebaut, passt das dann auch für dich. So wie es du gebaut hast, wäre ja die "globale Standardmenge" nicht mehr möglich gewesen, so ist nun beides möglich. Weil ein Feature wegnehmen ist ja immer doof, irgendjemand hats bestimmt benutzt, das hier zum Beispiel ich.

sleeplessnight2:
> So wie es du gebaut hast, wäre ja die "globale Standardmenge" nicht mehr möglich gewesen, so ist nun beides möglich.

Ja richtig.
Lag auch daran, das ich die Option, eine Eingabe zu erzwingen sehr schön fand. Die wäre mit deinem _2bdb6ab_ zwar immer noch möglich, ist aber eher eine versteckte Funktion - d.H., wenn man es nicht mitgeteilt bekommt, weiß man nicht, das eine 0 eine Zwangseingabe zur Folge hat.
Ich mags gerne, wenn der User diese doch recht schöne Funktion (die Zwangs-Eingabe) präsentiert bekommt.

In dem Fall könnte man dies sogar als eine Dritte Option einpflegen. "Force to enter a Number" oder so?!? Nur so als Idee:

```
[Checkbox] Alternative Usage for the Amount:
                 [RadioButton1] Use Products "Quick consume amount"
                 [RadioButton2] Force to enter a Value
```

Wenn das nichts für dein Projekt ist, dann sag mir hier eben kurz bescheid, da ich mir das sonst selbst einbaue. Das Gefällt mir so nämlich am allerbesten...

berrnd:
Ich denke, das reicht so, ist ja auch nicht die einzige Stelle mit solch einer Steuerung ("Default best before days" zum Beispiel) - ein Tooltip als Erklärung könnte man noch ergänzen. Andere Seiten haben ja jetzt im Vergleich zur letzten Version deutlich mehr Hilfe-Tooltips, die Settings-Seiten habe ich noch nicht durchgeschaut...

PR #1177, opened 2020-12-12
Correct typo in changelog

MrJohnZoidberg:
I saw that because the syntax hightlighting on GitHub only sees one `code` apostrophe and thus colors the whole following text wrong.


berrnd (approved):
Thanks! 🙂

PR #1185, opened 2020-12-14
Splitting up shoppinglist and it's print view

edenhaus:
Hi @berrnd,

first thanks for all your effort to improve grocy. I love it.

This is a draft for splitting it up the shopping list and it's print view. I have split them up, so it's possible to have different tables config for each of them. For example on the shopping list I will show a user specified column but I don't want this column to be printed.

Currently this is a draft with still some todos. @berrnd can you please give it a look and give me feedback. As I'm new to php it can be that the whole code is hacky... :) Before going further it would be great to have some feedback.

Short summary:
- Created a new component `shoppinglisttable.blade.php` and correspondig javascript file. The js file contains only the basic functionality, which is needed for both. Left the most js functions in `shopponglist.js`
- Use the new component in the shoppinglist view
- Created a new view `shoppinglistprint.blade.php`. When the page is loaded `window.print` is called.
- Added a button to the `shoppinglistsettings`, where the `shoppinglistprint` view is open in the preview mode (by adding the url parameter `preview`). This mode is can be used to customize which columns should be showed...

Todos:
- [x] Add checkbock to enable/disable row grouping. I will add it for all datatables, which have rowgroup configurated. #1189 
- [x] Extend the datatables state save/load methods to store also the rowgroup settings. #1189 
- [ ] The design of the print is currently wrong. The headers and content is not aligned correctly.

Tanks in advance for the feedback and any help

### DIFF
--- a/controllers/StockController.php
+++ b/controllers/StockController.php
@@ -332,7 +332,14 @@ class StockController extends BaseController
 			$listId = $request->getQueryParams()['list'];
 		}
 
-		return $this->renderPage($response, 'shoppinglist', [
+		$template = 'shoppinglist';
+
+		if (isset($request->getQueryParams()['print']))
+		{
+			$template = 'shoppinglistprint';
+		}
+
+		return $this->renderPage($response, $template, [
 			'listItems' => $this->getDatabase()->shopping_list()->where('shopping_list_id = :1', $listId),
 			'products' => $this->getDatabase()->products()->where('active = 1')->orderBy('name', 'COLLATE NOCASE'),
 			'quantityunits' => $this->getDatabase()->quantity_units()->orderBy('name', 'COLLATE NOCASE'),
--- a/localization/strings.pot
+++ b/localization/strings.pot
@@ -2008,3 +2008,6 @@ msgstr ""
 
 msgid "Show on stock overview page"
 msgstr ""
+
+msgid "Open print preview"
+msgstr ""
--- a/public/viewjs/components/shoppinglisttable.js
+++ b/public/viewjs/components/shoppinglisttable.js
@@ -1,0 +1,41 @@
+var collapsedGroups = {};
+
+//tableId is desclared in shoppinglisttable.blade.php
+
+var shoppingListTable = $(tableId).DataTable({
+	'order': [[3, 'asc']],
+	"orderFixed": [[1, 'asc']],
+	'columnDefs': [
+		{ 'orderable': false, 'targets': 0 },
+		{ 'searchable': false, "targets": 0 },
+		{ 'visible': false, 'targets': 1 }
+	].concat($.fn.dataTable.defaults.columnDefs),
+	'rowGroup': {
+		dataSrc: 1,
+		startRender: function(rows, group)
+		{
+			var collapsed = !!collapsedGroups[group];
+			var toggleClass = collapsed ? "fa-caret-right" : "fa-caret-down";
+
+			rows.nodes().each(function(row)
+			{
+				row.style.display = collapsed ? "none" : "";
+			});
+
+			return $("<tr/>")
+				.append('<td colspan="' + rows.columns()[0].length + '">' + group + ' <span class="fa fa-fw ' + toggleClass + '"/></td>')
+				.attr("data-name", group)
+				.toggleClass("collapsed", collapsed);
+		}
+	}
+});
+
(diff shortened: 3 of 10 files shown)


berrnd:
Thanks!

I don't know if a separat page/component is necessary for this.
When it's a about printing columns or not, why not add this generically for all tables by a new option "Hide when printing" or something similar next to checkbox on the "Hide/View column"-popup for example...

> Add checkbock to enable/disable row grouping. I will add it for all datatables, which have rowgroup configurated.
> Extend the datatables state save/load methods to store also the rowgroup settings.

That's a good idea, maybe also generically by something like "Group by this column".

edenhaus:
> Thanks!
> 
> I don't know if a separat page/component is necessary for this.
> When it's a about printing columns or not, why not add this generically for all tables by a new option "Hide when printing" or something similar next to checkbox on the "Hide/View column"-popup for example...

I was trying to do that, but datatables don't support to load or set a state manually. When we use a different state for printing, we can also customize the columns order and co. Do you know a way to set a state manually for datatables?
I search around and didn't find something not hacky...
Example: I would love to change the columns order for printing only. The amount column should come before the product column.

> That's a good idea, maybe also generically by something like "Group by this column".

I will open for that one a seperate PR so it's easier to review the changes. Hopefully I can implement it this week

berrnd:
Basically should be all to apply the Bootstrap class `d-print-none` to the column which should not be printed.

If there should be also a whole different column ordering, ok - then it's maybe more a complete separate table...

Personally I don't understand why I should really print the shopping list at all, using your smartphone/grocy directly while shopping is much more efficient, you can even start adding things to stock while waiting at the check-out - I use it like that every week...

Another approach maybe is to just generate a PDF directly:
https://datatables.net/extensions/buttons/examples/html5/columns.html

> Do you know a way to set a state manually for datatables?

Also haven't found anything, but `.draw()` seems to load the state - somehow providing another state in `stateLoadCallback` then would maybe work...

edenhaus:
> Personally I don't understand why I should really print the shopping list at all, using your smartphone/grocy directly while shopping is much more efficient, you can even start adding things to stock while waiting at the check-out - I use it like that every week...

Since corona and the lockdown in Italy our supermarkets prefer to deliver and therefore we send them each day a pdf of our shopping list.

> Another approach maybe is to just generate a PDF directly:
> https://datatables.net/extensions/buttons/examples/html5/columns.html

I will give it a try.

edenhaus:
As the extension buttons doesn‘t support row group yet and it‘s not trivial to give the user the option the configure the layout I come to the point to close this draft.
The best would be a template as requested in #740

edenhaus:
Hi @berrnd,

thanks for adding row groups in the shopping print with 832141a718564b9156f9269429db5ca32f20090c
I was thinking to open this PR again and update to get the last changes.

Splitting has in my opinion some advances over your solution:
- We can customize the print table in the same manner as the shoppinglist table, meaning order/sorting and also show/hide the columns we want
- We must only change at one point if the shoppinglist table changes in the future.
- The print table is only processed, when we really want to print it. Currently it will process each time, I know currently this is not a issue.

What do you think?
I will only work on it, if you think it's worth it :)

berrnd:
It's already kind of splitted (there is the separate table with id `shopping-list-print-shadow-table` for printing).
So showing that table on the new popup before printing, why not.

For me personally this is not worth it, but just because I don't need it - I'm open for everything which is useful for someone as long as the existing functionality will not break (unless the existing things made totally no sense) and as long not huge (external) dependencies are added unless _really_ needed and not solvable without them.

All things mentioned in #740 are done by the recent changes, I think.

For the shopping list, the most logical thing for me would be to use the existing table also for printing - would need just some improvement to hide the first column with the buttons and so on...

> The print table is only processed, when we really want to print it. Currently it will process each time, I know currently this is not a issue.

Unless it's really noticable, I don't care to much about performance. grocy is meant to manage a home, if you often have a shopping lists with hundreds of items, ok, then maybe it's worth to think about such things...
(The only thing which I want to improve everywhere is to avoid array searching for dependency objects name lookups, like done already for the stock overview page or today in https://github.com/grocy/grocy/commit/b2d7003335c43f22fb517f19b7fcfc09e1fa9334 for the shopping list, as this really makes a difference.)

edenhaus:
That's true. I leave it as it is and help you improve other things.

Frohe Weihnachten aus Südtirol :)

Issue #1168, opened 2020-12-10
Feature request: Let admins specify which columns are shown/hidden by default


edenhaus:
With #1058 we got the possibility to hide columns and since the commit dd8fa5ff66a3533dba00a191d299564f9c35d6db this settings will be stored by user on the server.

I think it would be great, that the admin has a possibility to set which columns are shown/hidden per default.
The button is only visible if the current user has enough rights and my idea was to add another button next to the OK one.
![grafik](https://user-images.githubusercontent.com/26537646/101766025-71ba8080-3ae2-11eb-883f-2afef24ee5a8.png)

On the other hand the method `public function GetUserSetting($userId, $settingKey)` in `UsersService.php` will return the default settings if no user ones are specified and default one exists.

berrnd:
> On the other hand the method `public function GetUserSetting($userId, $settingKey)` in `UsersService.php` will return the default settings if no user ones are specified and default one exists.

Please keep it to one topic per request, keeping track of things is hard otherwise - I've splitted that into #1169 (and also consider this a bug, thanks for testing).

PR #1189, opened 2020-12-15
Add row group customization

edenhaus:
Add the possibility to change the row group of a table similar to hide/view columns. The row group configuration will be stored with the other state settings of the table in the user settings.

Screenshot of the shoppinglist customizations:
grafik

Selecting "none" will disable the rowGroup feature and disable the fixed order.
Selecting a column will group it after the column and also activate a fixed order for that column. (In the code I have found, that when rowGroup is activated also fixedOrder on that column was applied)

Currently the Group by section is shown only if rowGroup extension is added to table, which is currently the case for shoppinglist, tasks, productform, recipeform.
@berrnd Should we enable the possibility for all tables?

### DIFF
--- a/controllers/UsersApiController.php
+++ b/controllers/UsersApiController.php
@@ -219,6 +219,19 @@ class UsersApiController extends BaseApiController
 		}
 	}
 
+	public function DeleteUserSetting(\Psr\Http\Message\ServerRequestInterface $request, \Psr\Http\Message\ResponseInterface $response, array $args)
+	{
+		try
+		{
+			$value = $this->getUsersService()->DeleteUserSetting(GROCY_USER_ID, $args['settingKey']);
+			return $this->EmptyApiResponse($response);
+		}
+		catch (\Exception $ex)
+		{
+			return $this->GenericErrorResponse($response, $ex->getMessage());
+		}
+	}
+
 	public function __construct(\DI\Container $container)
 	{
 		parent::__construct($container);
--- a/grocy.openapi.json
+++ b/grocy.openapi.json
@@ -1299,6 +1299,38 @@
 						}
 					}
 				}
+			},
+			"delete": {
+				"summary": "Deletes the given setting of the currently logged in user",
+				"tags": [
+					"Current user"
+				],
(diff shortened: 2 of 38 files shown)


edenhaus:
After rethinking what is the best translation of "None" in German, I think maybe we should rename it to "Deactivated" or what you think?

berrnd:
Cool, thanks - looks good from a first sight. :)

Will give it a short test tomorrow or so - and I think "None" is fine and it would be also ok to enable this for all tables (some columns will not make sense to group by them, but I think we don't need to limit that).

With more table options your other request #1168 is probably getting more important and maybe also a "Reset" button.

edenhaus:
I will add it for all tables this afternoon :)

For the Reset button there exist a method `dataTable.state.clear()` but I need to give it a closer look.

edenhaus:
ok there is still a bug somewhere. I was trying to add it for all tables and found out, that after the loadStateEvent a saveStateEvent is fired, where the rowGroup is set to the default value.

Example for reproduction:
- Open Shoppinglist and set row group to Amount
- Refresh the page and the row group is setting back to Product group

edenhaus:
Took me some time to get a working solution. RowGroup unfortunately cannot be managed via the state events, but I have implemented via the preDrawCallback.

Also I have activated rowGroup for all tables and also implemented a reset button.
grafik

After clicking on reset a confirmation window will be open. Confirming will delete the usersetting entries in the db and reload the page. The later one is required as `state.clear()` is only calling `stateSaveCallback` but not refreshing the table afterwards.
grafik

I have tested it with some tables without any problems.

berrnd (approved):
Works, thanks a lot! :)

> How is the openapi file generated? Manually?

Yes, I've added the new endpoint `DELETE /user/settings/{settingKey}`.

berrnd:
I've also added that row groups are collapsible (like we already had it for the shopping list) for alle tables in https://github.com/grocy/grocy/commit/cc2ea93313add269d7f36cce0ae404cc7fc20128. 🙂

PR #1190, opened 2020-12-16
Removing of resize event

PhyberApex:
Hey,

I removed that resize event as it get's thrown every time you scroll on a mobile device. Which prevented me from actually viewing the list view of the calender on mobile devices.

Let me know if you think it is still needed tho!

~Cheers

### DIFF
--- a/public/viewjs/calendar.js
+++ b/public/viewjs/calendar.js
@@ -43,17 +43,3 @@ $("#ical-button").on("click", function(e)
 		}
 	);
 });
-
-$(window).on("resize", function()
-{
-	// Automatically switch the calendar to "basicDay" view on small screens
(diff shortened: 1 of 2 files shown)


berrnd:
The idea was just to automatically switch the view for mobile as the default month view looks like that on small screen and doesn't give that much overview...

grafik

So I still think that makes sense, but should not be fired on every `resize` event, just once.
(Is more or less the same as for the meal plan, origin issue was #372.)

PhyberApex:
The event should fire only once now. I also added the same thing to the mealplan.
Hope that works out now.

~Cheers

berrnd:
Great, thanks! 🙂

berrnd (approved):

PR #1199, opened 2020-12-19
Fix: Show print button also when using only one shoppinglist

edenhaus:
Show the print button also when `GROCY_FEATURE_FLAG_SHOPPINGLIST_MULTIPLE_LISTS` is set to false


berrnd:
Thanks, but this then already hid the button on small screens - fixed it in https://github.com/grocy/grocy/commit/77e842a736a764a93aa13b1d86bba57d7ad34a15.

PR #1201, opened 2020-12-20
fix missing >

edenhaus:
added `>` on line 314 otherwise the header was not showed correctly.

There exists still a issue that the header is not aligned correctly with the rest of the table, but I didn't find a solution. Maybe you know what's wrong with it

### DIFF
--- a/views/shoppinglist.blade.php
+++ b/views/shoppinglist.blade.php
@@ -311,13 +311,12 @@
 </div>
 
 <div class="d-none d-print-block">
-	<div id="print-header"
-		<h1
-		class="text-center">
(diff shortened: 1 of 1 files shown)


berrnd (approved):
👍

berrnd:
> There exists still a issue that the header is not aligned correctly with the rest of the table, but I didn't find a solution. Maybe you know what's wrong with it

Looks ok here ... the header is centered, the rest not - of course we could also center or left-align everything...

grafik

edenhaus:
Strange that I get another result. Tested with chrome and firefox and I get the following:

grafik

As you can see, the amount column header is not aligned correctly with the columns rows...
Tested it also on Chrome on Android with the same result. I used for the tests https://de.demo-prerelease.grocy.info/shoppinglist

Issue #1206, opened 2020-12-22
Bug: No stock overview after upgrade to 3.0.0


Forceu:
After upgrading to 3.0.0 from 2.7.1 no stock is displayed on the overview. The stock is logged however.

Errors in HTML console:

```
Uncaught RangeError: Invalid currency code : â¬
    at Number.toLocaleString (<anonymous>)
    at HTMLSpanElement.<anonymous> (grocy.js?v=3.0.0:590)
    at Function.each (jquery.min.js?v=3.0.0:2)
    at S.fn.init.each (jquery.min.js?v=3.0.0:2)
    at RefreshLocaleNumberDisplay (grocy.js?v=3.0.0:581)
    at grocy.js?v=3.0.0:617
```

```
Uncaught TypeError: Cannot read property 'targets' of undefined
    at kb (jquery.dataTables.min.js?v=3.0.0:15)
    at HTMLTableElement.<anonymous> (jquery.dataTables.min.js?v=3.0.0:91)
    at Function.each (jquery.min.js?v=3.0.0:2)
    at S.fn.init.each (jquery.min.js?v=3.0.0:2)
    at S.fn.init.l [as dataTable] (jquery.dataTables.min.js?v=3.0.0:84)
    at S.fn.init.h.fn.DataTable (jquery.dataTables.min.js?v=3.0.0:168)
    at stockoverview.js?v=3.0.0:1
```

The directory was restored from backup before upgrading, I don't know if this is relevant

berrnd:
That basically means your `CURRENCY` setting is not a valid ISO 4217 currency code.

https://github.com/grocy/grocy/blob/e3504464e52b43a78e602f9c5dce024a93f0f3b3/config-dist.php#L33-L37

(There is a new, hidden by default, column to show the stock value - this column uses the currency symbol to properly format the value).

Forceu:
Thanks, that fixed it! Maybe check for that should be included as well? Because in my config it said the following:

```
# To keep it simple: grocy does not handle any currency conversions,
# this here is used to format all money values,
# so can be anything (e. g. "USD" OR "$", doesn't matter...)
Setting('CURRENCY', '€');
```

berrnd:
[...]

PR #1215, opened 2020-12-23
Added configuration validator class

Forceu:
Adds a class similar to PrerequisiteChecker that checks if the config.php does not contain any invalid entries (not all entries are checked, only the ones that are most likely to be invalid).

Related to issue #1206


berrnd (approved):
Great, thanks a lot! :)

PR #1218, opened 2020-12-24
Fix openapi for userentities

edenhaus:
Currently the openapi allows only the predefined enum value (`#/components/internalSchemas/ExposedEntity`) as parameter `entity` on the api call `/userfields/{entity}/{objectId}`.
When using user entities, they will use `userentity-{name}` as entity, which would be a violation of the openapi schema.

Extended the parameter `entity` of `/userfields/{entity}/{objectId}` to allow also string with the following pattern ` ^userentity-`.
Is the openapi generated or is this manual change fine?

### DIFF
--- a/grocy.openapi.json
+++ b/grocy.openapi.json
@@ -571,7 +571,14 @@
 						"required": true,
 						"description": "A valid entity name",
 						"schema": {
-							"$ref": "#/components/internalSchemas/ExposedEntity"
+							"oneOf": [
+								{
+									"$ref": "#/components/internalSchemas/ExposedEntity"
+								},
+								{
+									"$ref": "#/components/internalSchemas/ExposedUserEntity"
+								}
+							]
 						}
 					},
 					{
@@ -620,7 +627,14 @@
 						"required": true,
 						"description": "A valid entity name",
 						"schema": {
-							"$ref": "#/components/internalSchemas/ExposedEntity_NotIncludingNotEditable"
+							"oneOf": [
+								{
+									"$ref": "#/components/internalSchemas/ExposedEntity_NotIncludingNotEditable"
+								},
+								{
+									"$ref": "#/components/internalSchemas/ExposedUserEntity"
+								}
+							]
 						}
 					},
 					{
@@ -3941,6 +3955,10 @@
 					"stock_current_locations",
 					"api_keys"
 				]
+			},
+			"ExposedUserEntity": {
+				"type": "string",
+				"pattern": "^userentity-"
 			},
 			"ExposedEntityNoListing": {
 				"type": "string",


berrnd:
Thanks.

`grocy.openapi.json` is mostly handcrafted, but some things are dynamically changed when serving the file, is done here:
https://github.com/grocy/grocy/blob/d8069c569e4700cf9d51dedcc2eeac3da30bb936/controllers/OpenApiController.php#L29-L67

I think we can do the same there for all userentities, means adding them to `#/components/internalSchemas/ExposedEntity`.

edenhaus:
> Thanks.
> 
> `grocy.openapi.json` is mostly handcrafted, but some things are dynamically changed when serving the file, is done here:
> 
> https://github.com/grocy/grocy/blob/d8069c569e4700cf9d51dedcc2eeac3da30bb936/controllers/OpenApiController.php#L29-L67
> 
> I think we can do the same there for all userentities, means adding them to `#/components/internalSchemas/ExposedEntity`.

As you can generate from the openapi file code (e.g. python client), I think the better approach is to specify a general openapi file, which is valid for all instances.
With the current file it is not possible to use https://editor.swagger.io/ to generate some code.

Additional we can for sure dynamically adjust the api in the method above, but in my opinion this should be more a nice to have as the way to go.

When loading the openapi in https://editor.swagger.io/ I get the following error:
```
Structural error at components
should NOT have additional properties
additionalProperty: internalSchemas
Jump to line 2688
```

If I have your okay, I would make the file specification conform and remove the adjustments in https://github.com/grocy/grocy/blob/d8069c569e4700cf9d51dedcc2eeac3da30bb936/controllers/OpenApiController.php#L48-L64

berrnd:
If you remove the dynamic generation of the `internalSchemas`, either everything needs to be manually put into `grocy.openapi.json` (the "Generic entity interactions" endpoints have different entities to be used, some are not editable, and so on) or no enum can be used. Both is worse than how it's currently done, I would say.

`internalSchemas` instead of `schemas` was just used because then they don't show up at the bottom of the Swagger UI page. However, the integrated Swagger UI doesn't complain about that. If that's all what currently breaks something, then ok, why not just use `schemas`...

edenhaus:
> If you remove the dynamic generation of the `internalSchemas`, either everything needs to be manually put into `grocy.openapi.json` (the "Generic entity interactions" endpoints have different entities to be used, some are not editable, and so on) or no enum can be used. Both is worse than how it's currently done, I would say.

That's true but I think I have probably a better solution. We rename the file to `template_grocy.openapi.json` which stay at it is.
We add then a script, which is called by the release script or even better by the ci (if we add a ci), which generates the final version of the file. As the dynamic generation stays the same for a whole release, I think it would be better to generate it with the ci. What you think?

berrnd:
So the end result is then the same file which is currently served at [`/api/openapi/specification`](https://demo.grocy.info/api/openapi/specification) (and also feeds the Swagger UI instance at [`/api`](https://demo.grocy.info/api))? If so, what's the benefit then?

Introducing a CI pipeline only for this is again something that sounds like too much overhead for me...

Currently I'm the only one who needs "to do releases" and my "CI pipeline" is a [simple batch file](https://github.com/grocy/grocy/blob/master/.devtools/create_release_package.bat) which works perfectly since v0.1.0 so far.

grocy is also about self-hosting, a base principial for this (and especially for me) is to keep external (service) dependencies as small as possible...

berrnd:
A CI pipeline will definitely not be added for this.

Closing this for now, I've created #1516 to track adding userentities to `grocy.openapi.json` in the same way other dynamic things are currently handled.

Issue #214, opened 2019-04-25
Convert UTC timestamps to local time in stock journal front end


kingo55:
It appears UTC times are compared with local times in the UI - therefore items added in the stock journal misrepresent the times at which the entries occurred in the users' local time.

As an example, from GMT +10, journal entries made NOW will show in the interface as being made 10 hours in the past:

![Screenshot_20190425_184327](https://user-images.githubusercontent.com/2361388/56722848-32171780-678b-11e9-8ff9-80aa30b3fcda.png)

berrnd:
Currently I assume everywhere that all datetimes in the database are in the users local time.

So the timezone for PHP needs to be configured accordingly. SQLite takes it from the operating system setting, as far as I know...

Maybe it's best to add a `config.php` setting for the timezone and use/set it then for all places (when possible) to keep it consistent.

kingo55:
Wouldn't UTC timestamps be more suitable for storage?

I.e. Handle daylight savings / timezone changes on the client side.

berrnd:
Maybe, but the user time zone has to be known on the server side anyway - displayed times are partially rendered through PHP, on other places with JavaScript...

crossan007:
👍  for storing DateTime as UTC in the database and letting the user select their timezone.

rotlasks:
line 475 of 
.grocy-desktop\runtime-dependencies\2.3.1\php\php.ini in the desktop application should be edited to your TZ;
I cannot find the appropriate line in docker files;
 /etc/php8/php.ini does not seem to be in use / changes have no effect

berrnd:
> .grocy-desktop\runtime-dependencies\2.3.1\php\php.ini in the desktop application should be edited to your TZ;

Thanks for mentioning that, of course grocy-desktop should automatically do that, since it's supposed to do everything needed to run grocy properly. But that's off-topic here, I've created grocy/grocy-desktop#41 to track that.

PR #1223, opened 2020-12-28
Added /system/time API call

Forceu:
As Grocy is using local time for all database entries, I needed a call to get the server's timezone.

This also outputs local time and UTC and an offset can be applied, which can be useful for external programs.

### DIFF
--- a/controllers/SystemApiController.php
+++ b/controllers/SystemApiController.php
@@ -46,8 +46,9 @@ class SystemApiController extends BaseApiController
 	public function GetSystemTime(\Psr\Http\Message\ServerRequestInterface $request, \Psr\Http\Message\ResponseInterface $response, array $args)
 	{
 		$offset = 0;
-		if (isset($args['offset']))
-			$offset = $args['offset'];
+		$params = $request->getQueryParams();
+		if (isset($params['offset']))
+			$offset = $params['offset'];
 		return $this->ApiResponse($response, $this->getApplicationService()->GetSystemTime($offset));
 	}
 
--- a/controllers/SystemApiController.php
+++ b/controllers/SystemApiController.php
@@ -45,11 +45,26 @@ class SystemApiController extends BaseApiController
 
 	public function GetSystemTime(\Psr\Http\Message\ServerRequestInterface $request, \Psr\Http\Message\ResponseInterface $response, array $args)
 	{
-		$offset = 0;
-		$params = $request->getQueryParams();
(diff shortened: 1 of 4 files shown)


berrnd:
Thanks.

Related to #214.

If the users (local) time and the one from the server (OS configuration for SQLite + `php.ini` for the PHP runtime) is not the same, there will still be some differences here and there (I guess), just using the server is IMHO not enough for everything. At least, because SQLite always takes the OS timezone, this here will return (most probably) the PHP settings...

Forceu:
I only checked for the products `row_created_timestamp`, which is created with a timestamp from PHP, not sqlite3. I can however add another output for the sqlite3 time (as there can be in some cases differences between PHP/sqlite timezone).

The reason why I implemented the call is actually that the timezone needs to be known if the API is called from a different server or environment. Otherwise the timestamps cannot be used properly.

Personally I agree that a db timestamp would be better in UTC, as it makes a lot of things easier.

berrnd:
It's currently unfortunately not clear which place use timestamps from PHP or SQLite, there are a lot of views (SQLite), also the `row_created_timestamp` is not set explicitly anywhere, but has a default value (so must be also SQLite for those), but the comparison for due/expired products happens for example via PHP...

So let's that endpoint for now if that helps, of course #214 would be better.

Forceu:
Oh you are right, I must have mixed something up there. I just saw that they are created with sqlite `datetime('now', 'localtime')` In that case I will definitely add another output value with the Sqlite3 time.

berrnd (approved):

Forceu:
With the commit 16f400447ac9423d40ca3f82c09c57fc80f94828 the API now returns the error `Query parameter "offset" is not a valid integer` when 0 is supplied as an integer. Another reason why I somewhat dislike PHP :D

Change `if (!filter_var($params['offset'], FILTER_VALIDATE_INT))` to `if (!(filter_var($params['offset'], FILTER_VALIDATE_INT) === 0 || filter_var($params['offset'], FILTER_VALIDATE_INT)))` and it should work again

berrnd:
Thanks - [`filter_var`](https://www.php.net/manual/en/function.filter-var.php) returns `false` when the check failed, so the correct way is to check for `=== false` - optimized that for alle places where this is currently used in https://github.com/grocy/grocy/commit/b6139a69918c2425578603341ebc8db3ade7e3ed.

PR #1225, opened 2020-12-28
Fix for #1223

Forceu:
With the commit 16f4004 the API returned the error Query parameter "offset" is not a valid integer when 0 is supplied as an integer.


berrnd:
Is now done by https://github.com/grocy/grocy/pull/1223#issuecomment-751866894.

Forceu:
Ah interesting, yes that does indeed work. I copied the filter from https://www.w3schools.com/php/filter_validate_int.asp, I wonder why they did it that complicated. `=== false` works as well.

Issue #1216, opened 2020-12-24
/api requires authentication


blake1029384756:
I am using 'ReverseProxyAuthMiddleware' for authentication. I set my auth provider to bypass authentication on the /api endpoint so grocy-android can access it as suggested [here](https://github.com/patzly/grocy-android/issues/255). 

This results in an "invalid username" when accessing the api
`array(0) {
}
[...]

PR #1277, opened 2021-01-17
Show "manage user" menu entry also when the auth view is diabled

edenhaus:
I don't know, why in first place the menu entries for managing users and api keys are disabled, when the auth view is disabled.
I started to use the `ReverseProxyAuthMiddleware` and handle the login with authelia. With #1216 it's now possible to use api keys with `ReverseProxyAuthMiddleware` and therefore we need again the menu entries.

Workaround until a new version is build, it is possible to to access the missing funtionality by inserting the direct url `/users` or `/manageapikeys`

### DIFF
--- a/views/layout/default.blade.php
+++ b/views/layout/default.blade.php
@@ -628,13 +628,11 @@
 							<i class="fas fa-user-cog"></i> {{ $__t('User settings') }}
 						</a>
 						<div class="dropdown-divider"></div>
-						@if(GROCY_SHOW_AUTH_VIEWS)
 						<a class="dropdown-item discrete-link permission-USERS_READ"
 							href="{{ $U('/users') }}"><i class="fas fa-users"></i>&nbsp;{{ $__t('Manage users') }}</a>
 						<div class="dropdown-divider"></div>
 						<a class="dropdown-item discrete-link"
 							href="{{ $U('/manageapikeys') }}"><i class="fas fa-handshake"></i>&nbsp;{{ $__t('Manage API keys') }}</a>
-						@endif
 						<a class="dropdown-item discrete-link"
 							target="_blank"
 							href="{{ $U('/api') }}"><i class="fas fa-book"></i>&nbsp;{{ $__t('REST API & data model documentation') }}</a>


berrnd:
While it does not hurt to show those menu entries in any case, that's not complete since we not only have `ReverseProxyAuthMiddleware`.

- No user changes and API keys are needed when using LDAP authentication
- The same is the case when authentication is disabled at all

edenhaus:
>     * No user changes and API keys are needed when using LDAP authentication

When using LDAP as authentication the menu entries will be shown.


Currently in the code there exists only two places where `GROCY_SHOW_AUTH_VIEWS` is set to false:

https://github.com/grocy/grocy/blob/5ecd3a585ecc34828a793bf49e2cf873bc58ddf6/middleware/ReverseProxyAuthMiddleware.php#L15-L18

https://github.com/grocy/grocy/blob/5ecd3a585ecc34828a793bf49e2cf873bc58ddf6/app.php#L25-L33

@berrnd should I replace it with `@if(!GROCY_DISABLE_AUTH)`?

berrnd:
> When using LDAP as authentication the menu entries will be shown.

Ok, also not so good...

So this better has the rethought, what is needed in which case - when we show the password fields for example most probably someones expects that by that also the password on the correspanding backend will be changed...

edenhaus:
> So this better has the rethought, what is needed in which case - when we show the password fields for example most probably someone expects that by that also the password on the correspanding backend will be changed...

Yeah that`s right.
After giving it a closer look also changing the username should be disabled for reverse proxy and ldap. Do we really need it?
I never saw that a systems is allowing to change your username and don't see any reason to do it.

berrnd:
Depends, a reverse proxy only provides the username, never the first and last name. so you mabye want to add that in grocy.
LDAP can provide optionally the first and last name, but maybe you want to change it for grocy.

So probably the password and the username field should be always hidden (or better disabled), first/last name is ok to be edited.

berrnd:
Closing this for now, feel free to reopen it if you want to continue on this - just showing all "auth views" is not the solution as mentioned above.

Issue #1289, opened 2021-01-25
Bug: When editing a shopping list entry, it cannot be saved without entering a note or reselecting the item


Forceu:
To reproduce:

1. Add entry to shopping list
2. Click on edit entry
3. It cannot be saved and the message "A product or a note is required" is shown

berrnd:
Most probably caused by the same which affects #1261.

Forceu:
Ah thanks, I did not see that issue

berrnd:
There were changes lately for v3.0.1 that either a product or a note should be required for shopping list items (no field was required before), references https://github.com/grocy/grocy/commit/9f1692e31f996bf75407354a5ac150284203571c and #1226.

Issue #1261, opened 2021-01-08
Bug: add to shoppinglist do not accept the selected item


liffez:
When adding a item to the shopping list directly from stockoverview the item is not updated correctly in the "add to shoppinglist" view.
The shoppinglist state that "A product or a note is required"
This mean that you cannot just enter the amount and click the save button. 
However if you use "tap" to navigate to the save button you can activate it by using enter. 
But if you reload the item or add a note you will be able to click the save button.

PR #1297, opened 2021-01-30
Fixed shoppinglist JS, fix for #1289 and #1261

Forceu:
Now the fields are properly checked and the error does not show up anymore.
The only minor issue is that the save button is disabled if editing a shoppinglist note without a product, but as soon as the note is changed, the save button is enabled again. However as no errormessages are shown and it kind of makes sense that you cannot save the item without any changes first I did not add another validation in this PR.


berrnd (approved):
Thank you!

PR #1332, opened 2021-02-14
Hide unsuitable fields from printed recipe pages

### LINKED ISSUE #1330: Feature request: Cleaner recipe printing
Currently recipe printing includes stock information ("[X] Enough in stock" etc.) and there is an interactive question mark after the "Costs" title text. I think these should be removed from the printed version of the recipe.

Ape:
Resolves #1330.


berrnd (approved):
👍, thanks!

PR #1336, opened 2021-02-14
More night mode improvements

KTibow:
My PR got merged overnight (thanks!) but there was still a bit more stuff for me to do. This adds some of that.
Old:
image
New:
image
Old:
image
image
New:
image
image
Old:
image
New:
image
(Manually changed)
Old:
image
New:
image
Old (that's the version in my last pr):
image
New:
image
Don't worry about my API key, they just say "Hi"
Old:
image
New:
image

### DIFF
--- a/public/css/grocy_night_mode.css
+++ b/public/css/grocy_night_mode.css
@@ -42,16 +42,20 @@ body.night-mode {
 	border-color: #1ed1ee !important;
 }
 
+.night-mode .btn-outline-info:hover {
+	color: #e1e1e1 !important;
+}
+
 .night-mode .btn-info {
 	color: #c1c1c1;
-	background-color: #07373f;
-	border-color: #07373f;
+	background-color: #0c5f6e;
+	border-color: #0c5f6e;
 }
 
 .night-mode .btn-warning {
 	color: #c1c1c1;
-	background-color: #473604;
-	border-color: #473604;
+	background-color: #a9810a;
+	border-color: #a9810a;
 }
 
 .night-mode .btn-danger {
@@ -107,15 +111,15 @@ body.night-mode {
 }
 
 .night-mode .bg-warning {
-	background-color: #473604!important;
+	background-color: #473604 !important;
 }
 
 .night-mode .bg-info {
-	background-color: #07373f!important;
+	background-color: #07373f !important;
 }
 
 .night-mode .bg-danger {
-	background-color: #471116!important;
+	background-color: #471116 !important;
 }
 
 .night-mode .form-control:focus {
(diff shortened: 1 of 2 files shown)


berrnd:
Great, thanks again - will leave this open / merge it at latest before the next release in case you find more... 🙂

public/css/grocy_night_mode.css
```diff
@@ -312,3 +312,9 @@ body.night-mode {
 .night-mode .text-muted {
 	color: #8f9ba5 !important;
 }
+
+.night-mode .qr-code {
+	background-color: #eee;
+	padding: 10px;
+	border-radius: 10px;
+}
```
sur5r, line 320: I was just going to open a PR for the QR code in night mode as well.
The quiet zone (padding) should be 32 pixels, though, as the standard mandates 4 modules of quiet zone. The QR code is generated with a scale factor of 4 which results in a module size of 8 pixels.

To be more robust against future changes, it might make sense to override the background color in the day mode CSS as well.

KTibow: > The quiet zone (padding) should be 32 pixels, though, as the standard mandates 4 modules of quiet zone. The QR code is generated with a scale factor of 4 which results in a module size of 8 pixels.

Which standard are you talking about? I was able to scan fine personally.
> To be more robust against future changes, it might make sense to override the background color in the day mode CSS as well.

Good point. However that isn't really "Night mode improvements"... Feel free to make a PR once this gets merged.

sur5r: > Which standard are you talking about?

The standard itself is non-free, but see https://en.wikipedia.org/wiki/QR_code#/media/File:QRCode-2-Structure.png or https://www.qrcode.com/en/howto/code.html for an explanation.

>  I was able to scan fine personally.

Yes, most scanners don't need that much quiet zone around the code, but I would not depend on that.

> Good point. However that isn't really "Night mode improvements"... Feel free to make a PR once this gets merged.

Yes, absolutely. That's out of scope for this PR. Was just a thing that came to mind.

public/css/grocy_night_mode.css
```diff
@@ -266,14 +289,32 @@ body.night-mode {
...
+	color: #d1d1d1;
 }
 
 .night-mode .text-muted {
 	color: #8f9ba5 !important;
 }
+
+.night-mode .qr-code {
+	background-color: #eee;
+	padding: 10px;
```
KTibow, line 318:
```suggestion
	padding: min(1.5vw, 20px);
```

public/css/grocy_night_mode.css
KTibow: I updated that. If you think it's good enough, then you can resolve the conversation.

sur5r: I have to admit I don't understand the purpose of the `1.5vw` value as that scales with viewport size while the QR code size itself is fixed. For usual screen sizes it's fine as the `20px` value takes over.

KTibow: It's *just in case* if you're using it on a phone, or tablet.

berrnd (approved):

PR #1371, opened 2021-02-21
Exclude dateless tasks from calendar API

Ape:


berrnd:
Why?

Tasks are not enforced to have a due date.
Currently this results for example that the generated .ical fiile will point those events always to today, excalty what you probably want to have for suche a "reminder task"...

Ape:
> Currently this results for example that the generated .ical file will point those events always to today

This is the reason. I don't want to flood my calendar with tasks that are in no hurry (no due date). I have many tasks that are like "would be nice to have someday".

berrnd:
> I don't want to flood my calendar with tasks that are in no hurry (no due date).

So this fits you, but anyone who used like I described it before (and it behaves as of now) will have a breaking change without an option to get the old behaviour - so if, then needs somehow options to get both at the end.

berrnd:
Closing this for now, feel free to reopen it if you want to continue on this - having a configurable way to also keep the current behaviour would IMHO be necessary to not enforce this for everyone (like already mentioned above).

Issue #221, opened 2019-04-29
Feature proposal: Printable barcode stickers


lukebo:
Hi Berrnd,
first of all, thankyou for this really great tool, which i use to keep track of everything inside our freezers.
Most of the contents are "non grocerystore-goods", frozen from our own garden, selfmade stuff and so on. Consequently, there is no barcode on them, just (at least sometimes) a handwritten note.
I would thus like to propose to implement a feature (e.g. within the product registration process) where one can define a certain (random) code and a best before date. Then, on can print all that on some sticker (e.g. with a labelprinter). Useful information on the sticker would be... ID-barcode, Name, Best before date and maybe the freezing date.
Keep up the good work!

berrnd:
Great idea, will be added... :)

lukebo:
Maybe this feature could also be benefitial for all the batteries (and maybe even the kitchen devices).

berrnd:
Sure, let's call this "grocycode", a unique ID for products/batteries/chores/equipment/(users) - everything which is selected by a dropdown can have a grocycode and then can be scanned as an alternative to be typed/selected from the dropdown (just like for products as of now).

Just not sure about printing with a label printer - just thinking of a page which displays all codes to be printed or an optional CSV export for now...

Would then also be possible to not only generate barcodes, but also those fancy QRCodes with an embedded logo like that: :D

![grafik](https://user-images.githubusercontent.com/2161815/56914614-d1eae300-6ab4-11e9-90f4-3cb598492e79.png)

shady2k:
When are you planning to add this feature?

berrnd:
I plan nothing about when to add what... ;D

If there are news about this, you will find it here.

shady2k:
Ok, thanks

mistressofjellyfish:
Hey berrnd,

I would like to spend some time implementing this!

However, I am not the biggest fan of QR codes, especially not when embedding logos. This requires the labels to be huge; and my use-case would be to have a simple cheap-ish (~60€) thermo label printer next to the fridge, which ideally should print a code, and some basic info onto a stick-on-label which I can throw on my tupperware.

Barcode readers that can read 2D codes like QR Codes can also read DataMatrix, which doesn't really care about alignment and uses less space. 2D readers are about 45€ used on ebay or about 90ish € new (according to geizhals). Or they can be read with a smartphone camera as is with 1D barcodes.

So my proposed implementation looks like this:

[...]

PR #1500, opened 2021-06-06
Grocycode, label printing

mistressofjellyfish:
As discussed in #221 – here is a draft implementation of grocycode.

php code has a class that abstracts de-/serialization; admittedly I hacked together the javascript involved.

ToDo:

- [X] php can generate grocycode
- [X] Product Picker can read product grocycodes (`grcy:p:<id>:<stockid>`)
- [X] encoded stock item gets selected automatically
- [x] Image generation
- [x] Show DataMatrix code in UI
- [x] Webhook infrastructure
- [x] Webhook-Print-Triggers for Products / Stockentries in Dropdowns
- [x] Purchase Print Trigger: select label per purchase, per purchased item or per item count. Have a smart default, overridable in the product config.
- [x] Have a Product-dependent default setting for label printing (scan mode stuff)
- [ ] Support reading from camera (needs a plugin for quagga2, but seems simple enough)


mistressofjellyfish:
So, now that the low hanging fruits are gone, I have to think more about how and where to actually display / print the labels.

Directly printing to label paper [like this one](https://www.avery-zweckform.com/blanko-etiketten/rechteckig-70x51-mm) is probably not a thing I will implement, as it causes UI headaches (where to select what to print?) and implementation headaches (to have precise measurements, pretty much the only option is to generate a PDF). 

Instead, I'll probably have an empty page that displays a label and only a label, and an option to download the code image only. These can then be fed into whatever label printing workflow floats your boat.

mistressofjellyfish:
Also, as I skimmed through #1273 I realized that direct printing is implemented there. So I wondered if a webhook really makes sense here, or should I just code the functionality directly into the grocy php?

My gut feeling says both printing jobs should be webhooks, but I value consistency over my gut feeling, so that's that. Input on this would be much appreciated.

berrnd:
Thanks a lot so far!

Just one general remark: I think this should be optional, at least by showing nothing (for example on the product card) when there is no grocycode.

IMHO this all makes sense for everything, but products, since that's the only entity which already has barcodes we commonly not even need to print ourselves...

mistressofjellyfish:
> IMHO this all makes sense for everything, but products, since that's the only entity which already has barcodes we commonly not even need to print ourselves...

My best counterexample would be selfmade jam glasses; or generally anything self made (like in the original issue). If no thermo label printer is around, pretty much the only option is to use those label papers and a regular inkjet (which my family did extensively). Printing 20 or 40 different grocycodes for all of the stock on these without having any form of automation seems cumbersome and error-prone to me – and something my dad probably wouldn't do. One picture without the stock reference though is manageable enough to put it on every label (and then have the default consume rule hit without stockentry tracking). So that's why I included Product grocycodes.

> I think this should be optional, at least by showing nothing (for example on the product card) when there is no grocycode.

As grocycode at the end of the day is just a kind-of-fancy reference to some database ID, every product and stockentry can be referenced by a grocycode without any further action. I agree that the product card could use some UI work. For now, I'll remove the embed there; it's accessible enough with the dropdown menus.

mistressofjellyfish:
As for the autogenerated labels, it currently layouts them like this:
image

Printed this equals roughly 70 by 20mm; datamatrix code is roughly 10.5mm by 10.5mm.

I guess we could abbreviate "Due Date" and its translation with "BB" and "MHD" or such? Because that label gets wide otherwise.

mistressofjellyfish:
I have settled on this: 
image

Printed, this equals to 35 by 15mm, scans fine. Unfortunately, on my screen the Barcode looks blurred when viewed in the browser. It's sharp when printed though. Does anybody have an inkjet printer at hand to test printing there? My b/w laser has no problem printing these fine details, but I feel this should be tested on an inkjet aswell.

mistressofjellyfish:
Wheee, my hardware has arrived. [My fork of brother_ql_web](https://github.com/mistressofjellyfish/brother_ql_web) now has an API endpoint that grocy calls as a webhook. This prints labels like this:

IMG_1898_s

Right now, this webhook is only run when the corresponding box on the purchase page is checked. Next on my list is adding hook running to a few other places.

mistressofjellyfish:
re: Webcam reading – I currently don't have a webcam at hand that scans any barcode reliably, yet alone a 2D code. I have a quagga plugin that uses zxing-js for DataMatrix reading, but I really don't have any idea whether it actually works (haven't tried on my phone because I lack a working dev HTTPS setup).

Both quagga2 and zxing (and any other library I came across for that matter) are extremely hard to develop for in the context of a non-modularized js environment like grocy currently has. I've spent the better part of the evening figuring out how to bundle/pack everything so that imports and plugin loading work.

I realize that the approach I took is probably one of the worser ways how to do this, right after using emscripten to compile some c library to webassembly (which unsurprisingly didn't work out).

Personally, I don't need webcam scanning. If it's a dealbraker for you I'll revisit it at the end and try in a proper nginx-hosted environment, which should make my phone happy.

mistressofjellyfish:
Okay, @berrnd – I think I am finished now. Please review my changes, feedback would be much appreciated.

mistressofjellyfish:
Re: Webcam scanning – I was able to test it now and it didn't work. I don't really have a good idea at the moment how to debug this. Barcode scanning still works, so I didn't break that, FWIW.

I'll have a look at this again when I find some spare time, but as I mentioned integrating the quagga plugin into the codebase is a fragile thing. My proposal would be to postpone that until I have made progress on the JS modernisation PR I'm planning. In the meantime, Grocycodes (without stock reference / additional data) could still reasonably be encoded as code128, which should be readable by quagga2.

berrnd (approved):
Wow 🤩, this looks really good, and just works - thank you very much! 👍👍👍

Just did some small adjustments, see https://github.com/grocy/grocy/pull/1500/commits/7a7031106396e9baf3bb4be449b7baa621b7710f.

controllers/StockApiController.php
```diff
@@ -626,7 +626,7 @@ public function ProductPrintLabel(\Psr\Http\Message\ServerRequestInterface $requ
 			(new WebhookRunner())->run(GROCY_LABEL_PRINTER_WEBHOOK, $webhookData, GROCY_LABEL_PRINTER_HOOK_JSON);
 		}
 
-		return $this->ApiResponse($response, $webhookData);
+		return $this->EmptyApiResponse($response);
```
mistressofjellyfish, line 629: Are you sure this change does not break local (in-browser) running of the webhook? Because the webhook data is needed by client-side javascript when printing from the click trigger (context menus of products / stock entries) in public/viewjs/stockoverview.js (or in the case below from public/viewjs/stockentries.js).

Still works when running server-side though, and won't be needed if I continue down the modern js path because product / stockentry data would be available anyways.

controllers/StockApiController.php
```diff
@@ -649,7 +649,7 @@ public function StockEntryPrintLabel(\Psr\Http\Message\ServerRequestInterface $r
...
```
mistressofjellyfish, line 652: Same as above.

mistressofjellyfish (commented):
Thanks for the quick review and merge! Just two little questions: as far as I'm concerned, the changes to EmptyAPIResponse are actually not possible; at least if the webhook should run client-side (triggered by jquery in browser javascript). The webhook data there is needed to be passed through to the javascript runner helper.

Obviously, there are two ways to solve this – either revert the change, or fix the javascript to use other APIs. The latter will be the case anyways once I start working on the js modernization, as the necessary data will be around there anyways.

controllers/StockApiController.php
berrnd: Sorry, just thought at first why there needs to be a response at that API endpoint but haven't tested the client side webhook running afterwards...

Reverted in https://github.com/grocy/grocy/commit/fe59fac1c374aec0357e85a20a6925868a2df2bc.

mistressofjellyfish: No worries! We catched it before release and that's all what matters. Thank you for the quick fix :)

ericblade:
@mistressofjellyfish do you have a repo with the code you're trying to use to make quagga2 read it?  Someone dropped this https://github.com/ericblade/quagga2/issues/389 in, and if I can assist, I'd love to

Issue #1501, opened 2021-06-06
Feature request: slowly move to vue.js for frontend javascript


mistressofjellyfish:
I do realize that this is far fetched and heavily opinionated, but my main goal is to start a discussion, *not* to get down to business.

As I'm working on #1500, I'm digging through code like [this](https://github.com/grocy/grocy/blob/master/public/viewjs/components/productcard.js) and its associated [view file](https://github.com/grocy/grocy/blob/master/views/components/productcard.blade.php), which does nothing but data binding, in an incredibly cumbersome, complicated and error-prone way.

As I like simplicity when hacking on stuff, I wondered if there is a better way. Having worked with vue.js in the past, it's data binding capabilities have always intrigued me.

Now, I realize the introduction of a whole javascript framework is a huge deal. I'm not married to vue, but I think anything that solves the databinding problem reduces the amount of code for UI problems.

Specifically, if you're open to try vue, this would by no means be a "do it now for everything" scenario. Instead, most of the code could be kept as-is while slowly introducing vue into the areas which benefit most (i.e. everything that is dynamically databound first).

I'll probably hack some prototype together once I finish up on #1500.

berrnd:
Has been already mentioned on many other places...

The main point is still simple: I'm unfortunately currently not into any JS frontend framework (still one of the things I want to do/learn, but time...) but at the same time the one who does and maintains ~99% of what you see in grocy (well, the last couple of months not really, but again, time...).

I would more start a complete rewrite of the frontend, without server side code (I think that's also what a JS frontend framework is for), the API should already provide everything needed.

mistressofjellyfish:
I see your point completely, and that's why I don't want to have this as a "omg we need x! now! do something!" issue but more of a discussion on what to do / centralize the discussion. I've seen people complaining in other issues, but that is precisely not what I want this to be about.

In the end, it's your call where to go. From my experience, especially in OSS, complete rewrites rarely happen (successfully). I would *much* rather prefer an incremental approach, and this would include moving all of the templates out of the server side. This would also enable us to build grocy as an electron app if needed or wanted.

Now; as far as I see it there two popular choices around, react and vue. Personally, my distaste for facebook things makes me use vue, but I'm fine either way. I don't want to put pressure on you in any way – I'm fine either way, and we can always continue that discussion once I figured out how this and the transition could work.

berrnd:
> electron app if needed or wanted

[...]

PR #1510, opened 2021-06-18
Modernize Javascript to es6; use Sass + Rollup + Gulp; Some page load speed improvements; Don't load modals in iFrames

mistressofjellyfish:
As I mentioned in #1501, I want to build a prototype introducing vuejs. This PR serves as a precursor, and whether or not vue gets merged, I'm positive grocy benefits from the work here. This reduces the amount of asset loads per page, even if the bundle size for each view isn't optimal yet.

In any case, **this PR breaks every custom javascript that relies upon the existence of a Grocy object in the page header**. This is done intentionally. `window.Grocy` is now an instance of GrocyClass, and gets passed a config object `window.GrocyConfig`, which currently lives in global space. If you want to modify that class, this needs to be done *after* GrocyClass is instanciated.

Also keep in mind that **most helper functions are not global anymore**; `extension.js` has many functions that aren't exposed in `GrocyClass.createSingleton`. This is by design. `import` what you need.

TODO:
- [x] clean up unused functions
- [x] clean up global space
- [x] clean up viewjs files, move declared functions to gorcy bundles where practical
- [x] eslint integration
- [x] fix errors reported by eslint

Other than that, this *absolute commit monster* does the following things: (I hope I did not forget anything)

- Introduce gulp to build javascript and css files.

  This includes moving node_modules out of the public/ folder.
  Use `gulp --tasks` to get a list of all tasks; however some
  of them are automatically generated.

  Use `gulp live` to watch for changes and automatically recompile
  what's needed.

- Upgrade to yarn2
- Upgrade FullCalendar to 4.4.2

  I know that 5.x is the current version, but two major version upgrades
  are too much right now. Also v5 would break any custom css as they
  renamed a bunch of classes.

- Move Styles to sass

  (Most) global styles are now included in one sass file. This also
  means that we now compile our own bootstrap.

- Javascript is now in strict mode

  Because everything is a module now, `use strict` is now in effect
  for all javascript files. There are probably still some parts left
  where implicit variable declarations were used.

- grocy*.js were split up in modules.

  `window.Grocy` is now an instance of GrocyClass. API-wise nothing
  has changed (albeit some functions were added regarding Undo actions)
  At the Moment, this leaks a whole bunch of functions into window
  (that was easier than tracking those down).

- FindObjectIn... style functions were removed.

  Array.prototype.find and Array.prototype.filter suffice.

- Use babel to preprocess javascript.
- Use rollup to bundle javascript.

  rollup bundles and tree-shakes es6 javascript bundles.
  It also allows to "import" css files and generate css
  files specific to this javascript file. This is used
  in viewjs scripts, for example when importing FullCalendar,
  to generate an associated viewcss file.

- Use postcss to post-process css files.

  postcss uses autoprefixer to handle browser compatiblity.
  Internally this uses the package `browserslist`; and is currently
  configured to the default setting.

- Minify assets when building in production

  `gulp publish` builds all assets in production mode, that is,
  the assets get minified. This includes javascript as well as
  css files.

- css bundling

  concatCss is used to pull @imports of non-sass-files into one
  grocy.css

- animate.css is now in the main bundle

  animate.css was used in so many places that it is now located
  in the main style bundle.
- deduplicate some functionality
  
  Some viewjs files declared the same functions in parallel.
  These were moved to the grocy bundle (`Undo*`) and are now
  accessible via `Grocy.Undo*`.


berrnd:
Thanks a lot for starting this.

Please bear with me when this most probably get's not merged too soon...
As also mentioned in #1501, when at some point in time I should continue to maintain all this, I would first need to learn / get familiar with the general concept and all the involved tools - things which I haven't used for cooking so far...

Still a thing I want to do/learn, but first (when I hopefully really come back to work on grocy soon) I will focus on fixing all the currently open bugs.

berrnd:
Beside custom JS (which definitely would require a major version bump), this will also affect:
- release ZIP generation
  - currently done via a simple (windows) batch file: [create_release_package.bat](https://github.com/grocy/grocy/blob/master/.devtools/create_release_package.bat)
- how the demo environment is updated
  - currently a simple `git pull && yarn install && composer install`, using the corresponding branch
- [grocy-desktop](https://github.com/grocy/grocy-desktop) (most probably)

mistressofjellyfish:
> Please bear with me when this most probably get's not merged too soon...

My intention was to keep this PR completely separate from all the vue stuff, so it can get merged regardless. Still, the tools used are (mostly) personal preference; and there are many people who prefer webpack over rollup. I initially tried to do everything with webpack but never got it to work with all the globals.

gulp probably isn't the best task runner in the world, it's there because I am somewhat familiar with it and the docs are reasonably easy to understand. Also, as long as the viewjs files are around, it makes it very easy for incremental hot-reload-recompile.

I completely understand your point about learning the tools though, and also I don't want to force my preferences on you if you're not comfortable with the tooling I used in this PR. Still I think that even without deep understanding of how rollup/gulp/postcss/sass work, this PR will make it easier to work on grocy and reduce page loading times significantly.

> Still a thing I want to do/learn, but first (when I hopefully really come back to work on grocy soon) I will focus on fixing all the currently open bugs.

I've added eslint integration to the ToDos of this PR. This should – in theory – make fixing bugs easier. Also, in my experience the point where all open bugs are closed never comes. Seeing as only about 35 issues are currently labeled as ui-bug, I think that investing time into better tooling that helps fixing bugs is best spent now. Most of the viewjs code (sans calendar) is absolutely untouched, so those bugs will still be the same and reproduce the same. Calendar-unrelated bugs introduced in this PR boil down to missing globals, but these should be catched by eslint so I can fix them prior to being merge ready.

re: release zip generation – this could be integrated into gulp.

re: demo environment – the new command should be `git pull && yarn install && npx gulp publish && composer install`.

re: grocy-desktop: AFAICT, grocy desktop uses a chromium rendering engine. Any recent-ish version (read: chromium 76 and higher) should work. If anything, the specific cromium version CefSharp runs on needs to be added to the browserslist config, so autoprefixer and babel pick them up. So, from the running perspective nothing should change. Obviously, when building the package, at some point a `gulp publish` needs to be run to compile all assets.

berrnd:
> in my experience the point where all open bugs are closed never comes

We had that point for grocy before every single release so far. 😇

And most bugs (also most of the current ones) were not about those a static code analyzer would find, more logic problems or "UI workflow glitches" - of course there can be tests for every kind of possible bug and maybe that's what the world calls "best practice", but this here is not the software your business relies on, just a fancy fridge organization tool and it was and is sometimes just so relaxing (for me) to just do it (and focus on the practical outcome, not on theoretical problems)...

But enough off-topic - I will definitely give this a try, will also need a lot of testing if everything still works.

mistressofjellyfish:
Adding eslint was easier than I expected. Once I'm done fixing errors I'm moderately positive that I didn't break anything.

However, in the process of fixing linter complains, I found some duplicate code in some places. I took the liberty to consolidate those; these particular files will get their own commit.

OmgImAlexis:
Is there actually a need gulp? If not I’d suggest removing it and just using npm scripts.

mistressofjellyfish:
Currently, yes; for a few reasons:

- The `live` task watches `js/**/*.js` and `scss/**/*.scss` and rebuilds them when they change. Because at the moment each view has its own js bundle, this means that gulp right now has 92 tasks, so that the view bundles can have their individual tasks.
- Some assets need to be copied to the output directory. To do this platform-agnostic, you need a file copy util from node anyways.
- It's parallel and uses my cpu efficiently.

That being said: I'm not married to gulp, but I like gulp.watch() and incremental rebuilds; Both of which I have no Idea how to do with package.json scripts.

OmgImAlexis:
> Currently, yes; for a few reasons:
> 
> * The `live` task watches `js/**/*.js` and `scss/**/*.scss` and rebuilds them when they change. Because at the moment each view has its own js bundle, this means that gulp right now has 92 tasks, so that the view bundles can have their individual tasks.
> * Some assets need to be copied to the output directory. To do this platform-agnostic, you need a file copy util from node anyways.
> * It's parallel and uses my cpu efficiently.
> 
> That being said: I'm not married to gulp, but I like gulp.watch() and incremental rebuilds; Both of which I have no Idea how to do with package.json scripts.

All of that could easily be done via npm scripts. I’d suggest checking out the concurrency package for running multiple at the same time.

copying files can be done via https://www.npmjs.com/package/cpy-cli

watching can be done via https://www.npmjs.com/package/npm-watch

OmgImAlexis:
If you need any help concerting from gulp I’d be more than happy to give assistance.

The main reason for not using gulp is speed and not needing contributors to learn yet another tool when basic npm scripts can be used.

mistressofjellyfish:
If you have code to contribute, feel free! Pull requests are enabled on my fork, and I'm happy to learn. However, any solution that rebuilds about 90 bundles when one file from viewjs changes is not a good solution, in my opinion.

Also, I don't think "contributors don't have to learn another tool" is a good argument to be made. With what you're proposing you're exchanging one powerful tool with two mini tools. I don't really see the difference regarding the learning curve, but my opinion probably doesn't count too much. I'm just a rando hacking away, its berrnds choice of tooling in the end, not mine.

Ideally, I'd move away from building 94 bundles altogether. But this requires significantly more work than what I feel should be in this already massive PR.

mistressofjellyfish:
Regarding speed, I did some experimenting with a Makefile that parallel-calls `npx rollup --input js/viewjs/<view.js>`. To build all ~80 viewjs bundles, gulp took about 33 seconds. The same task running with `make -j 4` (for my quad-core machine) took 1min48s, so nearly four times as long.

The explanation is also very simple: the viewjs bundles are incredibly small. Each takes only a few milliseconds to build, so spawning processes and parallelizing by throwing more processes at the problem actually results in a higher runtime, because it takes node far longer to start up than to actually do a thing. That's the same reason why `gulp-multi-process` makes the problem worse.

So, there's your reason to use a task runner like gulp. There is nothing to be gained parallelizing the current viewjs bundle generation.

However, I agree that if the viewjs bundles aren't used in that way anymore, using gulp should be reconsidered.

OmgImAlexis:
> . With what you're proposing you're exchanging one powerful tool with two mini tools. I don't really see the difference regarding the learning curve, but my opinion probably doesn't count too much. 

I'd suggest reading this https://www.freecodecamp.org/news/why-i-left-gulp-and-grunt-for-npm-scripts-3d6853dd22b8/
It's not a matter of replacing a "powerful tool" it's a matter of not overcomplicating the job. Grunt adds more issues to the plate. Who's going to keep plugins up to date, etc? Those tools I listed aren't going to change any time soon. They're stable and maintained by well known open source authors.

> Regarding speed, I did some experimenting with a Makefile that parallel-calls npx rollup --input js/viewjs/<view.js>. To build all ~80 viewjs bundles, gulp took about 33 seconds. The same task running with make -j 4 (for my quad-core machine) took 1min48s, so nearly four times as long.

33s vs 1m48s is a big difference but honestly who's actually going to be rebuilding them all? You said each file used a single viewjs right? So if you're only editing one watch should only recompile that single file. Gulp and npm-watch can both do this. 👍 

> So, there's your reason to use a task runner like gulp. There is nothing to be gained parallelizing the current viewjs bundle generation.

Adding gulp now will cement it in and it won't be swapped over. I'm saying this from experience with the 100s of open source projects I've helped on. This is why I mentioned it as soon as I saw this discussion start. Figured I'd make the suggestion before it was too late.

mistressofjellyfish:
> I'd suggest reading this https://www.freecodecamp.org/news/why-i-left-gulp-and-grunt-for-npm-scripts-3d6853dd22b8/

json sucks, especially for anything that should be maintained by humans. If you don't like a task runner, that's fine. I'll prefer a make file where I can actually have comments, but that is not going to run easily on windows. Granted, my gulpfile right now isn't well documented. But that's why this PR is marked as WIP.

> 33s vs 1m48s is a big difference but honestly who's actually going to be rebuilding them all?

I do, right now during development. If the filesystem is being watched, there's not much difference to using raw rollup.

> You said each file used a single viewjs right? So if you're only editing one watch should only recompile that single file. Gulp and npm-watch can both do this.

... sorry for the question, but did you actually look at and understand the code and how it's currently structured? Because  that's how gulp is configured here. But, to reiterate, *there is no static list of all viewjs files*. It gets pulled dynamically, the tasks are created dynamically, the watchers are created dynamically. Yes, this is possible with npm scripts, but requires to run-your-own build system in some javascript file.

As I said, if you have code to contribute, feel free. If you can come up with a solution that is prettier than mine, I have no problem with nuking gulp. Until then I won't further engage in this to-gulp-or-not-to-gulp discussion.

mistressofjellyfish:
3b342f9 also fixes #1511, #1512 and #1513. This a "side effect" of the function extraction which reduces the amount of copy-paste code.

mistressofjellyfish:
To give a bit more context to 7f59805 – Previously, grocy included about 60kb of json data on every single request for localization. Between releases this won't change, but it means that browser caches won't be hit in production mode.

To address this, GrocyClass now loads the static part of the language json in an ajax request. It ain't too pretty, because it involves firing a synchronous request which is deprecated; but returning a promise on translation because language data might not be available yet would have an even larger impact on the codebase, which is not ready for that (yet). In any practical terms, once the language file gets served from local cache there is not much of a delay anyways, so that's that.

To have an easily accessible test environment for these cache-hit or miss scenarios, I've included a Vagrantfile with a very minimal nginx setup.

mistressofjellyfish:
b7a1c87 shaves about 0.3 to 0.4s off of the DOMContentLoaded and about a second off of the Load events on the stockoverview page. Most of the remaining time that could probably be optimized is spent Initalizing sb-admin and the Datatables, both of which I have no intention of touching at the moment.

More time could be saved by further reducing the amount of individual javascript files that get loaded. There are still component script loads that spend half a second in the request queue waiting. The easiest solution would be to wrap all component init code in a function and bundle all files together, so that the individual viewjs file can register those components, so to speak. Let's see.

mistressofjellyfish:
I think I'm now at the point of diminishing returns and/or trading faster initialisation on one page for slower speed on others, so I guess I'll leave it at that and start a second PR based off of this for all the vue stuff.

@berrnd I know this PR escalated quite a bit and moved code around like no tomorrow. I tested the things I could with the demo data and while I can't say for sure that I didn't break things, there is nothing really obvious left.

To give an overview on what changed from a developers' perspective:

- node_modules is no longer in public. Instead, bundling is done for sass and javascript. See `gulpfile.babel.js`. At some point, I want to nuke gulp (and ideally switch from rollup to webpack for its superior documentation), but this is not possible in the current environment.
- These Paths have changed:
  - `public/js`-> `js/` ; keep in mind that the files in this directory have mostly been broken up into smaller modules, the distinction in helpers, config and lib is mostly arbitrary.
  - `public/viewjs/components`-> `js/components` ; These components are now functions and no longer inserted by the blade views but through `Grocy.Use("componentname");` in view javascript.
  - `public/viewjs/` -> `js/viewjs` ; many duplicate code blocks are now functions in the main Grocy class, which reduced code size a bit.
  - `public/css/*.css`-> `scss/` ; I did my best to split `grocy.css` logically. It's now bundled in one large-ish CSS file (450kb non-minified).
- The static part of the translation po files is now split off into build-time generated json files (`gulp makeLocales`), so that webserver caching can be achieved.
- version.json and its contents is now autogenerated during `gulp bundle`, which builds a release zip file and places it into `.release/grocy-version.zip`. version is calculated using `git describe --tags`, which takes the last named tag, and if commits have passed since appends `-numberOfCommits-gshortHash`. For the last commit of this PR, this would result in `v3.0.1-60-g1515f45`
- All javascript is in strict mode now. `eslint` will complain. But I only fixed four strict mode violations or so – mostly variables that haven't been declared before.
- `GrocyClass` sets its options explicitly and needs to be passed an options object (`GrocyConfig`)

Finally, this PR does not include any functional changes. While it seems like there are a lot more additions than deletions; the actual code that was added is not that much and mostly resolves around the build system. For some reason GitHub doesn't detect `git mv` commands if whitespace (indentation) has changed. If you have further questions, feel free to ask anytime.

berrnd:
Many thanks again for all the time you put into this.

As you said, this changes more or less everything, except the backend part.

Probably things are now better (in terms of how to do it "the modern way") and faster (except some pages IMHO speed in general is already ok on an average machine/smartphone, but faster is of course better), but those are all tools and concepts I'm at this point not familiar with (as mentioned before).

I need get into all this before merging it (already tried it this weekend, hasn't worked without digging further into it - some `yarn install` problems, probably related to my personal dev setup), maybe that's the final starting point for me to learn this, but on the other hand I would maybe have started with something else...

Currently when checking out this PR grocy kind of just feels not like my project anymore (code wise), please don't get me wrong, but for me that's also important and the base for fixing things fast / developing new things. Time is (obviously) already limited, when I now have to first check how general things are working in my own project this won't get better from that side. Working on grocy without understanding how it's working is not an option for me. I'm open to learn when time permits, but often fixing bugs or creating new features is just more fun...

So really sorry, I would hate to drop all this, but also can't promise to merge it at this point

First I will start (hopefully soon) fixing open issues before (trying) to dig into this, hopefully this keeps somehow mergeable.

That said, I also don't know if it's really useful to start the Vue rework now - maybe a separate project, a new frontend, using only the API, is the better idea...

mistressofjellyfish:
> Currently when checking out this PR grocy kind of just feels not like my project anymore (code wise), please don't get me wrong

I understand you, and to be completely honest: my intention never was to get this PR merged as-is at all. There were far too many fundamental changes and decisions made that I probably shouldn't have, the tooling used is entirely my personal preference, and so on. As I said in the first Issue I opened on this: I want this to be a prototype, which I maintain as long as I can. Take the parts you like, throw away those you don't. And if you dislike it as a whole, that's fine as well.

The only reason I split this off of the vue PR is that introducing vue will inevitably break even more stuff, introduce even more concepts, and will require even more learning (on my part too, because honestly: I currently don't have too much of an idea on how I'm going to do it). If I had done this changes all in one go while throwing vue on top of the current views, I wouldn't even have the slightest Idea how to explain this in a way that anybody could understand it. Right now, I'm positive that we can work things out to a point where you can make an informed decision whether you want to merge this at all and/or what needs to be done so you're comfortable with merging. If you have any questions regarding the code, feel free to ask anytime! I'm happy to explain my reasoning behind every change.

That notwithstanding, I still firmly believe that merging this PR will reduce maintenance burden, not only in the long run.

> hopefully this keeps somehow mergeable.

I'll (try to) keep this PR and the vue part with even more breaking changes in sync with master (that is, resolve any merge conflicts and fix bugs). For the forseeable future this seems doable, if no extremely big changes to the frontend happen, that can't get modernized by some thrown-together python scripts and a touch of manual labor.

> That said, I also don't know if it's really useful to start the Vue rework now

My main point on this is that I don't want to maintain a grocy fork. I can neither do any long-term commitment nor do I want to split a community that's not that large to begin with.

However, speed is a huge concern for me. My machine that hosts grocy is a quad core Atom running debian, and the frequent page reloads when entering products through the purchase page make it borderline unusable for me. I currently don't have the funds to upgrade this machine. What I do have are somewhat free evenings to address what bothers me most. And it very well may be that this doesn't align with what you are prioritizing, and that's okay.

I still think that publishing my changes is beneficial for everybody.

> maybe a separate project, a new frontend, using only the API, is the better idea...

A new frontend using only the API is what any vue rework will be at the end. I simply don't know when I will reach that point, and knowing myself I'll probably stop spending time on this once I have dealt with my own main pain points. In the meantime, I'll need to bring myself into a position where I can use the current views in vue. [0] This PR lays the groundwork for that, so I can make somewhat incremental changes without having a full-featured interface and another crippled one in fancy tech. Because fancy tech alone doesn't remind me that my food spoils or that I need to mop the bathroom. [1]

Now that this is out of the way, re: yarn –

> already tried it this weekend, hasn't worked without digging further into it - some yarn install problems, probably related to my personal dev setup

This is probably because of the yarn upgrade. If you have installed yarn globally, try `npm install -g yarn` to update it.

[0]: This also includes: not breaking classic mode in the vue build.
[1]: tbh, there are more things that bother me using grocy on a 10" 1280x800 touchscreen. Some of them are easily fixable, anything that involves opening a modal with an iframe in it isn't; not even with the work done in this PR.

mohe2015:
What about creating a separate PR with all changes except probably https://github.com/grocy/grocy/pull/1510/commits/f7bc6a3f6d276566104cae042f91865b7a165890 and https://github.com/grocy/grocy/pull/1510/commits/598860017b70e16066ee06fd66c5213469e16776? This would probably be some more work for you @mistressofjellyfish because some of these depend on the first commit but I could believe that some of that has pretty good changes of getting merged.
@berrnd  what do you think. I believe that getting some of the e.g. code-duplication-removal and eslint commits in could be beneficial and comparably small changes.

mistressofjellyfish:
Unfortunately most – if not all – of the rest of the commits depend on there being some sort of pre-processing done. So, sure, you can cherry-pick the deduplication and eslint, but I don't think that I'll main a separate PR right now.

I know that the benefit of using a module bundler isn't huge at the moment. Tree shaking doesn't work for instance, and there is not much I can do about that because of how much third party stuff lives in global space and isn't well modularized. I also haven't spent any time digging through the css code to see what actually needs to be there and what could reasonably be configured in `_variables.scss` (which again needs preprocessing).

Also; the more development dependencies accumulate, the larger the release zip will get if it includes the whole node_modules folder. 90% of its contents are unnecessary for a production environment. Moving node_modules out of `public/` would go a huge step, but also require to have some sort of build system at the end. And then we're basically back at f7bc6a3 which – for the most part – embraced the moving and it's consequences. es6 modules don't particularly like global space, so I had to wrap things in functions. Grocy was already a class, just not a well defined one. Properly defining it had the consequence that `Grocy.xxx` wasn't available in the header anymore and so `GrocyConfig` was born.


`postcss` is here because it's part of bootstraps build system (which also requires `autoprefixer`). I had to upgrade FullCalendar because you can't `import` FullCalendar3 (you also can't `require` it for that matter, because it won't find jquery).

So, as long as `node_modules` stays at `public/node_modules`, you can reasonably cherry-pick just the code deduplication. But I believe that node_modules should be moved out of public, especially if you want to integrate eslint aswell. eslint for example is 3.16MB unpacked and absolutely useless in a production environment. grocy-3.0.1.zip is 45MB compressed. A release produced by `gulp release` from this PR produces a zip file that is 17MB.

Add to that that yarn is deprecating node_modules altogether for a more efficient module storage, and is also deprecating yarn1. So node_modules will maybe not be around for too long anymore.

berrnd:
I gave it another shot but was again not successful, the tooling seems to rely on more things which seem to be not fulfilled "out of the box", maybe that was the idea for Vagrant - but just another big tool...

My main motivation to go on with all this at some point would be speed - @mistressofjellyfish could you maybe please provide a release ZIP built from this PR and I'll throw it on a demo instance - using the same environment, speed should be comparable.

<details>
<summary>What I tried to get this running</summary>

1. Checking this PR out

2. `composer install` => Success

3. Latest `yarn` is installed globally 
`yarn install` => Failed with
  ```
  Error: Cannot find module 'C:\Users\bb\Workspace\grocy_modernjs\.yarn\releases\yarn-berry.cjs'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:76:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
  ```
`yarn set version berry` seems to be required to use `yarn2` => Failed with the same error as above
Temporarily removed `.yarnrc.yml` (best googled guess) => `yarn set version berry` and `yarn install` succeeded (and the before mentioned missing file `.yarn\releases\yarn-berry.cjs` was created)

4. `npx gulp publish` => Failed with
  ```
  TypeError: cb is not a function
    at C:/Users/bb/Workspace/grocy_modernjs/gulpfile.babel.js:163:21
    at bound (domain.js:416:15)
    at runBound (domain.js:427:12)
    at asyncRunner (C:\Users\beb\WorkspaceP\grocy_modernjs\node_modules\async-done\index.js:55:18)
    at processTicksAndRejections (internal/process/task_queues.js:77:11)
  ```

5. Replaced [line 163 of `guplfile.babel.js`](https://github.com/grocy/grocy/blob/e030fda01a1876a1d9a8e5cb22e1b35b3c3fa217/gulpfile.babel.js#L163) with (best guess) `done => { done(); if (cb) { cb() }; })();`

6. `npx gulp publish` => Failed with multiple errors like
  ```
  './js/viewjs/recipeposform.js' errored after 9.36 s
  GulpUglifyError: Streaming not supported
  ```
</details>

mistressofjellyfish:
Aaaargh. Nah, these errors come from me being an idiot and too liberal with gitignore. I‘ll provide a fix and a Release zip shortly. Very sorry about that, I didn‘t test with a new clone but should have.

mistressofjellyfish:
[grocy-3.0.1-63-g151722ea.zip](https://github.com/grocy/grocy/files/6694009/grocy-3.0.1-63-g151722ea.zip)

I fixed and tested the build scripts. Sorry about that, I totally forgot to test with a fresh checkout. This is a minified build; Minifying is expensive. Build time on my machine was around 5 minutes for building, minifying and zipping. Without the uglify-js step, it's closer to 90 seconds, and could be further optimised with incremental builds.

berrnd:
> I fixed and tested the build scripts. Sorry about that, I totally forgot to test with a fresh checkout.

Thank you, absolutely no problem - `gulp publish/release/live` now also works here.

=> Here is a demo instance for this (does no auto updates): https://modernjs.demo-prerelease.grocy.info

=> Here is a demo instance of v3.0.1 without "demo additions" (stuff in the top toolbar and so on, provided using custom JS/CSS) to really be able to compare it (does also no auto updates): https://v3-0-1.demo-prerelease.grocy.info

To be honest, I don't feel a real speed difference so far when clicking around a bit, but it's measurable:

`/stockoverview` page before:
grafik

and now:
grafik

Interesting that the total size is now bigger, probably because there is now one big `vendor.js` file - not all third party JS files were included for every view before, but one bigger request is of course cheaper than a lot more smaller ones...

mohe2015:
"Add to shopping list" doesn't seem to preselect the item with the changes in this PR. (tested in my phone so maybe something else is broken)

mistressofjellyfish:
> Interesting that the total size is now bigger

Yes, the initial load size probably increased, but that is a tradeoff with initial loading speed on other pages. Once the cache is warm, the page size actually reduced; I measure ~25kB for modernjs and 32kB transfered with warm caches. That's not the end of the world, but it's something.

I did some testing on the consume page:
image

I didn't feel a huge difference and it was much less noticeable on subsequent reloads. Digging through the performance analysis the next biggest culprits that cost time are the initialization of sb-admin and the datatables.net stuff.

The first will become a non-issue when moving to a Single-Page-Application model because it will get initialized once and that's it (or, more probably, nuked in favor of a virtual DOM solution with vue).

Datatables.net causes multiple style recalculations and forced reflows, which stall execution. There is not much we can do about that besides replacing the component as a whole, which may has other unintended side effects. Thats one thing I want to explore with the addition of vue, but ~270ms can easily be saved by just not initializing sb-admin on every request.

> To be honest, I don't feel a real speed difference so far

It'll probably be more noticeable on slower machines like my laptop. It's too hot right now to collect any meaningful data (my cpu temps are through the roof, absolute times are all over the place), but I'm seeing the same .3 to .5s improvement as you. Because my baseline was higher, what maybe doesn't feel like a huge improvement to you is significant on my machine. stockoverview DOMContentLoaded is closer to 1s on this PR than to 1.5 on 3.0.1, and I do feel this difference.

@mohe2015 I'm pretty sure that's something I broke. I moved some DOM blocking viewjs inclusions to async load after the page is loaded, and this included `viewjs/products.js` (and most probably `viewjs/shoppinglist.js`). For some reason, they don't load right now, so let's see where I shot myself in the foot.

I also just noticed that I broke the userpicker, which I'll fix shortly as well.

mistressofjellyfish:
Okay, both issues are actually the same and boil down to how dialogs and their embedded iFrames work. My can-I-do-this-change-without-breaking-stuff-script didn't catch this case, so instead of moving I removed some viewjs inclusions where it actually wasn't possible.

My gut response would be to nuke the iFrames, which has other benefits like not screwing with toastr or the viewport (and reducing AJAX size) but I need to educate myself on the consequences. One thing that will absolutely be required is that the viewjs files and their initialization code needs to be scoped to the loaded element, which will require some jquery trickery and wrapping them into functions.

Both things are needed for the vue PR anyways, but introduce another fundamental change on how grocys frontend works.

mohe2015:
I would really recommend to split this up into multiple PRs from likely-to-get-merged to not-so-likely-to-get-merged. I wouldn't feel confident reviewing this either. Also then you could do some releases in-between to be sure that nothing major broke.

mistressofjellyfish:
Well, the logical splitting point would be b7a1c87 which is the last commit before the "problematic", deep changes started. ESLint without moving `node_modules`, the bundler and modularization is pointless IMHO. And even these changes alone break enough to warrant a major version change.

We can split the PR there without too much work, and probably clean up the rest in separate PRs. But honestly, just extracting the code deduplication misses the point of this PR, and I'd rather do strict QA (which, admittedly, I haven't done to a point where I can safely say that everything works). I don't see much benefit in releasing in between merges, as most [0] of the commits that follow b7a1c87 also break major things and would require even more major version releases, while doing absolutely nothing for the end users. The load time reduction of this PR is only possible because small amounts of time were saved across many places.

re: vitejs and snowpack – as far as I understand those tools, they are doing what gulp is currently doing. I like snowpacks approach, but coming from the experience that is this PR, this would require significant work on top of what is already done here.

[0]: Specifically: 5988600, and whatever I come up with to address the iFrame issue.

mohe2015:
I'm currently reviewing the code a bit. Please split up mass-renames into a separate commit next time as git is sometimes stupid and doesn't detect this. Also the split up of grocy.js would've been easier to review if the split up was in a separate commit than the formatting / etc changes although I can understand that splitting that into seperate commits is hard.

mistressofjellyfish:
> Please split up mass-renames into a separate commit next time as git is sometimes stupid and doesn't detect this.

Hmmm, I was actually hoping that using `git mv` would solve that issue. It didn't, sorry if that is causing you extra work. I will try to remember this in the future.

mohe2015:
> > Please split up mass-renames into a separate commit next time as git is sometimes stupid and doesn't detect this.
> 
> Hmmm, I was actually hoping that using `git mv` would solve that issue. It didn't, sorry if that is causing you extra work. I will try to remember this in the future.

No problem. I agree that this is counter intuitive.

mohe2015:
Oh no Github's mark as viewed feature is broken for per-commit reviews. aaargh

js/grocy.js
```diff
@@ -150,21 +152,21 @@ class GrocyClass
 
 	translate(text, ...placeholderValues)
 	{
-		if (this.Mode === "dev")
+		/*if (this.Mode === "dev")
```
mohe2015, line 155: intentional?

mistressofjellyfish: Semi. I was toying around with some Ideas how to extract translations statically using https://github.com/sinedied/gettext-extract, but reverted the changes. This seems to have fallen through.

js/grocy.js
```diff
@@ -212,6 +214,24 @@ class GrocyClass
 		this.IdleTime += 1;
 	}
 
+	Use(componentName)
```
mohe2015, line 217: Would a similar functionality be possible with `import` ing the components. Because that could also run some code on import (although this is discouraged) but then the methods itself wouldn't be on a global but instead would be accessed from the import. And what would you think about that?

mohe2015:
I'm through the first round of code-only review. 
> "Add to shopping list" doesn't seem to preselect the item with the changes in this PR. (tested in my phone so maybe something else is broken)

I didn't do a great job because I didn't find that bug but I know why because I really wasn't *that* thorough.

I'm also totally unfamiliar with the source code of this project - this is basically the first time I read it.

The amount of removed code-duplication is impressive.

I also think splitting up the huge grocy.js is really needed.

> Currently when checking out this PR grocy kind of just feels not like my project anymore (code wise), please don't get me wrong, but for me that's also important and the base for fixing things fast / developing new things.

I can understand that. I think some abstractions like the FrontendHelpers (e.g. https://github.com/grocy/grocy/pull/1510/files#diff-e54175062fde75a24a8e845c53a1200158654865b8a08ef33e66d0988ac80b72R14) can be unfamiliar at first but I think that they are really beneficial as they reduce code by a lot and make it easier to change the behaviour / fix bugs at all places at once. I think they also make the code easier to understand although they create a little higher learning-curve.

@mistressofjellyfish If you only plan to add more commits but don't force push I would try to split up the renaming and test whether this makes it easier to review. You can also try that of course (interactive git rebase if you don't know it but learning it probably takes some time).

js/grocy.js
mistressofjellyfish: In principle, yes, that would be possible. However, this means that importing is done statically, can't be guarded by ifs and that breaks with the userfields component which – depending on the defined fields – dynamically decides if it needs to initialize the datetimepicker component.

Some components, like the product card, already expect to be only once in the DOM, which is violated by example opening the product card from the stock overview page, opening the stock entries from the card, and opening another product card. What jquery does in that instance is pretty much undefined, but [it seems to pick up the first instance only](https://jsfiddle.net/1d2tLwzb/). I can't tell if it's problematic, but scoping solves the issue.

However the scope needs to be passed dynamically, and this can only be done when opening the modal. Scoping isn't implemented in the components yet, but I'll do the same trick I'm currently throwing at the viewjs/ folder by defining `$scope = $(container-selector).find` for scoped components / views and `$scope = $` otherwise. Unfortunately, this requires me to manually check each jquery usage in every view file, because the object returned from `$(selector)` and `$(selector).find(foobar)` aren't exactly the same type.

mistressofjellyfish:
> @mistressofjellyfish If you only plan to add more commits but don't force push

I won't force push, so feel free. I understand the basics of interactive rebase, but I'm unfamiliar with most things but squashing.

mohe2015:
Current state: https://github.com/mohe2015/grocy/tree/modernjs-rebased4 (need to fix commit messages and authorship though) (and a small merge fix)

mistressofjellyfish:
> I didn't do a great job because I didn't find that bug but I know why because I really wasn't that thorough.

I'll try to give more context. In master, views/stockoverview.blade.php loaded the viewjs for the purchase page (viewjs/purchase.js). purchase.js has (besides others) a dependency on the datetimepicker component. All direct usages of the datetimepicker component via `Grocy.Components.DateTimePicker` are safeguarded such that they only run when the component is detected on the page. This works, because only if the component is in the php blade view the component javascript is run.

5988600 moves the components to the main grocy bundle and makes the view javascript now responsible for `Grocy.Use`ing the components they need. This breaks purchase.js if its included from the stockoverview page because the datetimepicker component fails to initialize properly if it doesn't find its element(s) in the DOM - which aren't there yet. Because I and my scripts could not find any apparent use of code within purchase.js on the stockoverview page, I simply removed the inclusion of purchase.js on the stockoverview page. But this can't be done: if the view is loaded through the submenu and an iframe in a bootbox, it doesn't automatically re-run purchase.js.

There are basically two ways to address this:

1. fix the safeguards in the views, re-include the javascripts. This has the drawback of costing time on loading and also doesn't handle that some components expect them to be only once in the DOM. All other problems with the iFrame-Loading (js/configs/globalstate.js:187ff) like wrongly placed toastr notifications and broken nested overlays persist.
2. fix the views so that they can be scoped to another element and dynamically loaded. Then let all views that do open modals with embedded views preload the viewjs after DOMContentLoaded and initialize it (and its components) scoped to the container. Because everyting is scoped now, we can replace `Grocy` with a proxy class that respects the slightly different configuration required by some embedded controls while avoiding loading and layouting an entirely new page.

Im currently implementing method 2, because that's what would be necessary before trying to use vue anyways.

mistressofjellyfish:
Okay, views and components can now be scoped to a specific element in the DOM, which the userfieldsform compoent uses to fix #1520.

These commits also fully broke all modals, but lays the ground work to embed views properly. Now I just have to figure out how to handle bootbox correctly, and implement a json embedded mode which returns only part of a DOM.

Each scoped view gets a scoped GrocyProxy. This is from the public API the same as GrocyClass, but overrides some methods and properties to properly scope them. FrontendHelpers and `Grocy.Use` is automatically scoped as well without the view noticing.

Because the components are now prober classes, I could easily extract a base class of some of them and removed about 700 duplicate LoC.

mistressofjellyfish:
It's done! Dialogs are now loaded with AJAX instead of an iFrame. The improvement in UI speed is massive, but see for yourself. Should I generate a new release zip?

Still have to fix some minor layouting issues though.

berrnd:
The https://modernjs.demo-prerelease.grocy.info demo is updated.

From what I understand by having a quick look, you just pipe the view (only the `@section('content')` as far as I can see) into JSON and return that, then using that for the dialog - that can't IMHO work alone, since the page's JS still seems to rely on load events and such...

All I can see are empty dialogs (and some other things I noticed):
https://user-images.githubusercontent.com/2161815/123324827-91be0880-d537-11eb-8241-56654ea2d808.mp4
Much to test...

Thanks a lot again, but turns more and more into a whole different/separate frontend project technology/concept/tool-preferences wise...

> the log_missing_translation API calls stall everything here

Pretty much irrelevant, since that only runs when `MODE == 'dev'` (is only there to auto-add new localization strings to `strings.pot`) - and personally doesn't have that while not using a CPU beast (old i5-8250U / HP ProBook 450 G5) on my couch:
https://user-images.githubusercontent.com/2161815/123324801-88cd3700-d537-11eb-9f13-44331841cef8.mp4

mistressofjellyfish:
> Much to test...

I'm at it. The particular issue from the first video is fixed already, I just haven't pushed the commit yet / decided if the fix is complete. The version I currently have hicks up with the dropdown menus in datatables, and the old fix doesn't work on iframeless modals.

This is a shortcoming of bootstraps dropdown javascript, which leaves no possibility to detach the dropdown div from the button it's displayed on. This is fine for most scenarios, but all of the datatable dropdown headaches are solved by placing the dropdown div DOM-wise outside of the table, which I'm currently investigating how to do.

> From what I understand by having a quick look, you just pipe the view (only the @section('content') as far as I can see) into JSON and return that, then using that for the dialog - that can't IMHO work alone, since the page's JS still seems to rely on load events and such...

It's more than that. Yes, on the php side of things I just shove everything into a json and return that. The magic happens in `GrocyProxy` and the `GrocyClass.OpenSubView` which makes each scope think it's on it's own page. GrocyProxy even has a virtual URI, to support views loading stuff from GetUriParam.

Technically speaking, each view gets a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) that intercepts accesses to GrocyProxy, and if they are not defined on the proxy, forwards them to the root GrocyClass instance.

The main work I did was on scoping everything and making sure that everything initializes properly. Aside from surprises out of state initialized in former big grocy.js, I am very positive that I catched 99% of it, after some debugging.

> Thanks a lot again, but turns more and more into a whole different/separate frontend project technology/concept/tool-preferences wise...

I understand the difference in tooling and the fundamentally different approaches with precompilation / bundling vs. writing plain javascript.

However, I don't see big technology / concept changes right now. Most of my changes either pay back technical debt by consolidating duplicate code or fix actual bugs, like the whole iFrame stuff, which screws with the viewport, toastr, the combobox, ... The bugs I report while doing this are mostly there because the frontend JS is full of duplicate code and technical debt (like the datetimepicker stuff). Don't get me wrong – you can always bolt-on something, but adding more techincal debt to a frontend that's fragile already doesn't really make changes easy or without somewhat easily understandable consequences. And that's where I am now, because everything is so intertwined, tightly coupled and scattered all over the place, that you have to make the same edit in 10 places. Let's try to reduce these places to one or two.

It's inevitable that these changes have side effects, because some assumptions the code had when it was written aren't there anymore. I try my best to catch what I can before pushing / posting, but it's inevitable that something (major) slips through – partly because I don't know all ways to do X in grocy, partly because these changes go so deep that I cannot test every button between commits. Once I'm confident again that I finished my major work, I will do this.

Or, in other Words:
There is still a Grocy instance around. It's just not global anymore.
There are still the same components with the same public API around. They are just requested by view javascript, and not the php views.
Fundamentally, the views are the same code. Just some reoccuring parts have been extracted to methods.
Everything still uses jquery. It's just sometimes a jquery that's scoped to a parent element.

> Pretty much irrelevant,

I work on an i5-3450M, so that's that. I think the real problem why these requests stall is the `php -S` development server. I know that these requests only run on dev mode, but that's the mode I'm currently in and sometimes they stall everything. Can't reproduce it well enough, though. Maybe for now an optimization like only pushing strings to the server that aren't in the translation dict yet does the trick aswell. I haven't thought it out yet.

mistressofjellyfish:
@mohe2015 Since you're editing the history of this PR, I just noticed I screwed up the merge commit. Could you by chance squashe fa320c4 and b9493f0 while you're at it?

mohe2015:
> @mohe2015 Since you're editing the history of this PR, I just noticed I screwed up the merge commit. Could you by chance squashe [fa320c4](https://github.com/grocy/grocy/commit/fa320c448d24c47e59a3c1fb83f98859e6a73086) and [b9493f0](https://github.com/grocy/grocy/commit/b9493f0e0be758dca449b153edb15fd7495eb3cd) while you're at it?

I don't think this helps you as I can't push to this branch. But I think you should be able to just do that yourself and won't break my rebase. If you need help - feel free to ask.

I'm not at that commit yet but I also think that I'm not rebasing more of it as I think I got the most important parts and also don't have that much time.

mohe2015:
You probably need `--rebase-merges` for the rebase command

mohe2015:
https://github.com/mohe2015/grocy/tree/modernjs-rebase-merge-fix should hopefully be correct

mohe2015:
It's not!

mohe2015:
Add there are still some files with merge commits e.g. views/quantityunitconversionform.blade.php it seems

mistressofjellyfish:
I think I got it. Let's hope I didn't destroy your rebasing efforts.

mistressofjellyfish:
Of course I missed the same file. Oh well.

mistressofjellyfish:
re: localisation callback - I also just noticed that these API calls add dynamic strings to the po file (and that’s why I have to reset it prior to commit, and fix it with headaches if I do add new strings). Static extraction would solve that issue as well.

mistressofjellyfish:
I could finally reproduce the empty datatable issue, and I have fixed it now for good. The problem was stockentries.js leaking a custom filter function to global space, which of course screwed with the other datatables.

It gets properly unloaded now when the modal closes.

mistressofjellyfish:
05bbb26 adds two things:
(1) the mentioned check if the key is already in the language json,
(2) It defers all Calls to `FrontendHelper.SaveUserSetting` until after the page has finished initial rendering. This can be safely done because `SaveUserSetting` doesn't return anything and doesn't change local state of the user settings. This further improves page speed as more request slots are available for actual data-getting-API calls.

mistressofjellyfish:
Modal nesting is now somewhat fixed.

v3.0.1:
image

modernjs:
image

Still not ideal – I think modals shouldn't be nested at all (and I removed the possiblity to open a chain of the same modals where I catch it), but that would be a behavior change I'm not comfortable with right now.

mistressofjellyfish:
Okay, I think I am now at a point where I can reasonably say that I tested most of Grocys functionality.

That includes every button on every non-modal page, the first layer of modals (what happens when modals get stacked is anyones guess; but all in all state should not leak), Purchase/Consume/Transfer/Inventory, Scan Mode, Editing / Creating everything in the master data, Undo of Purchases, Conusmes, Transfers, Consumptions of specific stockentries; Task creating, executing; battery editing, load cycling; equipment editing, creating; calendar, mealplan, shoppinglists; general settings (everything from the dropdown); usersettings; nightmode; wakelock; header clock.

Additionally, this PR fixes all bugs I encountered along the way and could reproduce on the demo system: #1511 #1512 #1513 #1522 #1523 #1524 

I guess we can once again compare speed. The AJAX modals are very fast now; next thing on my list is to get rid of most page reloads, but that is (a) not happening without vue (and even there only gradually) and (b) not happening in this PR.

Also, not all changes in a subview bubble up and change stuff in all open views. This is especially noticeable on the stockoverview page when manipulating stock through the stockentries modal. But fixing this properly includes home-running some sort of data binding anyways, so it's probably better to just get vue in, and do binding and truth manipulation properly.

berrnd:
The https://modernjs.demo-prerelease.grocy.info/ demo is updated once again.

And, no doubt, speed is of course better, measurable and also noticeable (for me personally not so much on the devices I'm using).

However, as much as I appreciate your efforts, I have to say that I can't imagine that I'm getting comfortable with merging this in the foreseeable future.

It's definitely more a kind of complete rewrite of everything, which is of course ok.

I would suggest maintaining this in your own seperate project if you have fun with it, propose it to the community, if it really works, is faster, get new features over time, I'm sure a lot will be happy with it and consider switching, myself included.

mistressofjellyfish:
> I would suggest maintaining this in your own separate project if you have fun with it, propose it to the community, if it really works, is faster, get new features over time, I'm sure a lot will be happy with it and consider switching, myself included.

Well, if you want this to go this way, then that's fine by me. I won't add more commits to this branch then, which obviously means that it will probably get unresolved merge conflicts very fast.

> It's definitely more a kind of complete rewrite of everything, which is of course ok.

I think we have very different understandings of *complete*, *rewrite* and *everything*. But okay, I'll accept that we won't get on the same page and continue in my fork.

mohe2015:
> > I would suggest maintaining this in your own separate project if you have fun with it, propose it to the community, if it really works, is faster, get new features over time, I'm sure a lot will be happy with it and consider switching, myself included.
> 
> Well, if you want this to go this way, then that's fine by me. I won't add more commits to this branch then, which obviously means that it will probably get unresolved merge conflicts very fast.

If it's fine for @berrnd I think this PR could be kept open (maybe as draft but it shouldn't matter) to create some visibility that there are major efforts to refactor to Vue.

> > It's definitely more a kind of complete rewrite of everything, which is of course ok.
> 
> I think we have very different understandings of _complete_, _rewrite_ and _everything_. But okay, I'll accept that we won't get on the same page and continue in my fork.

I can understand @berrnd's point of view. Modern frontend frameworks like React, Vue, Angular, Svelte, ... all have a learning curve (that in my opinion is really worth the effort) and it's (almost) always harder to understand somebody else's code than your own. Line of code wise there are lots of changes and lots of moving around things so I can understand that this feels a bit overwhelming and also if you don't have the needed time to follow up with the changes this is bad. But I also agree that there is big technical debt (duplicate code, no proper build process, etc.) so a refactor is needed/beneficial in my opinion.

@mistressofjellyfish I would be fine with doing a code-review at some point (but probably not the diff then but the new code instead) and also some testing. Unfortunately I don't have the time to code myself on this project.

mistressofjellyfish:
Re-Reading my last comment, I think it came across a little bit more aggressive than it was intended. Sorry for that. It's just... from my point of view, we aren't even at a point where I began adding anything like a framework. What started with a basic house-keeping task to make the project fit for actually tackling such a thing turned into this monster, because the frontend code as it is currently in master **is just plain bad**. It's a mess, and nothing short of what I did here would fix that. All speed improvements and bug fixes are side-effects of the housekeeping.

I know that reviewing such big changes, *especially* when they come from a person which just turned up and has ample free time and is stuck in a hyperfocus so works 10 hours a day straight for a week (this is exaggrated, but the hyperfocus part is true) is a huge task. And I don't expected this to be merged fast plainly because of this reason.

But since I repeatedly got the feedback that this is basically a rewrite (which it isn't, the rewrite hasn't even started yet. It's nearly the same HTML, and for the most part the same javascript albeit in some different places) and I should move that to it's own project, that's what I'm now doing. No one has any obligation to like my code, my tooling preferences or my architectural choices. I will however sit down with any interested person and explain every single changed line if needed. Even if at times it doesn't seem like it, I'm not in the "change for the sake of change" business, and I understand the complexity of these things and how daunting it can seem. Nontheless, even if berrnd hasn't decided if he want's to merge this at some point, I basically feel being told to go away. Which I'm doing, even if that is precisely not what I wanted. I'm somewhat good at contributing code, but mediocre at best when it comes to community stuff.

I'm not entitled to anything. All I ask is that my work is looked at before being reasoned about. And if the time to look at my work isn't available right now, that's fine. I can wait, but I won't if I'm being told to fork.

For anyone that's interested, my updates go to [mistressofjellyfish/not-grocy](https://github.com/mistressofjellyfish/not-grocy). I don't want this to be a hostile fork, but I will not maintain cross-compatibility, neither in the front- nor in the backend. I hope that at some point the projects will be merged, though.

@mohe2015 Since I'm now free to just move along with moving things to vue, I'll add some automated testing when I get to it – even if I hate writing tests. This project is already complex enough to warrant the overhead, and spending 12+ hours straight on verifying all functionality between commits is something I absolutely don't want to do again. Before doing a release, I want to fully move not-grocy to a SPA, even if vue will mostly just use the old views. I think I'm far enough with the refactoring that the diff to 7512b66 won't be as huge anymore, but I will definitely need to rip out a few more things.

I'll probably announce it in a few places once I got the rest of the big performance hogs sorted and brought the rest of my issues in order. not-grocy also doesn't care about Docker or the desktop application at the moment, and I haven't decided yet what I want to do with the latter. Me personally not using windows when I can avoid it begs for some electron shenanigans, but that's a tech I'm absolutely unfamiliar with and frankly, a browser does the job just fine at the moment.

berrnd:
I don't think that kind of discussion brings out anything useful, so probably my last words for now:

I have of course looked over this, at least a couple of hours the last week, trying to get familiar with it. And even that some (!) JS code is exactly the same before, everything is restructured (to your likings, maybe I would have done it differently), the base-tooling stack used for everything related frontend and the build process is different - so we obviously have different understandings of the word "rewrite", but that's it for me.

So all I wanted so say is by far not that you should go away, you can even take over grocy if it makes fun - that's what I meant with the suggestion to move this into an own project. I have no problem with anything and be more sorry to drop such a big effort you brought into grocy. But I also have some experience regarding this or Open Source projects in general and it always turned out so far that I'm the one (for grocy as of now) who needs (or wants, since I'm not forced to do it) to maintain this, regardless of promises before. And that's, definitely, the more harder the more is changed and the more new tools/concepts are introduced.

It seem also to doesn't fit since my focus for grocy is the output, means I really don't care about good code quality, all what counts for me is that it works (at okayable speed). A lot call this "plain bad" (or that this project as a whole is just crap, heard everything in the last years, sometimes this dropped my motivation, however, nobody is forced to use it or look at its code). The maybe first thing related to that I heard was "how can you use such a crap database like SQLite, please use a real database" (or not using tests, not using a CI pipepline, not having documentation, not auto-generating the OpenAPI specification, having Windows specific build-scripts) - well...
For such a plain bad project, it works IMHO pretty well, obviously not only for myself...

This changes everything (for me), would need so much time to get into it, but the output is just the same (while it's a little faster or a little more on slower devices). That's just not my focus for grocy and all for me at the end to not going this way at the moment. Even if it has duplicated code and whatever, it was (and is) managable (for me).

If not-grocy becomes so much faster and will be actively maintained (kind of reminds me of the big announcement of [HouseHold/HouseHold](https://github.com/HouseHold/HouseHold), which is apparently still at this stage), I will consider switching, which is then maybe the end here at all. I'm personally using a lot of other things which I partly customize (like [berrnd/jellyfin-web](https://github.com/berrnd/jellyfin-web)) and where it's ok for me that I don't understand the tooling used fully, but that just not fits for a project I (should or want to) maintain as a whole.

> If it's fine for @berrnd I think this PR could be kept open

Of course this here can stay open for a very long time.

And again thanks a lot for all on this, despite that this as a whole doesn't fit, maybe parts are worth considering. For example dialog nesting or somehow combining JS files (while not using a big build process for it) to reduce web requests drastically and I maybe will adapt them in the future.

berrnd:
Closing this now as out of scope for this project.

> > If it's fine for @berrnd I think this PR could be kept open
>
> Of course this here can stay open for a very long time.

The primary goals for this and not-grocy seem to be so vastly different, that I don't see any benefit to keep this open anymore.

> If not-grocy becomes so much faster and will be actively maintained, I will consider switching, which is then maybe the end here at all.

That's still true for me - see you there, in a faster, harder, stronger fridge management world, where everything will be just so much better than in this ugly and just plain bad messy world here.

PR #1521, opened 2021-06-24
Migrated bootstrap3 col-xs-* to bootstrap4 col-* classes

mistressofjellyfish:
`col-xs-*` classes were renamed in bootstrap 4 to `col-*`.

The breakpoint widths changed too, but I don't remember the particularities right now, and this process also can't be automated.

To make these changes, I ran this command:
`find . -name '*.php' -exec sed -i '' s/col-xs-/col-/g {} +`

diffs look sane.


berrnd (approved):
👍, thanks - but can't really notice any differences or any issues on the state before...

mistressofjellyfish:
I came across that because nothing fit the iFrameless modals. Don't know the full impact on master, but most probably some margins appear on mobile devices where they were previously missing.

PR #1540, opened 2021-07-20
Extend REQUIRED_PHP_EXTENSIONS check about core extensions

tetragir:
After installing Grocy on FreeBSD, even with all extensions installed that are listed in REQUIRED_PHP_EXTENSIONS, Grocy still couldn't start. The added 3 PHP Extensions are also needed to run Grocy.

### DIFF
--- a/helpers/PrerequisiteChecker.php
+++ b/helpers/PrerequisiteChecker.php
@@ -4,7 +4,7 @@ class ERequirementNotMet extends Exception
 {
 }
 
-const REQUIRED_PHP_EXTENSIONS = ['fileinfo', 'pdo_sqlite', 'gd', 'ctype', 'json', 'intl', 'zlib'];
+const REQUIRED_PHP_EXTENSIONS = ['fileinfo', 'pdo_sqlite', 'gd', 'ctype', 'json', 'intl', 'zlib', 'filter', 'iconv', 'tokenizer'];
 const REQUIRED_SQLITE_VERSION = '3.9.0';
 
 class PrerequisiteChecker
--- a/helpers/PrerequisiteChecker.php
+++ b/helpers/PrerequisiteChecker.php
@@ -4,7 +4,11 @@ class ERequirementNotMet extends Exception
 {
 }
 
-const REQUIRED_PHP_EXTENSIONS = ['fileinfo', 'pdo_sqlite', 'gd', 'ctype', 'json', 'intl', 'zlib', 'filter', 'iconv', 'tokenizer'];
+const REQUIRED_PHP_EXTENSIONS = ['fileinfo', 'pdo_sqlite', 'gd', 'ctype', 'json', 'intl', 'zlib',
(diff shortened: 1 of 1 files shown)


berrnd:
All the mentioned extension are [core extensions](https://www.php.net/manual/en/extensions.membership.php#extensions.membership.core) which are enabled by default ([ref](https://www.php.net/manual/en/tokenizer.installation.php), [ref](https://www.php.net/manual/en/iconv.installation.php), [ref](https://www.php.net/manual/en/filter.installation.php), also true for [`json`](https://www.php.net/manual/en/json.installation.php), but only since PHP 8), so I guess you use a custom distribution (or maybe that's a thing of FreeBSD)?

Potentially there are more of those core extensions used in grocy (directly or indirectly by dependencies; dependencies most of the time don't list them as required)...

tetragir:
I'm not sure about Linux, but on FreeBSD there are no PHP Extensions installed together with PHP by default, all of them has to be installed explicitly. So far these 3 are the only ones that were needed in addition to the ones that are already on the list.

berrnd:
> on FreeBSD there are no PHP Extensions installed together with PHP by default, all of them has to be installed explicitly.

Also the case for Linux (Debian) and Windows, but not for the core extensions, since they are not extensions - [PHP states](https://www.php.net/manual/en/extensions.membership.php#extensions.membership.core):

> _These are not actual extensions. They are part of the PHP core and cannot be left out of a PHP binary with compilation options._

  &nbsp;
  
> So far these 3 are the only ones that were needed in addition to the ones that are already on the list.

Kind of strange then, since it can't be about those 3 core extensions alone - when I look at the others, grocy definitely uses functions of kind of all...

tetragir:
So far only these 3 seem to be missing, if I see any other issues that are caused by missing PHP Extensions on FreeBSD, I would be happy to report them.

berrnd:
> So far only these 3 seem to be missing

And that's the thing which is weird.

Just picking one random other: [`Password Hashing`](https://www.php.net/manual/en/book.password.php), the by that extension provided function [`password_hash`](https://www.php.net/manual/en/function.password-hash.php) is for example used [here](https://github.com/grocy/grocy/blob/f47ca963ab4f4f49bdadc1d4b66dd7ebb1868795/middleware/DefaultAuthMiddleware.php#L46). So why is that extension not missing in the same way?

I still think this is a broken/custom distribution, PHP's definition of core extensions is that they are always enabled (in fact, they can't be disabled unless you use custom compile flags as also quoted above, or if FreeBSD does this in this way, it's still not compliant).

So if, then please make this complete by checking for any usage of any core extension and include all "needed" as required.

berrnd (changes requested):
Like already mentioned above:

> So if, then please make this complete by checking for any usage of any core extension and include all "needed" as required.

tetragir:
Hi,
Grocy is in use since a couple weeks now and I didn’t encounter any other issues, so it seems only these were missing. 👍

PR #1553, opened 2021-08-01
Make default best before time configurable (closes #1552)

### LINKED ISSUE #1552: Feature request: Make the default best before date configurable
It would be nice to configure the default best before time shown on the "New Product" page. E.g. you could set the default to 1 year or to "never expires" depending what's most common and adjust it only when needed.

dkruzic:


berrnd (changes requested):
Thanks, but then please also add an UI option for that like for any other product preset: https://github.com/grocy/grocy/issues/1552#issuecomment-890764202

And please mind the naming, we call it "due date" generically do distinguish between expiration and best before dates.

berrnd:
This is now finalized in https://github.com/grocy/grocy/commit/10bd5ce9007b60352b69602c897f12e41bf1147b.

PR #1557, opened 2021-08-03
Fixed the bug where grocy would return "Not a grocycode" all the time

larsverp:
When making a request to `/api/stock/products/by-barcode/{barcode}` using a "normal" barcode, Grocy would return a "Not a grocycode" Exception. 

This is really annoying when using the API to search for products using their barcode. 

Turned out the Try-Catch wasn't build correctly and the Exception wasn't catched but just returned as an Exception. Also changed some small variable mistakes in the function I worked on.

### DIFF
--- a/helpers/Grocycode.php
+++ b/helpers/Grocycode.php
@@ -77,7 +77,7 @@ class Grocycode
 			$gc = new self($code);
 			return true;
 		}
-		catch (Exception $e)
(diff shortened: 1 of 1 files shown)


robinvw1:
You would think a normal code editor would catch typo's like this. 🤔

berrnd (approved):
Thank you! :)

PR #1559, opened 2021-08-04
Hide elements when printing

MrKrisKrisu:


berrnd (changes requested):
Thank you, I think the intention was to optimize the meal plan page to be printed in general - but there are far more elements (probably all buttons), which are not really useful when printing the meal plan, than this particular two or three...

So if, then I think we should make this complete.

MrKrisKrisu:
You're right. I don't really know the Grocy code yet and haven't gotten my head around it, so I've only partially hidden it. I will have a closer look at the code later and then hide the other buttons also on the printouts.

MrKrisKrisu:
Sorry for the inconvenience. Now all buttons are invisible! :)


image

berrnd (approved):
👍, thanks!

PR #1565, opened 2021-08-06
Fix exception handling

MrKrisKrisu:
I have tried to use the API endpoint ```/stock/products/by-barcode/{barcode}/add```. But I always got the error ```{"error_message":"Not a grocycode"}```. Even if I use a normal barcode (not a grocycode).

The try catch was not caught correctly because an import is faulty.

### DIFF
--- a/helpers/Grocycode.php
+++ b/helpers/Grocycode.php
@@ -77,7 +77,7 @@ class Grocycode
 			$gc = new self($code);
 			return true;
 		}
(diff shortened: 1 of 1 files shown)


berrnd:
Thanks, but that was not the only problem related to `Not a grocycode` and is already fixed in #1557 - please always reference pull requests to the current `master` branch.

PR #1566, opened 2021-08-06
Change PHP Dependency from >= to ^

MrKrisKrisu:
In composer.json it is better not to work with `>=` in PHP. When the next PHP version is released (with breaking changes) there will be problems. `^` is more suitable, because you only allow minor patches here.

I also added a name and a description so the json is valid.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,9 @@
 {
+	"name": "grocy/grocy",
+	"description": "ERP beyond your fridge - grocy is a web-based self-hosted groceries & household management solution for your home",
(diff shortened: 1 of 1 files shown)


berrnd (changes requested):
It's ok for me about the PHP version constraint change.

The other fields are only required for published packages, grocy is not a published package (`composer validate --no-check-publish` returns no errors).
https://getcomposer.org/doc/04-schema.md

In general all more theoretical problems I don't really care about...

berrnd (approved):
👍

PR #1567, opened 2021-08-06
Add missing urldecode

MrKrisKrisu:
I wanted to use an API endpoint that requires {barcode}. My barcodes contain special characters. For example, spaces.

Example barcode: `CHEESE 123` (with space)

In the URL this is passed as `CHEESE+123` or `CHEESE%20123` due to encoding which cannot be found in the database.

The code has not yet decoded the input from the URI. This is not normally necessary with "normal" barcodes. I have added this in any case. 

This should actually have no further effect on other barcodes. 
I have tested GrocyCode -> works

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -161,7 +161,7 @@ class StockApiController extends BaseApiController
 	{
 		try
 		{
-			$args['productId'] = $this->getStockService()->GetProductIdFromBarcode($args['barcode']);
+			$args['productId'] = $this->getStockService()->GetProductIdFromBarcode(urldecode($args['barcode']));
 			return $this->AddProduct($request, $response, $args);
 		}
 		catch (\Exception $ex)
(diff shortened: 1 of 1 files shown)


berrnd:
The used framework ([Slim](https://www.slimframework.com/)) takes care of that, otherwise any access to `$args` in any controller would need that handling, not only the "by-barcode"-routes...

Also cannot reproduce that - demo test instance with exactly that (`CHEESE 123`, [product "Cheese"](https://test-e0w2ozm5l8gfd460hepzrq.demo-prerelease.grocy.info/product/10)) barcode:
=> [`GET /stock/products/by-barcode/CHEESE%20123` ](https://test-e0w2ozm5l8gfd460hepzrq.demo-prerelease.grocy.info/api/stock/products/by-barcode/CHEESE%20123) works

(`+` is not a URL encoded space, that's only allowed in `application/x-www-form-urlencoded` encoded form data or also often used in URL query parameters, IMHO still not valid there, [ref](https://stackoverflow.com/questions/1634271)).

Issue #1061, opened 2020-10-17
Workflow for purchase, group purchase transactions


white-gecko:
I just went shopping, and

- bought the things on the list (using grocy-android) and some extra stuff

When I came home 
- my wife took the phone to scan everything (grocy-android app) while I put it in the locations
- during that we add all of the new products (grocy-android app) using the batch mode
- afterwards I went through the batch list to add the missing unit sizes and expiry dates (grocy-android app)

- now I wanted to use the grocy web interface to add the prices of the purchase but I can't find a way to edit the transactions
- also I wanted to check if all of the added quantities were tracked correctly, as I have just scanned one bottle for each case.

(The only option I see at the moment to get the prices in and check the quantities is to do an inventory, which does not seem right.)

Am I doing something wrong in my purchase workflow?

My idea would be to group all of the transactions for one shopping event so I can go thorough it after scanning the products and add the price information. Also this would allow to compare the total of the purchase and maybe later attach a photo/scan of the receipt to the purchase event.

Sorry, this is a report of my first experiences mixes with a question and a not so simple feature request.

berrnd:
> (The only option I see at the moment to get the prices in and check the quantities is to do an inventory, which does not seem right.)

You can also edit each existing stock entry, without inventory, on [`/stockentries`](https://demo.grocy.info/stockentries) (button at the top of the stock overview page).
If you order that table by purchased date descending, you would have the latest bought products at the top.

Without the app (which has it's [own repo](https://github.com/patzly/grocy-android) and also is not maintained by me), the easiest way would be using the button "Add all items to stock" on the shopping list page, which just goes through all items on the shopping list / opens the purchase page while all information (amount, default best before date and so on) is already prefilled.

Stock transactions which belong to the same "action" (so for example when using the "Add all items to stock"-button) are already grouped together by an id, we currently do nothing with it (maybe would be something for the stock entries page), don't know if the app also handles this "correctly".

I personally do the scanning at home, put everything on the table and use a wireless laser barcode scanner while just using the normal purchase page, afterwards I put the things where they belong to. But I personally don't care about price tracking...

white-gecko:
Thank you very much.

For the issue the feature request to **group transactions as a purchase** is left over. Maybe this can be implemented using the id that you mention, but I'm not sure if this would do the trick already because a purchase would can be split in multiple transactions if I scan part of the products with my phone and my wife uses her phone to scan the other half.

[...]

PR #1579, opened 2021-08-15
Make length of stock id configurable.

lowlee:
Stock id values are unique to their product, so the same value for two different products is fine. The values are generated using a strong hash, so for most instances 6 or 8 digits is enough to make the likelihood of collisions acceptably tiny, a value like 16 means we could generate unique entries for every tin of tomatoes ever made.

Why make it shorter? So you can print shorter grocycode barcodes. But probably not shorter than 6.

### DIFF
--- a/config-dist.php
+++ b/config-dist.php
@@ -101,6 +101,19 @@ Setting('DEFAULT_PERMISSIONS', ['ADMIN']);
 // 1D (=> Code128) or 2D (=> DataMatrix)
 Setting('GROCYCODE_TYPE', '1D');
 
+// Length for stock id tags. Stock id values are unique
+// to their product, so the same value for two different
+// products is fine. The values are generated using
+// a strong hash, so for most instances 6 or 8 digits
+// is enough to make the likelyhood of collisions
+// acceptably tiny, a value like 16 means we could
+// generate unique entries for every tin of tomatoes
+// ever made.
+//
+// Why make it shorter? So you can print shorter
+// grocycode barcodes. But probably not shorter than 6.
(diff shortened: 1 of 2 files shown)


berrnd:
> Stock id values are unique to their product

That's wrong. The lookup "first by product id" only happens for grocycodes, in fact, the product part could be omitted for stock entry grocycodes to make it even shorter. [Here](https://github.com/grocy/grocy/issues/1061#issuecomment-716139435) is a short explanation about the ids used for stock entry relationship.

So I define a `STOCK_ID_LENGTH` of `1` or `2` and this still guarantees no collisions? I doubt it...

So would at least more checks related to that. I'm definitely not a fan of making own "uniqueness decisions", however, if it's about shortness, then the most short id would be most probably an own counter.

And this doesn't matter so much for 2D barcodes, `DataMatrix` is the prefered grocycode variant...

berrnd:
And I would say this doesn't need to be configurable, it just need to work, causes only confusion about what a stock_id is for and it's definitely not really required to dive into that topic for just using grocy...

lowlee:
Fair enough - This was an attempt to solve a problem we had with our setup that we thought might be useful for others stuck using a 1D barcode (due to only have a 1D scanner)

Also this was one change on the road to getting stockentry grocycodes to work

berrnd (changes requested):
So as mentioned above:

- I don't think this need to be a config option, the shortest reasonable id is fine (so maybe an own counter)
  - Existing stock entry ids should be migrated

berrnd:
Looks outdated, so closing this - can be reopened anytime you like to continue on that.

Issue #1580, opened 2021-08-15
GetProductIdFromBarcode does not enforce the grocycode being of the correct type


lowlee:
GetProductIdFromBarcode does not enforce the grocycode being of the correct type. Battery and Chore grocycodes if supplied to the consume-stock-bybarcode endpoint would consume a product if there was a matching row in the database

PR #1581, opened 2021-08-15
fix: confirm grocycode is of PRODUCT type in GetProductIdFromBarcode

lowlee:
Enforce only PRODUCT grocycodes in GetProductIdFromBarcode (fixes: #1580 )


berrnd (approved):
👍, thanks!

PR #1584, opened 2021-08-16
fix: make stockentry grocycodes consume the actual stock entry not the product in general

lowlee:
if the stock_entry_id is in the request body use this instead of the stockentry grocycode
this may not be the correct way to interpret this but one of them has to win

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -282,6 +282,10 @@ class StockApiController extends BaseApiController
 			{
 				$specificStockEntryId = $requestBody['stock_entry_id'];
 			}
+            else if (array_key_exists('stock_id', $args) && !empty($args['stock_id']))
+            {
+				$specificStockEntryId = $args['stock_id'];
+            }
 
 			$locationId = null;
 			if (array_key_exists('location_id', $requestBody) && !empty($requestBody['location_id']) && is_numeric($requestBody['location_id']))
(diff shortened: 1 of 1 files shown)


berrnd:
Thanks - haven't looked into it yet, is definitely not a problem for the frontend currently since the "specific stock id" is there just set after scanning a grocycode ([ref](https://github.com/grocy/grocy/blob/8e552f11464460449f025998ec11b327db423fc9/public/viewjs/consume.js#L235)).

controllers/StockApiController.php
```diff
@@ -322,7 +326,23 @@ public function ConsumeProductByBarcode(\Psr\Http\Message\ServerRequestInterface
...
+				if ($gc->GetType() != Grocycode::PRODUCT)
+				{
+					throw new \Exception('Invalid grocycode type');
+				}
+				$args['productId'] = $gc->GetId();
+				if ($gc->GetExtraData())
+				{
+					$args['stock_id'] = $gc->GetExtraData()[0];
+				}
+			}
```
berrnd, line 341: Exactly that check is already present (you added it) in `StockService->GetProductIdFromBarcode`, why not reuse it?

controllers/StockApiController.php
```diff
@@ -322,7 +326,23 @@ public function ConsumeProductByBarcode(\Psr\Http\Message\ServerRequestInterface
 	{
 		try
 		{
-			$args['productId'] = $this->getStockService()->GetProductIdFromBarcode($args['barcode']);
+			if (Grocycode::Validate($barcode))
```
berrnd, line 329: Sorry, but have you tested this at least once? `$barcode` is undefined here...

controllers/StockApiController.php
```diff
@@ -282,6 +282,10 @@ public function ConsumeProduct(\Psr\Http\Message\ServerRequestInterface $request
 			{
 				$specificStockEntryId = $requestBody['stock_entry_id'];
 			}
+			elseif (array_key_exists('stock_id', $args) && !empty($args['stock_id']))
+			{
+				$specificStockEntryId = $args['stock_id'];
+			}
```
berrnd, line 288: I don't think we need this 2-step handling (simple is better), if a grocycode was supplied, its stock entry id (if present) should be used and take precedence.

berrnd (changes requested):

controllers/StockApiController.php
lowlee: Sorry this was a push from the wrong repo locally

controllers/StockApiController.php
lowlee: Reordered the precedence in the new commit - grocycode stock_id's now go first

controllers/StockApiController.php
lowlee: The fix in `StockService` is for all the endpoints that get the product id from the barcode - the `ConsumeProductByBarcode` endpoint needs additional processing to get the stock entry id from the barcode.

While it could be changed to still hit the `StockService` the gain is minimal as we still need to check for a valid product grocycode to extract the ExtraData.

PR #1587, opened 2021-08-18
Fix link to access user settings from the dropdown menu not beeing re…

daufinsyd:
…cognized as link by the browsers

Hi, I can't click on the user settings link (using either firefox or falkon) because of the link not being recognized as one (because of the leading data-).

Since others are plain href, I guess it's a typo, so here's a fix.

### DIFF
--- a/views/layout/default.blade.php
+++ b/views/layout/default.blade.php
@@ -639,7 +639,7 @@
 							href="{{ $U('/batteriessettings') }}"><i class="fas fa-battery-half"></i>&nbsp;{{ $__t('Batteries settings') }}</a>
 						@endif
 						<div class="dropdown-divider"></div>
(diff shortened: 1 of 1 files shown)


berrnd (commented):
> Hi, I can't click on the user settings link (using either firefox or falkon) because of the link not being recognized as one (because of the leading data-).

Can't reproduced that.

It's converted into a valid link (since the whole link should be dynamic) on page load, code ref:
https://github.com/grocy/grocy/blob/61a58ddef0378c0493ead07c03279151d73cd3e5/public/js/grocy.js#L855-L867

daufinsyd:
Indeed, I figured out last year I already installed grocy for testing purpose and didn't remove the volumes. After deleting the viewcache folder as stated in the doc, everything is working fine.
Sorry for the merge-request ^^

PR #1600, opened 2021-09-04
Add a setting to treat -1 as singular

b1n4ryj4n:
To ensure that all languages get always their right singular form
or plural form its better to let the user choice whether '-1' should
be interpreted as singular form.

### DIFF
--- a/config-dist.php
+++ b/config-dist.php
@@ -22,6 +22,9 @@ Setting('MODE', 'production');
 // one of the other available localization folders in the "/localization" directory
 Setting('DEFAULT_LOCALE', 'en');
 
+// For most languages it might be a good idea to treat -1 as singular
+Setting('TREAT_MINUS_ONE_AS_SINGULAR', true);
+
 // This is used to define the first day of a week for calendar views in the frontend,
 // leave empty to use the locale default
 // Needs to be a number where Sunday = 0, Monday = 1 and so forth
--- a/services/LocalizationService.php
+++ b/services/LocalizationService.php
@@ -73,7 +73,7 @@ class LocalizationService
 	{
 		$this->CheckAndAddMissingTranslationToPot($singularForm);
 
-		return sprintf($this->Translator->ngettext($singularForm, $pluralForm, $number), $number);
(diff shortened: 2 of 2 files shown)


berrnd:
Thanks, but each language already has its `Plural-Forms` definition ([default GNU gettext handling](https://www.gnu.org/software/gettext/manual/html_node/Translating-plural-forms.html)), so we really don't need a separate config option for that - more config options, more confusion...

If you noticed that a definition for any language is wrong, feel free to fix that (localizations are managed via Transifex as [mentioned in README](https://github.com/grocy/grocy#localization)).

Some examples:

https://github.com/grocy/grocy/blob/14bb04d285a3c04508f4ba8b837a1de805fd2c34/localization/ru/strings.po#L24

https://github.com/grocy/grocy/blob/14bb04d285a3c04508f4ba8b837a1de805fd2c34/localization/cs/strings.po#L25

https://github.com/grocy/grocy/blob/14bb04d285a3c04508f4ba8b837a1de805fd2c34/localization/de/strings.po#L20

https://github.com/grocy/grocy/blob/14bb04d285a3c04508f4ba8b837a1de805fd2c34/localization/nl/strings.po#L42

b1n4ryj4n:
Sry but im not familar with Transifex and im not sure where i could edit this.
My problem is here:
here
In my native language german the journal shows the wrong form. If i consume just one product it should displayed as singular form and not as plural form.

berrnd:
Transifex verwaltet das erstmal automatisch, es gibt eine "offizielle" eingebaute Liste ([ref](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html)).

Nach ein bisschen recherchieren: [PHPs `ngettext`](https://www.php.net/manual/en/function.ngettext.php) ist das Problem (siehe die Kommentare dort, `$count` [müsste ein unsigned Integer sein](https://www.gnu.org/software/gettext/manual/gettext.html#index-ngettext).)

Wie auch immer, dafür brauchen wir, wie schon erwähnt, keine Option, Sprachen haben doch Regeln, fertig.

Wird ein #1601 weiter getrackt.

PR #1617, opened 2021-09-17
StockLogEntry Property "used_date" is Format date instead of date-time

sebiecker:
The column "used_date" in "stock_log" is created as Type DATE.

In the grocy.openapi.json File the Property "used_date" of StockLogEntry is mentioned as Type date-time


berrnd (approved):
👍

PR #1628, opened 2021-09-28
Display ingredients and recipe side by side

adamschmalhofer:
In full screen mode display both the ingredients and the recipe side by
side instead of having to switch between them via the tabs if there is
enough space.

### DIFF
--- a/public/css/grocy.css
+++ b/public/css/grocy.css
@@ -654,11 +654,12 @@ canvas.drawingBuffer {
     }
 
     .fullscreen #ingredients-0 {
-        min-width: 20rem;
-        max-width: 20rem;
-    }
-    .fullscreen .nav-link[href="#ingredients-0"] {
         width: 20rem;
+        padding-right: 3rem;
+    }
+
+    .fullscreen .grocy-tabs {
+        display: none;
     }
 
     .fullscreen .tab-content {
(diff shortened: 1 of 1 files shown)


berrnd (changes requested):
Thanks, but doesn't completely fit I would say...

- The tabs have then no functionality and should be hidden

- A little spacing or border or something is needed between ingredients/preparation since currently the text is nearly overlapping:
&nbsp;
grafik

(Or maybe/also inline and left align calories/costs.)

adamschmalhofer:
Thanks. I made the changes. All my preparations use numbering which automatically adds  some spacing, so I didn't notice. I had repurposed the tabs as table headings, but just leaving them away is good too.

berrnd (changes requested):
Thanks again, but I still think that doesn't fit completely...

- Isn't the ingredients column a little bit too small?
  &nbsp;
  Standard 1920x1080 resolution:
grafik
  &nbsp;
  1366x786 resolution (common on smaller notebooks):
grafik
&nbsp;
- When having a "multi recipe" (included recipes, [demo ref](https://demo-prerelease.grocy.info/recipes?recipe=6)) it also doesn't really fit (1080p resolution):
grafik

berrnd:
Looks outdated, so closing this - can be reopened anytime you like to continue on that.

PR #1647, opened 2021-10-10
Fix printing label per unit

joshvazquez:
If a product allows label printing per unit, the current code prevents you from setting "stock entry label" to "label per unit" on the purchase page because the option is disabled. Only a single label can be printed.

Even if you set the product default to "label per unit" and don't change it on the purchase page, the selection is ignored because the option is disabled, so `jsonForm.print_stock_label` is undefined. In this case, no labels are printed at all because the value is undefined.

This change enables/disables the option in the correct cases.

### DIFF
--- a/public/viewjs/purchase.js
+++ b/public/viewjs/purchase.js
@@ -313,11 +313,11 @@ if (Grocy.Components.ProductPicker !== undefined)
 							{
 								$("#default_print_stock_label").val("0");
 							}
-							$('#label-option-per-unit').prop("disabled", true);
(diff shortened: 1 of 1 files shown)


berrnd:
> So to simplify that I would suggest:
> - Remove the product option `Allow label printing per unit` (and the corresponding en-/disabling of dropdown menu items)
> - Rename the product option `Stock entry label` to something like `Default stock entry label on purchase`
>   - Just use that as a default on the purchase page

That is now done by https://github.com/grocy/grocy/commit/fc413a05d1ef24d6edb74f056b2eeac7376d2b82.

PR #1664, opened 2021-11-05
dark bg was missing in dropdown, text was dark on dark bg in product …

corbolais:
…ellipsis

Signed-off-by: corbolais <corbolais@gmail.com>


berrnd (approved):
👍

Issue #1160, opened 2020-12-07
When adding a product that's default location is set to a freezer, it pulls the default expiry time of a non freezer location


maglite4cell:
I purchase alot of meat to freeze and have it default location as a freezer location.  When I add it to my inventory, it tells me that it will expire in a week, instead of what I would expect it to be of what I set it as the frozen location.  Is this by design?  What is the best way to handle converting multiple items to the Freezer?

berrnd:
> Is this by design?

Currently yes, as also the hint next to the option indicates - only on moving, so when using the [/transfer page](https://demo.grocy.info/transfer):

> **On moving** this product to a freezer location (so when freezing it), the best before date will be replaced by today + this amount of days

Of course we can (optionally) also apply this always, let's track that in this issue.

B1ue-W01f:
IMO, this should be the default. Im not sure how it compares for everyone else, but 99% of items I put in the freezer go from shop to freezer. Arbitrarily adding them to the fridge and then transferring them is inefficient for me.

Purchase -> Freezer means frozen from purchase.

PR #1672, opened 2021-11-07
StockService: Set the product's bestBeforeDate to the freezer date if it is being purchased to a freezer

grahamc:
Not yet tested, but I'm about to try it!

### DIFF
--- a/public/viewjs/purchase.js
+++ b/public/viewjs/purchase.js
@@ -286,34 +286,7 @@ if (Grocy.Components.ProductPicker !== undefined)
 						$("#tare-weight-handling-info").addClass("d-none");
 					}
 
-					if (Grocy.FeatureFlags.GROCY_FEATURE_FLAG_STOCK_BEST_BEFORE_DATE_TRACKING)
-					{
-						var dueDays;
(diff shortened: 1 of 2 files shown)


berrnd:
Thanks, related to / requested in #1160.

But if, then please also prefill that on the /purchase page (like the `Default due days`) to make this more comprehensible and still be able to change it if needed.

grahamc:
I just pushed a fixup making the "Purchase" flow's date picker default to the freezer's best before date, too. Thanks for the suggestion! I've also checked that this works now.

berrnd (changes requested):
Thanks again, noticed this while testing the changes:
On purchase, the due date is only prefilled when the products default location is a freezer, not when that's not a freezer, but I just put it this time directly into the freezer (so prefilling also needs to take place when changing the location)

grahamc:
That makes sense, I don't know how to implement that, though. Can you offer any pointers or suggestions? I don't do a lot of web development, and the stuff I did here took a lot of experimenting and looking around to figure out. I'm happy to fix this issue but I think I'd need some help.

berrnd:
No problem, thanks again - that's now finalized by the two commits referenced above.

berrnd (approved):

PR #1686, opened 2021-11-15
Don't print labels when asked not to

anwright:
v3.1.3 prints a label no matter what, even if label printing is set to "No label.

### DIFF
--- a/services/StockService.php
+++ b/services/StockService.php
@@ -268,7 +268,7 @@ class StockService extends BaseService
 				]);
 				$stockRow->save();
 
-				if (GROCY_FEATURE_FLAG_LABEL_PRINTER && GROCY_LABEL_PRINTER_RUN_SERVER)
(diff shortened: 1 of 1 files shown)


anwright:
Note that this change is untested, because I don't have a dev environment set up, and I'm running grocy-docker, which has a read-only filesystem.

berrnd:
Thanks for finding that, but that was the wrong check - `$stockLabelType == 2` is already handled in the `if` statement above.

Fixed in https://github.com/grocy/grocy/commit/b83e4f53b1a878f3619b1d1bd8ab64dc2803a6bb.

anwright:
I think your commit and mine are functionally equivalent, though your version is probably very slightly faster if the label printing feature is enabled, but "no label" is selected for the current operation.  Either way, thanks for applying the fix.

berrnd:
When I've checked that this morning I saw you've added `$stockLabelType == 2` (which is already checked above), but you did exactly the same I did in https://github.com/grocy/grocy/commit/b83e4f53b1a878f3619b1d1bd8ab64dc2803a6bb.

So sorry for not merging this, somehow overlooked that or it just was too early... 😅

Issue #1685, opened 2021-11-13
Bug: Timeago (fuzzy time display) is incorrect in Hungarian


Tallyrald:
This is not a Grocy localization issue, rather a JQuery plugin error which I don't think ever will be fixed on their side. Let me explain in a bit more detail.

Whenever a time is displayed like 'X days from now', the same string is incorrect in Hungarian (probably in other languages too). It says 'X napja' which means 'X days ago' - the exact opposite of the expected meaning. This only occurs for future dates like upcoming expiration dates.

The error comes from [this jQuery plugin](http://timeago.yarp.com/). Unfortunately the software is not really maintained anymore and the bigger issue is actually the language and how it works. You see in Hungarian you can't use fixed suffixes / prefixes to indicate past or future. Hungarian uses appendixes which is not supported by the plugin and so future dates remain in the default past tense.

I believe using Momentjs.From() and Momentjs.FromNow() would solve the problem. Their implementation uses the correct appendixes and wording. I belive MomentJS is already heavily used in the project so it wouldn't even be a new dependency.

To recap the difference (and the problem):
Currently using Timeago: Expiration is EN: 'X days from now' - HU: 'X napja' (X days ago)
Using MomentJS: Expiration is EN: 'in X days' - HU: 'X nap múlva'

berrnd:
Thanks for the explanation, but not a grocy bug, so the wrong place here.

There are fixed suffixes possible in jquery-timeago, so most probably a simple missing translation problem - that's the localization file used: https://github.com/rmm5t/jquery-timeago/blob/master/locales/jquery.timeago.hu.js

When the project is not so actively maintained - fork it, create a pull request to change the dependency of grocy to your fork, done.

Of course using Moment.js for that would be also an option, but honestly - if that only affects one language, I don't really care that much for now...

Tallyrald:
Okay, I understand, that's totally fair.

The prefix/suffix translation doesn't apply to Hungarian so even if that translation file was better implemented it would still be incorrect simply because there is no structure in the language that supports that kind use of telling the time difference.

Just out of curiosity: if I were to get familiar with the grocy codebase and try to implement momentjs strings instead of timeago... would you consider accepting it?

berrnd:
> Just out of curiosity: if I were to get familiar with the grocy codebase and try to implement momentjs strings instead of timeago... would you consider accepting it via a PR?

Of course.

There is also some special handling added currently which would then need to be adjusted accordingly, for example:
[...]

PR #1687, opened 2021-11-15
Replace Timeago with momentjs

Tallyrald:
As described in #1685 fuzzy time display is incorrect in Hungarian (and possibly other languages). This PR replaces the timeago package with momentjs inbuilt functionality `fromNow()`.

I also fixed an issue with the relevant table rows where the blade template set the datetime to ` 23:59:59` when no date was provided for the `best_before_date` or `purchased_date` fields. This resulted in an invalid date.

This is my first PR for this repo, please advise if I missed anything or if I should check other things as well.

### DIFF
--- a/localization/component_translations.pot
+++ b/localization/component_translations.pot
@@ -12,12 +12,6 @@ msgstr ""
 "Language: en\n"
 "X-Domain: grocy/component_translations\n"
 
-msgid "timeago_locale"
-msgstr ""
-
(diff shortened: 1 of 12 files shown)


berrnd (approved):
👍, thanks - works and behaves pretty much the same.

Moment.js seems to round for full days a little bit differently than jquery-timeago, but should not be a problem in general...

PR #1695, opened 2021-11-25
Improve sorting for stock entries view

MasterofJOKers:
In the stockentries view the ordering by expiration date, amount, price
and purchase date did not work as expected. b5fc64cf already addressed a
similar issue for other views. This commit now does the same: set
DataTables types on the specific columns.

Since the units being part of the "amount" column break numeric sorting,
this commit adds a "data-order" field for that column, so numeric
sorting can still work. This is done in aligment with the stockoverview
page, that already contains such an entry to facilitate proper sorting.


berrnd (approved):
👍, thanks!

PR #1696, opened 2021-11-26
Docker(-Compose) setup to aid local development.

roomcays:
I'm used to work locally with Docker, so when I started to look around the Grocy project I had to create Docker Compose environment.
I have explained my way of working with Docker in `.docker/README.md` file.
Changes do not affect any of application's behavior, they're just Docker Compose setup.
Maybe someone will find it helpful :)


berrnd:
Thanks, but a duplicate of #904 - quoting myself from https://github.com/grocy/grocy/pull/904#issuecomment-675562451:

> Thanks again, but I decided to not merge this.
> 
> I'm still thinking that dev environments heavily depend on personal preference, what someone finds useful and easy to set up, others don't. For me this feels much more complicated than just checking out the repo, opening it with VSCode while having some extensions installed which do the rest...
> 
> So I think we have enough shared ideas/experiences on how we work on grocy development here and in the referenced issue which others may find useful (or not).

PR #1710, opened 2021-12-06
LDAP: Use UID Attribute instead of POST Value

FloSet:
The LDAP provider used the transmitted POST Value as the username instead of the username attribute. Through this, the username was case-sensitive and for every possibility of mixing the char, a new user was created.


berrnd (approved):
👍, thanks!

PR #1746, opened 2022-01-12
Added support for reading auth header from env variable

Forceu:
This is a modification for ReverseProxyAuthMiddleware in order to support server env variables.

These are more secure than regular headers. An example for authelia running behind nginx would be:

```
       [...]
    server_name grocy.yourserver
    include snippets/authelia.conf;
    location / {
            include snippets/auth.conf;
            try_files $uri /index.php;
       }

    location ~ \.php$ {
        fastcgi_param  SCRIPT_FILENAME  $document_root$fastcgi_script_name;
        include        fastcgi_params;
             
        fastcgi_param REMOTE_USER $user;
        fastcgi_param REMOTE_GROUPS $groups;

        fastcgi_param  PATH_INFO        $path_info;
        fastcgi_pass unix:/run/php/php7.3-fpm.sock;
    }
```

In this case `fastcgi_param REMOTE_USER $user;` would set the env variable REMOTE_USER with the logged in user name


berrnd (approved):
Thank you!

PR #1761, opened 2022-01-24
Quick fix login page not centered

robocrax:
A very horrible way of doing it but hey, it works!

Horrible because content shifts after page load.


berrnd (changes requested):
Thanks.

> Horrible because content shifts after page load.

You named it. So if that's important for you, let's do it a little bit better.

berrnd:
Done in https://github.com/grocy/grocy/commit/7e6efb4a14700374adbc456fb50e5a551c39ba76.

PR #1819, opened 2022-03-14
fix: can't print in landscape

miguelangel-nubla:
Layout settings are hidden in the print dialog when printing a page like the mealplan.
Bootstrap 4 bug https://github.com/twbs/bootstrap/issues/25629


berrnd:
👍

Probably also landscape should be the default for the meal plan page.

berrnd (approved):
Thanks!

Issue #1119, opened 2020-11-04
Feature request: Automatic transfer workflow


matthiakl:
It would be nice to have an automatic transfer option for a product, when it is opened.
Say for example, I open a bottle of juice and then put it into the fridge. 
Or the pasta is stored in the basement, but when I open a pack I leave it in the kitchen.

berrnd:
Merged into #830.

berrnd:
This is now done in combination with #1365 (new product option "Default consume location") - changelog excerpt:

> - New product option "Move on open" (defaults to disabled)
>    - When enabled, on marking the product as opened, the corresponding amount will be moved to the products default consume location

Issue #830, opened 2020-05-05
Feature Request: Move stock to fridge after opening if indicated by the product ("transfer workflow")


CrushaKRool:
Some products can generally be kept at room temperature in your pantry, but then require to be moved to a refrigerated location after opening them. This is usually written on the product description near the hint about the expiration date.

It could be helpful to automate this process in Grocy:
1. Make it possible to mark locations as "is refrigerated" (similar to "is freezer").
2. Make it possible to mark products as "needs to be refrigerated after opening".
3. When marking a stock entry of such a product as opened, automatically transfer it to the refrigerated location. If multiple locations are marked as "is refrigerated", display a dialog to manually pick a destination.

berrnd:
This is now done in combination with #1365 (new product option "Default consume location") - changelog excerpt:

> - New product option "Move on open" (defaults to disabled)
>    - When enabled, on marking the product as opened, the corresponding amount will be moved to the products default consume location

Issue #1365, opened 2021-02-20
Feature request: standard consumption location


Saturo78:
Purchased Milk goes to the pantry. One pack gets transfered the the fridge and there it will be consumed.
It would be nice if i could specify a consumption location the same way a the standard storage location.

berrnd:
This is now done, there is this new optional product option for that:

![grafik](https://user-images.githubusercontent.com/2161815/161444172-7a300407-d6c7-4358-8d5d-c1d0cc2bd320.png)

Long explanation - changelog excerpt:

> - New product option "Default consume location" (not mandatory, defaults to not set / empty)
>     - When set, stock entries at that location will be consumed first
>     - => This will be automatically taken into account when consuming from the stock overview page and all other places where no specific location can be selected
>     - => On the consume page the location is preselected in the following order:
>         - 1. The new default consume location, if the product currently has any stock there, otherwise
>         - 2. The products default location, if the product currently has any stock there, otherwise
 >        - 3. The first location where the product currently has any stock

berrnd:
Cross referencing [#830](https://github.com/grocy/grocy/issues/830#issuecomment-1101549050) here - stock entries can now also be automatically moved to the products default consume location on opening them:

> - New product option "Move on open" (defaults to disabled)
>    - When enabled, on marking the product as opened, the corresponding amount will be moved to the products default consume location

PR #1863, opened 2022-04-18
Add support for "Move on Open"

RosemaryOrchard:
Related to #830  and #1119, this adds support for "move on open". It's implemented as a checkbox below the "Default consume location" select, and is disabled that matches the default product location. If checked, when you mark a product as open it moves it from wherever it is, to the default consume location (and falls back to wherever the product currently is, just in case there is ever a scenario where the consume location is not set). 

This is my first PR to Grocy, and my first time working with Slim, so I hope I've not done anything too crazy—feedback is very welcome! If this goes well I'd like to add support for fridges next.


services/StockService.php
```diff
@@ -1048,15 +1048,20 @@ public function OpenProduct(int $productId, float $amount, $specificStockEntryId
...
-				]);
+				];
 
 				$amount = 0;
 			}
+			if ($product->move_on_open) {
+				$update['location_id'] = $product->default_consume_location_id ?? $stockEntry->location_id;
+			}
+
+			$stockEntry->update($update);
```
berrnd, line 1064: To keep it consistent, transfer from/to stock journal entries should also be created when the product is "moved on open" (just like when the transfer transaction would have been done manually).

berrnd (changes requested):
Thanks!

This is not what #830 describes exactly, but maybe is enough, also to keep it simple - thanks for the idea.

Don't know how common it is to have more then one fridge while also storing _the same_ product sometimes in fridge1 and the other day in fridge2 after opening it. So I would say that's enough.

public/viewjs/consume.js
```diff
@@ -386,26 +388,22 @@ Grocy.Components.ProductPicker.GetPicker().on('change', function(e)
...
+								value: stockLocation.location_id,
+								text: stockLocationName
+							}));
+							$("#location_id").val(defaultLocationId);
+							$("#location_id").trigger('change');
+							setDefault = 1;
+							stockAmountAtDefaultLocation += Number.parseFloat(stockLocation.amount);
 
-							if (setDefault == 0)
+							if (!setDefault)
```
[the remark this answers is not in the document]
RosemaryOrchard: Oh whoops, this snuck in as I started with the fridge and thought I only staged the right parts, I'll revert that change!

public/viewjs/productform.js
```diff
@@ -239,7 +239,24 @@ $('#product-form input').keyup(function(event)
...
+	var consumeLocationLocation = document.getElementById('default_consume_location_id');
+
+	var moveOnOpen = document.getElementById('move_on_open');
+	if (consumeLocationLocation && defaultLocation !== consumeLocationLocation) {
+		moveOnOpen.removeAttribute('checked');
+		moveOnOpen.setAttribute('disabled', "");
+	} else {
+		moveOnOpen.removeAttribute('disabled');
+	}
+}
```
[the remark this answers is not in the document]
RosemaryOrchard: Thanks!

services/StockService.php
RosemaryOrchard: Thinking about it, should I flip the logic and check first, and then set the `stock_log` straight to the new location if there is one?

RosemaryOrchard: Read some more code: nope!

berrnd: A normal/manual transfer action creates two stock journal entries, the negative amount at the "from location" and the positive amount at the "to location":

grafik

Is done by [`StockService->TransferProduct`](https://github.com/grocy/grocy/blob/0152f1c69d3cb2bb3ceba668aa1006caa19a91c6/services/StockService.php#L1180) maybe just call that function there...

RosemaryOrchard: Thanks for the pointer, it makes much more sense to use the existing function.

RosemaryOrchard:
I would like to add support more in line with specifically what 830 was requesting, but this seemed like a less intrusive start—and solves the problem that I myself have right now.

berrnd:
Yes, I would also say that's what practically makes the most sense. When #830 was requested, the "Default consume location" idea wasn't there yet (ref #1365) - so in combination with that it definitely fits best.

Thanks again, I'll test that again later or tomorrow.

RosemaryOrchard:
Thank you very much! Both for the review, and the great app 😁

berrnd (approved):
👍, works.

I've additionally added some UI feedback when auto-moving happened to make that more transparent.

PR #1891, opened 2022-05-15
Handle stock_id in OpenProductByBarcode

jtommi:
While playing around with the API, I realised that /stock/products/by-barcode/:grocycode/open (OpenProductByBarcode) ignores the stock_id provided in the barcode.

But [ConsumeProductByBarcode](https://github.com/grocy/grocy/blob/af7de61c43e250bf03f1a3209e96c19c5efa9e9e/controllers/StockApiController.php#L327) does respect the stock_id.

So I compared both functions and adapted OpenProductByBarcode to mimic ConsumeProductByBarcode.

I tested the API afterwards and it works as expected


berrnd (approved):
Thanks!

Issue #1961, opened 2022-08-04
Bug: Note field of stock entries is not "persisted" or copied with transfer


akoshpinter:
Hi Bernd,

I am using the new Note field in stock entries to differentiate between multiple items in one kind of stock entry (e.g. what kind of dark chocolate it is). But I noticed a bug, that loses this information when I move some entries from one place to another, or mark them open.  I want to rely on this information to know what is expiring and what is in stock, without creating many products for different flavours/manufacturers etc..  Did not test it with custom/user fields, but maybe they have the same bug also.

EDIT: Also consuming multiple items seems to lose Notes (example Eggs).

I created a demo instance reproducing the problem (Dark Chocolate transfer/Milk chocolate open):

https://test-zke0d7xzff81wzxczuqrio.demo-prerelease.grocy.info

![image](https://user-images.githubusercontent.com/13643475/182839130-7f4de974-0712-4084-9e2c-d5b6ce5e260b.png)
![image](https://user-images.githubusercontent.com/13643475/182839143-780b1569-a017-4961-83fb-303c852c5a61.png)
![image](https://user-images.githubusercontent.com/13643475/182839173-934f10fa-5421-4506-8ce3-64b421b1c826.png)

Thank you once more for your great work!

berrnd:
> Did not test it with custom/user fields, but maybe they have the same bug also.

No. Userfields are handled entirely differently than any built-in field. `stock` Userfields are attached to the stock entry by its `stock_id`, that id stays the same during the entire lifecycle of the corresponding stock entry.

PR #1962, opened 2022-08-04
Fixes for losing Stock Note field content #1961

akoshpinter:
Fixes for losing Note field content when splitting stock entries, either by transfer, consume, opening or undoing transactions... #1961


berrnd (approved):
Thanks!

PR #1991, opened 2022-09-06
Translation update pt-br

webysther:
- Fixed absence of whitespace
- Few new translations


berrnd:
Thanks, but translations are managed via Transifex, not here.

README: https://github.com/grocy/grocy#localization

Transifex: https://www.transifex.com/grocy/grocy/language/pt_BR/

PR #1995, opened 2022-09-09
Wrong calculation of amounts when subproducts are substituted

derwassi:
There have been two bugs when calculating consumption of products when using subproducts:

1. The total amount of all subproducts had just been summed up which gives wrong results when subproducts have different base quantity units (e.g. "gramm" in the main product, and "Piece" (which is e.g. 1000g) in the subproduct). Therefore I replaced the simple sum with a simulation of the process.
2. When consuming data from multiple subproducts the amount is not transformed correctly. Therefore before (simulated) consumption the remaining amount is transformed back to it's original quantity unit.

Would be great, if you could merge this for one of the next releases.

Great project by the way!

### DIFF
--- a/services/StockService.php
+++ b/services/StockService.php
@@ -406,17 +406,15 @@ class StockService extends BaseService
 				$potentialStockEntries = FindAllObjectsInArrayByPropertyValue($potentialStockEntries, 'stock_id', $specificStockEntryId);
 			}
 
-			$productStockAmount = floatval($productDetails->stock_amount_aggregated);
-			if ($amount > $productStockAmount)
-			{
-				throw new \Exception('Amount to be consumed cannot be > current stock amount (if supplied, at the desired location)');
-			}
-
 			if ($transactionId === null)
 			{
 				$transactionId = uniqid();
 			}
 
+			//first simulate stock consumption. Otherwhise calculations are wrong when stock units of parent and child products have a big conversion factor
+			$stockeEntriesToConsume = [];
+
+			// collect all stockentries
 			foreach ($potentialStockEntries as $stockEntry)
 			{
 				if ($amount == 0)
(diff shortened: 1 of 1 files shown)


berrnd:
Thanks, but would be great to have the mentioned problems practically reproducible first, to also to be able to verify if things work afterwards.

&nbsp;
> The total amount of all subproducts had just been summed up which gives wrong results when subproducts have different base quantity units (e.g. "gramm" in the main product, and "Piece" (which is e.g. 1000g) in the subproduct)

Simple example: https://test-e9r3njsmsuf52rcz8o8pm9.demo-prerelease.grocy.info
`Rice Parent` is the parent product in Grams, `Rice Child` is the child product in Packs.
I can't reproduce any problem when consuming either the parent product or the child.

&nbsp;
> When consuming data from multiple subproducts the amount is not transformed correctly.

Another example scenario: https://test-a8n99w3m3wujqsjgqz3pbj.demo-prerelease.grocy.info
`Beer (Bottle)` is the parent product, which has `Beer (6xPack) ` and `Beer (20xCreate)` as childs.
When having 2 of each child in stock (= 52 Bottles, all have the same due date) and consuming 50 Bottles (the parent product), the consumption amount of `Beer (20xCrate)` is wrong.
=> So this is reproducible.

derwassi:
I see the Problem:
In the version i am using (installed it last week), this line: https://github.com/grocy/grocy/blob/master/services/StockService.php#L409 
was still different and caused the problem. 
I didn't check against current master/main branch (just copied it over). Sorry for that. I will update my pull request accordingly.

I did not spot the second problem directly, when changing the code I just saw, it was wrong. Thanks for producing working example.

berrnd:
The second reproducible example from above was tracked in #2076 for reference and fixed in dd409b4bf9a99b132d4cd7992ac2fff9da64ccad.

Issue #2039, opened 2022-11-13
Bug: Incorrect ingredient quantity displayed in recipe


esclear:
I configured a [demo instance](https://test-n4a1oa5u94vqeaehk7ot.demo-prerelease.grocy.info/) to highlight the issue.

I added [baking powder](https://test-n4a1oa5u94vqeaehk7ot.demo-prerelease.grocy.info/product/27) as a product, with a stock quantity unit of `pack`s and configured the following unit conversions:

![image](https://user-images.githubusercontent.com/7432848/201545937-79beed05-2a53-4a61-8773-05c9b316efa6.png)

In a [recipe](https://test-n4a1oa5u94vqeaehk7ot.demo-prerelease.grocy.info/recipe/7) I added two tea spoons of baking powder as ingredients.

![image](https://user-images.githubusercontent.com/7432848/201545991-20ca0812-1a04-4eac-aa8d-4fa57e9d7f55.png)

As you can see, the conversion is off, both in the blue text below the amount field, and in the ingredients table:
2 tea spoons = 2 * 4 g = 8 g, but not 8 packs nor 8 tea spoons.

I find this surprising, as grams are neither one of the units specified there nor the stock / purchase quantity unit.
Though there are some issues related to quantity conversions / recipes, I haven't found any that are open and/or seem to be related to this problem.

berrnd:
I don't find this surprising and it's not a bug, it's about #1360 - QU hierarchy is currently limited to one level.

Let's take your provided demo instance as the maybe 54th example about that here:

QU stock is always the base, so we need to go from _Pack_ to _Tea Spoon_ for that example:

Product stock QU _Pack_ **_Level 0_**
&nbsp;&nbsp;Product-QU conversion _Pack_ <-> _Gram_ **_Level 1_**
&nbsp;&nbsp;&nbsp;&nbsp;Product-QU conversion _Gram_ <-> _Tea Spoon_ **_Level 2  <-- Currently not supported_**

Create a direct conversion between _Pack_ and _Tea Spoon_ and everything will work.

&nbsp;
> Though there are some issues related to quantity conversions / recipes

Feel free to create a reproducible example for any issue and it will be fixed for the next release, at least this was the case for any reported problem in the last 5 years.

esclear:
> I don't find this surprising and it's not a bug, it's about #1360 - QU hierarchy is currently limited to one level.
> 
> Let's take your provided demo instance as the maybe 54th example about that here:
> 
[...]

Issue #1360, opened 2021-02-19
Feature request: Quantity unit hierarchy more than 1 level


ArchdukeTim:
I current have a unit Liter that is linked to milliliters and gallons directly. Gallon is linked to Quarts, Quarts to Pints, Pints to Cups, etc. 

I should be able to choose 'Teaspoon' as a unit for my product when adding to recipe because there is a conversion path between Liters (the stock unit) and 'Teaspoon'. I currently can access 'teaspoon' by checking the 'only check if any amount is in stock' option, but this seems more like a bad hack than a proper solution.

berrnd:
Duplicate of #1217, please check existing requests/issues before opening new ones.

berrnd:
Sorry, was the wrong issue - this is more about that you want to have more than one level of nested QU conversions - that's currently limited to 1 level (like parent/child products), will rename the issue accordingly to track that there.

Lykos153:
I'd really love to see this feature in Grocy! Would you accept a PR on this? I briefly glanced at the code and saw that a lot of the logic is done on the SQL side (with which I don't have that much experience yet, unfortunately). Would you mind sharing where you would start if you were to implement this feature?

berrnd:
> Would you accept a PR on this?

Every contribution was and is accepted, as long as it works, doesn't break anything else and doesn't introduce whole new concepts (like e.g. [rewriting the whole project](https://github.com/grocy/grocy/pull/1510), just because the tech-stack used and the tech-design-decisions made are not up to someones personally likings - something like that is better fits in an [own](https://github.com/HouseHold/HouseHold) [fork](https://github.com/mistressofjellyfish/not-grocy)).

&nbsp;
> saw that a lot of the logic is done on the SQL side

That's one of those base-tech-design-decisions for grocy and should stay, unless it's really not possible by that approach.

&nbsp;
> Would you mind sharing where you would start if you were to implement this feature?

The central SQL view which resolves QU conversions is [`quantity_unit_conversions_resolved`](https://github.com/grocy/grocy/blob/e8f76e5694f2defb69435bccfaeb8de96870322c/migrations/0189.sql) (the "final list", used everywhere else, which holds all possible conversions/factors per product).

[...]

Issue #389, opened 2019-09-26
Add support for fractions throughout


DeeeeLAN:
This is a great package:
https://github.com/infusion/Fraction.js/

Not sure if it would be feasible, but this would be really nice to have.

berrnd:
If it's about displaying numbers, that's already possible in plain JS, also already used in many places, but not all, like:

https://github.com/grocy/grocy/blob/c5927a10f3a682cfa917ff845e4c3191dc471e85/public/js/grocy.js#L516-L532

If it comes to how the numbers are stored in the database, that would be totally different as currently all number are stored as integers or decimals and AFAIK there is not SQL datatype (above all not in SQLite) which handles fractions itself...

DeeeeLAN:
Partially it is to make the numbers display nicer, but it would also be nicer for adding and consuming amounts, it would make a lot of it more even. I know that the package I linked stores the numerator and denominator as two separate integers, so I imagine that would be a good way to manage it in a database. You could just store whole numbers as #/1.

berrnd:
> I know that the package I linked stores the numerator and denominator as two separate integers

Yeah, I think that's the only possible way...
But that would make a big change how amounts are calculated on every place...
I give this a very low priority for now...

Personally I try to keep it as simple as simple, so don't track any partial amounts at all, means most of the time I only track a product once when it's used completely, not when it's partially used. In first place I was against all partial amount stuff ([ref](https://github.com/grocy/grocy/issues/123), [ref](https://github.com/grocy/grocy/issues/105), [ref](https://github.com/grocy/grocy/issues/32#issuecomment-408946222)), but obviously I were convinced enought to add it though... :D

> Partially it is to make the numbers display nicer

That I already tried to address in https://github.com/grocy/grocy/commit/21503c26d2f1eb44377056ac8035e0f91892d4fa (but not using fractions, only browser-locale-specific proper formatted decimal numbers).

DeeeeLAN:
Yeah that makes sense, definitely just a "nice to have" type thing, not a major issue. I would definitely be all for only tracking complete amounts, but that is tricky with recipes somewhat. I guess anything that uses a partial amount I would just not track stock, that would work. I might play around with that. 

Maybe if I want it enough I will look at adding this eventually. It is definitely a big shift like you said, so I would want to understand the code-base better first.

checketts:
With this feature not happening. Here is a silly hack that makes the stock `consume` a bit more firendly 

[...]

PR #2056, opened 2022-12-03
Add support for transitive unit conversions

esclear:
As discussed in #2039, here's a WIP pull request that adds support for transitive unit conversions to grocy.

When using it, I noticed that floating point inaccuracies may arise.
E.g. referring to the conversions defined in #2039, instead of `2 TL`, the UI will ` 1.9999 TL`.
However, this likely would have happened for a single conversion as well, depending on the factors, as the recipe ingredients are stored internally in the storage unit (?).

Regarding the query, I tried to keep the original one mostly the same and put it in the `single` CTE.
However, I removed the joins with the `quantity_units` table, as those were only used for the names and not relevant for the calculations, and moved them to the final `SELECT`.

When writing this, I thought that the `source` column mostly serves as a kind of debugging help.
I am inclined to remove the `source` column from the final result and possibly also remove it completely, if that's okay for you.
From the query in https://github.com/grocy/grocy/issues/2039#issuecomment-1336116371 I gathered, that this may be used for some sort of disambiguation of conversions, but this could be used to the `closure` CTE (and then the `LIMIT` would not be necessary anymore).

### DIFF
--- a/migrations/0203.sql
+++ b/migrations/0203.sql
@@ -1,0 +1,129 @@
+DROP VIEW IF EXISTS quantity_unit_conversions_resolved;
+
+CREATE VIEW quantity_unit_conversions_resolved AS
+WITH RECURSIVE
+	single(product_id, from_qu_id, to_qu_id, factor, source)
+AS (
+	-- 1. Product "purchase to stock" conversion factor
+	SELECT
+		p.id AS product_id,
+		p.qu_id_purchase AS from_qu_id,
+		p.qu_id_stock AS to_qu_id,
+		CASE WHEN p.qu_id_stock = p.qu_id_purchase THEN 1.0 ELSE p.qu_factor_purchase_to_stock END AS factor, -- Enforce a factor of 1 when QU stock = QU purchase
+		'1 product purchase to stock factor' AS source
+	FROM products p
+	UNION -- Inversed
+	SELECT
+		p.id AS product_id,
+		p.qu_id_stock AS from_qu_id,
+		p.qu_id_purchase AS to_qu_id,
+		1 / p.qu_factor_purchase_to_stock AS factor,
+		'1 product purchase to stock factor (inversed)' AS source
+	FROM products p
+	WHERE p.qu_id_stock != p.qu_id_purchase -- => Only when QU stock is not the same as QU purchase
+
+	UNION
+
+	-- 2. Product specific QU overrides
+	SELECT
+		p.id AS product_id,
+		quc.from_qu_id AS from_qu_id,
+		quc.to_qu_id AS to_qu_id,
+		quc.factor AS factor,
+		'2 product override' AS source
+	FROM products p
+	JOIN quantity_unit_conversions quc
+		ON p.id = quc.product_id
+
+	UNION
+
+	-- 3. Default (direct) QU conversion factors
+	SELECT
+		p.id AS product_id,
+		p.qu_id_stock AS from_qu_id,
+		quc.to_qu_id AS to_qu_id,
+		quc.factor AS factor,
+		'3 default direct factor' AS source
+	FROM products p
+	JOIN quantity_unit_conversions quc
+		ON p.qu_id_stock = quc.from_qu_id
+		AND quc.product_id IS NULL
+	JOIN quantity_units qu_from
+		ON quc.from_qu_id = qu_from.id
+	JOIN quantity_units qu_to
+		ON quc.to_qu_id = qu_to.id
+
+	UNION
+
+	-- 4. Default (indirect) QU conversion factors
+	SELECT
+		p.id AS product_id,
+		(SELECT from_qu_id FROM quantity_unit_conversions WHERE to_qu_id = quc.to_qu_id AND product_id = p.id) AS from_qu_id,
+		quc.from_qu_id AS to_qu_id,
+		(SELECT factor FROM quantity_unit_conversions WHERE to_qu_id = quc.to_qu_id AND product_id = p.id) / quc.factor AS factor,
+		'4 default indirect factor' AS source
+	FROM products p
+	JOIN product_qu_relations pqr
+		ON p.id = pqr.product_id
+	JOIN quantity_unit_conversions quc
+		ON pqr.qu_id = quc.from_qu_id
+		AND quc.product_id IS NULL
+	WHERE NOT EXISTS(SELECT 1 FROM quantity_unit_conversions qucx WHERE qucx.product_id = p.id AND qucx.from_qu_id = pqr.qu_id) -- => Product override exists
+),
+	closure(depth, product_id, path, from_qu_id, to_qu_id, factor, source)
+AS (
+	SELECT
+		1 as depth, product_id,
+		-- We need to keep track of the conversion path in order to prevent infinite loops
+		'/' || from_qu_id || '/' || to_qu_id || '/',
+		from_qu_id, to_qu_id,
+		factor, source
+	FROM single
+	
+	UNION
+	
+	SELECT
+		c.depth + 1, c.product_id,
+		c.path || s.to_qu_id || '/',
+		c.from_qu_id, s.to_qu_id,
+		c.factor * s.factor,
+		c.source || ' ~~> ' || s.source
+	FROM closure c
+	JOIN single s
+		ON  c.product_id = s.product_id
+		AND c.to_qu_id = s.from_qu_id
+	WHERE
+		c.path NOT LIKE ('%/' || s.to_qu_id || '/%')
+),
+	closure_rn(row_num, product_id, from_qu_id, to_qu_id, factor, source)
+AS (
+	SELECT
+		ROW_NUMBER() OVER(PARTITION BY product_id, from_qu_id, to_qu_id ORDER BY depth ASC),
+		product_id,
+		from_qu_id,
+		to_qu_id,
+		factor,
+		source
+	FROM closure
+)
+
+SELECT DISTINCT
+	-1 AS id, -- Dummy, LessQL needs an id column
+	c.product_id,
+	c.from_qu_id,
+	qu_from.name AS from_qu_name,
+	qu_from.name_plural AS from_qu_name_plural,
+	c.to_qu_id,
+	qu_to.name AS to_qu_name,
+	qu_to.name_plural AS to_qu_name_plural,
+	factor,
+	source
+FROM closure_rn c
+	JOIN quantity_units qu_from
+		ON c.from_qu_id = qu_from.id
+	JOIN quantity_units qu_to
+		ON c.to_qu_id = qu_to.id
+WHERE
+	c.row_num = 1
+ORDER BY
+	product_id, from_qu_id, to_qu_id;
--- a/migrations/0203.sql
+++ b/migrations/0203.sql
@@ -2,128 +2,145 @@ DROP VIEW IF EXISTS quantity_unit_conversions_resolved;
 
 CREATE VIEW quantity_unit_conversions_resolved AS
 WITH RECURSIVE
-	single(product_id, from_qu_id, to_qu_id, factor, source)
+
+-- First, determine conversions that are a single step.
+-- There may be multiple definitions for conversions between two units
+-- (e.g. due to purchase-to-stock, product-specific and default conversions),
+-- thus priorities are used to disambiguate conversions.
+-- Later, we'll only use the factor with the highest priority to convert between two units.
+
+	conversion_factors_dup(product_id, from_qu_id, to_qu_id, factor, priority)
 AS (
-	-- 1. Product "purchase to stock" conversion factor
+	-- 1. Product "purchase to stock" factors ...
 	SELECT
-		p.id AS product_id,
-		p.qu_id_purchase AS from_qu_id,
-		p.qu_id_stock AS to_qu_id,
-		CASE WHEN p.qu_id_stock = p.qu_id_purchase THEN 1.0 ELSE p.qu_factor_purchase_to_stock END AS factor, -- Enforce a factor of 1 when QU stock = QU purchase
-		'1 product purchase to stock factor' AS source
+		p.id, p.qu_id_purchase, p.qu_id_stock, p.qu_factor_purchase_to_stock, 50
 	FROM products p
-	UNION -- Inversed
+	WHERE p.qu_id_stock != p.qu_id_purchase
+	UNION
+	-- ... and the inverse factors
 	SELECT
-		p.id AS product_id,
-		p.qu_id_stock AS from_qu_id,
-		p.qu_id_purchase AS to_qu_id,
-		1 / p.qu_factor_purchase_to_stock AS factor,
-		'1 product purchase to stock factor (inversed)' AS source
+		p.id, p.qu_id_stock, p.qu_id_purchase, 1.0 / p.qu_factor_purchase_to_stock, 50
 	FROM products p
-	WHERE p.qu_id_stock != p.qu_id_purchase -- => Only when QU stock is not the same as QU purchase
+	WHERE p.qu_id_stock != p.qu_id_purchase
 
 	UNION
 
 	-- 2. Product specific QU overrides
+	-- Note that the quantity_unit_conversions table already contains both conversion directions for every conversion.
 	SELECT
-		p.id AS product_id,
-		quc.from_qu_id AS from_qu_id,
-		quc.to_qu_id AS to_qu_id,
-		quc.factor AS factor,
-		'2 product override' AS source
-	FROM products p
-	JOIN quantity_unit_conversions quc
-		ON p.id = quc.product_id
+		product_id, from_qu_id, to_qu_id, factor, 30
+	FROM quantity_unit_conversions
+	WHERE product_id IS NOT NULL
 
 	UNION
 
-	-- 3. Default (direct) QU conversion factors
+	-- 3. Default QU conversions are handled in a later CTE, as we can't determine yet, for which products they are
+	-- applicable.
 	SELECT
-		p.id AS product_id,
-		p.qu_id_stock AS from_qu_id,
-		quc.to_qu_id AS to_qu_id,
-		quc.factor AS factor,
-		'3 default direct factor' AS source
-	FROM products p
-	JOIN quantity_unit_conversions quc
-		ON p.qu_id_stock = quc.from_qu_id
-		AND quc.product_id IS NULL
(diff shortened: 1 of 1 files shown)


berrnd:
Thanks a lot!

Haven't looked at it technically so far, just gave it a quick try - demo instance (your added migration is manually applied there): https://test-3z8g2vbc6tp9uko4btp2jo.demo-prerelease.grocy.info

The only existing product there has mixed conversions for testing different cases of "conversion paths":
- Base / QU stock = Bottle
  - 1 Liter = 1 Bottles (Product override conversion)
  - 1 Pack = 6 Bottles (Product override conversion)
  - 1 Liter = 1000 Milliliters (Default conversion)
  - 1 Pack = 2 2xPacks (Default conversion)

Adding:
- 1 Bottle => shows 1 Bottle => correct 🟢
- 1 Liter => shows 0.167 Bottles => not correct (should be 1 Bottle) 🔴
- 1000 Milliliters => show 0.167 Bottles => not correct (should be 1 Bottle) 🔴
- 1 2xPacks => shows 2 Bottles => not correct (should be 12 Bottles) 🔴
- 1 Pack => shows 6 Bottles => correct 🟢

So based on that less cases work than before.

Looking at the output of `quantity_unit_conversions_resolved` directly it looks better (from what I can see at a first glance), so maybe it's "just" about the frontend.

Here is the database file of the above linked demo instance if you want to fiddle there:
[grocy.db.zip](https://github.com/grocy/grocy/files/10146867/grocy.db.zip)

In the in #1360 referenced duplicate issues a lot of other practical examples are mentioned, a couple of them should be used to verify that it works.

berrnd:
Would be great if you can add a little explanation about how you solved it / it works (for example what the idea about the CTEs `single`, `closure`, `closure_rn` is / what they represent).

&nbsp;
> E.g. referring to the conversions defined in #2039, instead of `2 TL`, the UI will ` 1.9999 TL`.
> However, this likely would have happened for a single conversion as well, depending on the factors, as the recipe ingredients are stored internally in the storage unit (?).

Yes, all amount and money values stored anywhere are related to the products stock QU. This can't be solved generally unless not using decimals (ref #389). If you store the values related to the destination QU (so for example for the QU used in the recipe ingredient), you just have the problem in the other direction (so proper display on the recipe ingredient, but having the rounding issue when doing the stock fulfillment check / consume transaction).

&nbsp;
> When writing this, I thought that the `source` column mostly serves as a kind of debugging help.

That's currently not only used for debugging, but also to make the hierarchy work. Product specific overrides of a maybe existent same default conversion should override it. The view is ordered by that column and the consumer should do a SELECT as mentioned in https://github.com/grocy/grocy/issues/2039#issuecomment-1336116371 to get the right factor for a specific conversion (maybe that's also a problem of the current state / why it doesn't work for the frontend).

esclear:
> Would be great if you can add a little explanation about how you solved it / it works (for example what the idea about the CTEs `single`, `closure`, `closure_rn` is / what they represent).

Sure, I'll do that.

> That's currently not only used for debugging, but also to make the hierarchy work. Product specific overrides of a maybe existent same default conversion should override it. The view is ordered by that column and the consumer should do a SELECT as mentioned in [#2039 (comment)](https://github.com/grocy/grocy/issues/2039#issuecomment-1336116371) to get the right factor for a specific conversion (maybe that's also a problem of the current state / why it doesn't work for the frontend).

The query / view, as I have written it, does not contain any duplicate conversion.
That is, for every triple (product, start quantity, end quantity), there is exactly one row and therefor one factor.

Regarding the examples you have given in https://github.com/grocy/grocy/pull/2056#issuecomment-1336190121, I was wondering, how the conversion factor for the conversion `1000 Milliliters => show 0.167 Bottles` is calculated.
Is it just the inverse of the Bottles → Milliliters conversion factor?

I'm asking that because it seems that the (inverse?) default conversion Milliliters → Liters is not taken into account here for the original query, as otherwise there would be a corresponding row here.

If that's okay for you, I can rewrite the query a bit more, though I'm not quite sure regarding the semantics of the factors yet.

As I gather from the [original query](https://github.com/grocy/grocy/blob/e8f76e5694f2defb69435bccfaeb8de96870322c/migrations/0189.sql), there are four sources for conversions:

1. purchase to stock conversion factors (and their inverses)
2. product specific QU overrides
3. default QU conversions

And for every product, the precedence of these conversions is the order they are listed above, i.e. if there is a purchase to stock conversion, the product specific override is ignored?

Can you elaborate on what the purpose of the 4th variant (default indirect qu conversion) is exactly?

Is there any documentation on the conversions that may exist and their precedences?
I didn't find any, but this would clear this up, I think.

berrnd:
> Can you elaborate on what the purpose of the 4th variant (default indirect qu conversion) is exactly?
> Is there any documentation on the conversions that may exist and their precedences?

It's just "one (and the only -> current limitation) more level", beside directly related conversions. So for my example above, that is the `Milliliters <-> 2xPacks` conversion. Most probably that's not longer needed when having your / a generic approach to resolve conversion indefinitely.

There is no documentation about anything, beside some code comments or what's discussed here and on [reddit](https://www.reddit.com/r/grocy/). If something is unclear, just ask, I'm happy to answer anything.

&nbsp;
> Regarding the examples you have given in https://github.com/grocy/grocy/pull/2056#issuecomment-1336190121, I was wondering, how the conversion factor for the conversion `1000 Milliliters => show 0.167 Bottles` is calculated.

That's the output of `quantity_unit_conversions_resolved` for my example product above:

grafik

The `productamountpicker` component consumes that [here](https://github.com/grocy/grocy/blob/3b3212d101f74d180af50c7c4dbc08adc2d9bac4/public/viewjs/components/productamountpicker.js#L14-L26) (in the order of the view, so `ORDER BY source`), but only the first per `to_qu_id`. That means the above yellow highlighted conversion is picked for that example (1000 / 6000 = 0,1666666666666667).

There was currently no need to filter on `from_qu_id` there (since there was no unlimited hierarchy). It's of course also ok to change that part, maybe this is all to pick the `Bottle<->Milliliters` conversion (untested):
https://github.com/grocy/grocy/blob/3b3212d101f74d180af50c7c4dbc08adc2d9bac4/public/viewjs/components/productamountpicker.js#L22

Changing this line to (picking only factors where `from_qu_id` = the products stock QU):
```js
if (!$('#qu_id option[value="' + conversion.to_qu_id + '"]').length && conversion.from_qu_id == destinationQuId)
```

&nbsp;
> Does the query need to return a factor of `1.0` for products where the purchase unit matches the stock unit?

Currently not needed because of:
https://github.com/grocy/grocy/blob/3b3212d101f74d180af50c7c4dbc08adc2d9bac4/public/viewjs/components/productamountpicker.js#L17-L20

berrnd:
About the products `Factor purchase to stock quantity unit`:

That only exists due to historic reasons, it was was once the only factor / conversion before the generic conversion handling was added in [v2.5.0](https://grocy.info/changelog#v2.5.0)).

I'm also open to and already thought of removing it. So migrating that factor (when not `1`) to a product specific conversion + removing that field. Maybe that would make this here easier.

esclear:
Okay, I decided to rewrite the view from the ground up and added some comments.
If there's anything unclear with the query, just ask.

I have tested it with the following setup that should, I think, cover all possibilities:
* Five units `A`, `B`, `C`, `D`, `E`
* Just one product (id = 1)
* The following conversions:
  * Default conversion A → B with a factor of 2
  * Purchase→Stock conversion B → C with a factor of 3
  * Default conversion B → C with a factor of 11
  * Product-specific conversion C → D with a factor of 5
  * Default conversion D → E with a factor of 7

As a "graphic":
```
Unit  Factor  Type
A
↓        2    Default conversion
B
↓        3    Purchase → Stock conversion
C
|      (11)   Default conversion (shall NOT be used)
↓        5    Product-specific conversion
D
↓       7    Default conversion
E
```

So in the end the result should consist of 20 conversions (with 20 distinct values), namely those, where the start and end units are different.
In particular, there should be a conversion from A to E with a factor of 2*3*5*7 = 210 and an inverse one from E to A with a factor of ~ 0.00476.

And indeed, the result looks as follows:

|id|product_id|from_qu_id|from_qu_name|from_qu_name_plural|to_qu_id|to_qu_name|to_qu_name_plural|factor|source|
|--|--|--|--|--|--|--|--|--|--|
|-1|1|1|A|As|2|B|Bs|2.0|/1/2/|
|-1|1|1|A|As|3|C|Cs|6.0|/1/2/3/|
|-1|1|1|A|As|4|D|Ds|30.0|/1/2/3/4/|
|-1|1|1|A|As|5|E|Es|210.0|/1/2/3/4/5/|
|-1|1|2|B|Bs|1|A|As|0.5|/2/1/|
|-1|1|2|B|Bs|3|C|Cs|3.0|/2/3/|
|-1|1|2|B|Bs|4|D|Ds|15.0|/2/3/4/|
|-1|1|2|B|Bs|5|E|Es|105.0|/2/3/4/5/|
|-1|1|3|C|Cs|1|A|As|0.166666666666667|/3/2/1/|
|-1|1|3|C|Cs|2|B|Bs|0.333333333333333|/3/2/|
|-1|1|3|C|Cs|4|D|Ds|5.0|/3/4/|
|-1|1|3|C|Cs|5|E|Es|35.0|/3/4/5/|
|-1|1|4|D|Ds|1|A|As|0.0333333333333333|/4/3/2/1/|
|-1|1|4|D|Ds|2|B|Bs|0.0666666666666667|/4/3/2/|
|-1|1|4|D|Ds|3|C|Cs|0.2|/4/3/|
|-1|1|4|D|Ds|5|E|Es|7.0|/4/5/|
|-1|1|5|E|Es|1|A|As|0.00476190476190476|/5/4/3/2/1/|
|-1|1|5|E|Es|2|B|Bs|0.00952380952380952|/5/4/3/2/|
|-1|1|5|E|Es|3|C|Cs|0.0285714285714286|/5/4/3/|
|-1|1|5|E|Es|4|D|Ds|0.142857142857143|/5/4/|

I have also added the database used for these tests here: [grocy.db.zip](https://github.com/grocy/grocy/files/10147504/grocy.db.zip)

berrnd:
Thanks again a lot for your work!

The view-output looks good and I also like and it's understandable how you structured it - thanks for adding the comments. 👍

I have applied the updated view on the [demo instance from yesterday](https://test-3z8g2vbc6tp9uko4btp2jo.demo-prerelease.grocy.info) and the "UI problems" are the same there - I will have a look at that later. Most likely not a big thing, as also mentioned above.

I also faced this:
On my dev machine I'm currently using SQLite 3.33.0 and there the view produced this error:
```
SQLSTATE[HY000]: General error: 1 circular reference: closure
```

It's no problem on the demo server, which is currently using SQLite 3.34.1.

Turns out that it's this enhancement of [SQLite 3.34.0](https://sqlite.org/releaselog/3_34_0.html):
> Enhance [recursive common table expressions](https://sqlite.org/lang_with.html#recursivecte) to support two or more recursive terms as is done by SQL Server, since this helps make [queries against graphs](https://sqlite.org/lang_with.html#rcex3) easier to write and faster to execute.

I guess you use SQLite >= 3.34.0?

So we need to raise the required SQLite version (I have no problem with that).
Probably makes sense to also finally raise the PHP requirement to 8.1 (there are breaking changes which affect grocy, so something to do there) and to also do "remove the products qu-purchase-to-stock-factor", to have all breaking changes in one release (means grocy 4.0).

berrnd (approved):
The view looks good and works, thanks again - required changes for the frontend to properly pick the conversion factors will be done outside of this pull request.

berrnd:
The required changes to make this work for the frontend are done and the pre-release demos are updated, if you want to try it.

From what I can tell, your original example in https://github.com/grocy/grocy/issues/2039#issue-1447104244 now also works as expected: https://test-n4a1oa5u94vqeaehk7ot.demo-prerelease.grocy.info/recipe/7

esclear:
Great, thanks a lot.

> I guess you use SQLite >= 3.34.0?#
Yes, I'm on 3.39.4

The query could technically be rewritten to work with versions older than 3.34.0, however it would not be nearly as readable.
And considering that SQLITE 3.34 is two years old at this point, I think it is fair to use this as a minimal requirement.

Today it occurred to me that there is a small problem with the query in this PR, regarding products to which only default conversions apply.
However the fix is easy and I'll open another PR for it.

berrnd:
There seem to be performance issues when having a couple of more products which need to be resolved before releasing that - see https://github.com/grocy/grocy/issues/2081#issuecomment-1364539612 (an example database is attached there in the first comment).

> About 550 products, no global QU conversions, no product specific conversion, so only "Factor purchase to stock quantity unit"-based conversions.
> 
> Querying `quantity_unit_conversions_resolved` for a single product takes about 7 seconds, of course not so fast and this needs to be improved before releasing the new QU conversion handling.
> 
> ```
> Ausführung wurde ohne Fehler beendet.
> Ergebnis: 3 Zeilen in 6927ms zurückgegeben
> In Zeile 1:
> SELECT *
> FROM quantity_unit_conversions_resolved
> WHERE product_id = 3
> ```

esclear:
Oh yeah, that's a problem.
I didn't experience any problems myself, though I only have way less products.

I'll have a look at the database provided in #2081.

A first thought would be to materialize the view in some way, though that's not possible with SQLITE.

However, this should only be a last resort and I'm honestly surprised that the query is so slow.

esclear:
My first observation is that for me it doesn't take nearly as long, just about 0.5 seconds.
```
λ time sqlite3 grocy.db 'SELECT * FROM quantity_unit_conversions_resolved WHERE product_id = 3;'
-1|3|3|Packung|Packungen|4|Gramm (g)|Gramm (g)|16.0|/3/4/
-1|3|4|Gramm (g)|Gramm (g)|3|Packung|Packungen|0.0625|/4/3/
-1|3|4|Gramm (g)|Gramm (g)|4|Gramm (g)|Gramm (g)|1.0|/4/4/

________________________________________________________
Executed in  505.29 millis    fish           external
   usr time  504.24 millis  259.00 micros  503.98 millis
   sys time    1.07 millis   68.00 micros    1.00 millis
```

That is still way too long for my taste, but – considering that the device I tested the query on is not that powerful – makes me suspect that there may be more to it.
I'm thus wondering with version of sqlite you used to test this, @CommanderRoot, @berrnd.
I'm currently on 3.39.4.

berrnd:
Thanks for having such a quick look.

I use SQLite 3.34.1 on my development machine and on the demo server and 3.35.5 when working with [DB Browser for SQLite](https://sqlitebrowser.org/) (guess it was bundled with that).

Did a quick test with the latest 3.40.0 SQLite release, and yeah - it's faster there.

But making SQLite >= 3.39.4 a requirement would be a big step...

esclear:
Okay, so reading the changelog may give some hints as to what changed and where the improvements stem from.
For now I'll have a look at the query plans and maybe find some info there.

esclear:
A quick look confirmed my suspicion that the bottle neck is related to the default conversions that are only added when they are "reached" by product units.

I'll have a closer look at this when the festive social obligations are done.
Cheers! 🎄

berrnd:
Thanks again and absolutely no need for a hurry - the next release is still kind of far away mostly due to things related to the PHP 8.1 upgrade.

And if the `Factor purchase to stock quantity unit` is the cause for the problems, maybe let's finally remove that (as [mentioned above](https://github.com/grocy/grocy/pull/2056#issuecomment-1336208338)).

But now it's christmas things - have a beautiful time.

esclear:
My suspicion proved correct, on my machine the query time got down from ~1800 ms to ~40 ms for the database provided in #2081.
The reason was that sqlite produced rather inefficient query plans (doing a full table scan for the joins) and by rewriting the query a bit, we can give it enough hints to prevent that.

There's some "quality control" left that I'd like to do, including comparing the query plans with older, i.e. >= 3.34, versions of sqlite.

Although that isn't needed, I think we could still replace the purchase to stock factor as mentioned before.
I think the main improvement would be that the whole conversion affair becomes easier to think about – I have to admit the factor confused me in the beginning.
Also, I suppose that this would make the code simpler elsewhere – the query wouldn't be affected much.

By the way, I noticed that there are two migrations for the transitive conversions, `0203.sql`, the one from this PR and `0204.sql` that are almost identical (safe for a column name and some formatting changes).
Is that on purpose? Would it be sensible to coalesce these again?
Or should I put the view with the improvements in a new migration?

berrnd:
> Is that on purpose? Would it be sensible to coalesce these again?
> Or should I put the view with the improvements in a new migration?

Migrations are only supposed to work between releases, so an existing unreleased migration about the same view could be changed. However sometimes I decide to create a new migration file, even for small changes - in this case there were other issues related to QU conversions reported and reproduced on pre-release demo instances - existing demo instances wouldn't catch up changed migrations, that was the only reason there.

berrnd:
I've removed the products `qu_factor_purchase_to_stock` (migrating existing factors to normal product specific QU conversions) in 0229d187aedfe50446ad42887810e70ba0b86109.

esclear:
Okay, I finished my tests with older sqlite versions.
The query plan look all sensible.
I did a very unscientific benchmark on the database from #2081 that yielded the following times:

|sqlite version|query duration as of 15f7ac3da02556a74e1d80e7d090af63faa1ba5b |duration with improvements|
|---|---:|---:|
|3.34.1|1040 ms| 84 ms|
|3.35.5| 719 ms| 82 ms|
|3.36.0| 702 ms| 83 ms|
|3.37.0| 673 ms| 84 ms|
|3.38.5|520 ms| 56 ms|
|3.39.4|528 ms|16 ms|

That factor ~12 reduction on 3.34.1 is a win in my book, though I wish that more improvement was possible.
Furthermore, although it is not needed per se, I would personally lean to _recommending_ sqlite >= 3.38 or even >= 3.39 for the (stressing this again: anecdotal) speed gains I observed.

I'm currently preparing a PR with the proposed change to the view.

berrnd:
Thanks again for your time you put into this - looks good from the tests I would say.

&nbsp;
> I would personally lean to recommending sqlite >= 3.38 or even >= 3.39 for the (stressing this again: anecdotal) speed gains I observed.

For me requiring current runtime stuff (which most of the times runs faster) would be no problem, however, there will be (as it was always the case when raising requirements) a kind of shitstorm when doing that. People run grocy on shared hosting which tend to use ancient runtime stuff just like always, for whatever reason. And e.g. the current PHP(-FPM) 8.1 package (at least on Debian) seems to bundle SQLite 3.34.1 - so I would prefer to leave it at that for now, but we can of course add a note on the changelog and/or README that a newer SQLite version is better for performance.

PR #2057, opened 2022-12-04
Fix quantity_unit_conversions_resolved view

esclear:
As I wrote in https://github.com/grocy/grocy/pull/2056#issuecomment-1336459947, the query from the #2056 has a problem in which it would not return any rows for products to which only default conversions apply.

This change makes this work and also subsumes the product-to-product conversion with a factor of 1.0 of [daa0a59](https://github.com/grocy/grocy/commit/daa0a59c5f07b24526a55cb9bf40f29fffd431be#diff-790995ae53772a2f70114217da92c75a9bb6d027099fd7a04543f7d2de699c72R175-R189).


berrnd (approved):
Great, thanks again!

PR #2070, opened 2022-12-20
Update config docs-in-comment for BASE_URL.

alexfornuto:
I'm running Grocy in Docker behind a reverse proxy, and when looking to adjust the URL so I wasn't redirected to `localhost`, I skipped over this setting at first because of the edited language.

Please consider this a starting point. I didn't want to overstep, but I think this could be adjusted further to explain that this value should be the protocol and domain Grocy is hosted at.

### DIFF
--- a/config-dist.php
+++ b/config-dist.php
@@ -53,7 +53,7 @@ Setting('CURRENCY', 'USD');
 Setting('BASE_PATH', '');
 
 // The base URL of your installation,
-// should be just "/" when running directly under the root of a (sub)domain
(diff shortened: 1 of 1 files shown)


berrnd:
Well, those two words don't change the world or make it more clear, as it is IMHO already pretty clear that it should be simply the public URL of your grocy instance. If it needs to have a trailing slash or not also depends on your webserver configuration.

And furthermore from my experience with (not only) 5+ years grocy support requests, people don't read documentation or config options comments in reality.

Thanks anyways.

Issue #2051, opened 2022-11-24
Feature Request: Recipe fullscreen ingredients and preparation both visible


mx03:
In recipe fullscreen mode, it would be great if the ingredients and the prepation are both visible if the width of the site is large enough.

Not sure if its only me, but its a way better cooking experience on tablet, if i can see the amount and the ingredients beside the preparation, without switching all the time between these two tabs.

As an example
https://codepen.io/0x03x/pen/KKeRwje

berrnd:
Something for that was started by @adamschmalhofer in #1628 about a year ago, but not finished so far.

Feel free to catch up on that, most likely "only" CSS fiddling...

adamschmalhofer:
Yeah, I lost interest in upstreaming the changes. I've been using the changes ever since and it works without merge conflicts with current grocy (v3.3.2 as of writing). The change is rather small and only touches one css file.

Dehalion:
@berrnd @adamschmalhofer What would be necessary to get it into upstream?
Should this be a system-wide setting or a per-view choice (e.g. button in the recipe view)?

berrnd:
> What would be necessary to get it into upstream?

To fix the UI glitches mentioned in https://github.com/grocy/grocy/pull/1628#pullrequestreview-771759543.

&nbsp;
> Should this be a system-wide setting or a per-view choice (e.g. button in the recipe view)?

What was started so far in #1628 was not not to make this optional / introduce a new setting somewhere - but feel free to start also implementing that, if you think it's useful and/or necessary.

Dehalion:
My bad, I just looked at the changes, not at the issues mentioned below.
I'll look into it.

adamschmalhofer:
Just tested it. Works great! Thank you.

PR #2073, opened 2022-12-20
Display ingredients and recipe side by side

Dehalion:
This tries to improve PR #1628 by @adamschmalhofer to implement Feature Request #2051

### DIFF
--- a/changelog/70_UNRELEASED_xxxx.xx.xx.md
+++ b/changelog/70_UNRELEASED_xxxx.xx.xx.md
@@ -9,6 +9,8 @@
 ### Stock
 
 - Quantity unit conversions now support transitive conversions, means the QU hierarchy has now unlimited levels (thanks a lot @esclear)
+- Fixed that hiding the "Purchased date" column (table options) on the stock entries page didn't work
+- Fixed that the consumed amount was wrong, when consuming multiple substituted subproducts at once and when multiple/different conversion factors were involved
 
 ### Shopping list
 
@@ -16,7 +18,8 @@
 
 ### Recipes
 
-- xxx
+- Fixed that hiding the "Requirements fulfilled" column (table options) on the recipes page didn't work
+- Fixed that ingredient costs and calories were wrong when product substitution and unit conversions were involved at the same time
 
 ### Meal plan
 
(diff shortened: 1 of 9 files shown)


Dehalion:
This just displays ingredients and recipe steps side by side and hopefully fixes the problems with multi recipies.
For those it would be even nicer to have them stacked from top to bottom instead of hidden by recipe tabs, but the order is in reverse for that and there are not much unique classes. That's also why I resorted to attribute selectors.

berrnd (changes requested):
Standard 1920x1080 resolution:

The ingredients column is IMHO not really that readable when having it so narrow (maybe collapsing the stock fulfillment info into a tooltip or something would be a good approach):

grafik

Even more a problem on smaller screens - tested by the common 1366x786 resolution:

grafik

For multi-/nested recipes and the rest definitely better. 👍

berrnd:
Technical side note: Since those CSS styles are only required for the /recipes page, it better fits into that view than in the global grocy.css file - so here:

https://github.com/grocy/grocy/blob/2ebb9b2cd91e52e132171ef53fbb63fcb0373da2/views/recipes.blade.php#L8-L13

Dehalion:
> Technical side note: Since those CSS styles are only required for the /recipes page, it better fits into that view than in the global grocy.css file

I was thinking the same, but didn't find any more CSS files. Didn't look into the view files on my own.
I moved it there, but Github Code spaces messed up my branch with merges, I will clean it up and address your concerns with the width of the ingredients column.

What do you mean by "stock fulfillment info"?
And is there a library for nice tooltips and an example location to look at? Or do you mean just adding a `title` attribute?
Would it be a problem on mobile platforms to move this info into a tooltip?

berrnd:
> What do you mean by "stock fulfillment info"?

This:
grafik

&nbsp;
> And is there a library for nice tooltips and an example location to look at?

grocy uses Bootstrap (v4) and therefore their tooltips throughout - example:
grafik

Doc ref: https://getbootstrap.com/docs/4.2/components/tooltips

berrnd:
I've finished this now (using another approach) in daf5ad33c7010dea9e8d8c85d5ecc42d7c9a9b3e.

Issue #2081, opened 2022-12-24
Improve new transitive QU conversion handling performance


CommanderRoot:
Hello,

Today I tried to upgrade my grocy instance to get a few bug fixes which affect me. I however encountered an issue as the page loading times increase from 1 - 3 seconds to 2 minutes or more (and the PHP-fpm process is around 100% CPU usage during that time), making it basically unusable.
I'm running grocy on a Raspberry Pi 3b+ with PHP 8.1. It's not the fastest machine but it was fine until now.
I tried switching between the commits to find out what was causing it and the only way to get the old loading times back was to run migration `0189.sql`, which replaces the new conversion method with the old one.

I also tried to run the view `uihelper_stock_current_overview` with `DB Browser for SQLite` on Windows and that took ~ 1 minute as well. So it's faster on my normal PC but even that wouldn't be fast enough for normal usage.

I have attached my grocy.db file to hopefully make reproducing easier.

[grocy.zip](https://github.com/grocy/grocy/files/10298651/grocy.zip)

berrnd:
Any change related to #2056 is not yet released, the upgrade to PHP 8.1 is also not yet finished, therefore also not yet released - so from that side nothing to be tracked as a bug. It's nowhere stated that the current development state is stable in any case.

Database migrations maybe change multiple times during development, so triple check you've applied them correctly when using the master branch.

When it's reproducible related to the new QU conversion handling, then you most probably have a circular reference in your defined conversions, which is subject to [GIGO](https://en.wikipedia.org/wiki/Garbage_in,_garbage_out).

In general, performance will never be optimized for something like a Raspberry Pi - I personally also run Pis for some things, but would never use it for web applications like grocy, but I also like stuff to be fast.

I maybe have a look at your attached database later.

berrnd:
I had a look at the attached database and found no definition problems so far.

About 550 products, no global QU conversions, no product specific conversion, so only "Factor purchase to stock quantity unit"-based conversions.

Querying `quantity_unit_conversions_resolved` for a single product takes about 6 seconds, of course not so fast and this needs improvement.

```
Ausführung wurde ohne Fehler beendet.
Ergebnis: 3 Zeilen in 6927ms zurückgegeben
In Zeile 1:
SELECT *
FROM quantity_unit_conversions_resolved
WHERE product_id = 3
```

So thanks for testing, let's track here to improve that, even it belongs more to #2056 and #1360.

[...]

PR #2082, opened 2022-12-25
Add migration for improved quantity_unit_conversions_resolved view

esclear:
This PR proposes the migration discussed in https://github.com/grocy/grocy/pull/2056#issuecomment-1364540297 to address the performance problems raised in #2081.

Notably, as 0229d187aedfe50446ad42887810e70ba0b86109 removed the `qu_factor_purchase_to_stock`, the query could be simplified again, removing the priority system.

The main change, however is the `default_conversions` CTE, which is later used when calculating the transitive hull.


berrnd (approved):
I can confirm the performance improvements. 👍

PR #2098, opened 2023-01-06
Update strings.po

vanhoof-tech:
Add a space between the number and the rest of the string.


berrnd:
Thanks, but translations are managed via Transifex, not here.

README: https://github.com/grocy/grocy#localization

Transifex: https://www.transifex.com/grocy/grocy/language/nl/

PR #2126, opened 2023-02-05
Consume all available products for unfulfilled recipe

### LINKED ISSUE #386: Consume all 'in stock' ingredients (recipes)
I think the "consume all ingredients" button of a recipe shouldn't be disabled if not all the ingredients are in stock. If I chose to skip an ingredient and not stock it, clicking the button should just consume what it can and skip the empty ones. 

A slight variation, I had 0.89 lb of meat but the recipe called for 1 lb. Clicking "consume all" should decrease this value to 0 as well.

JJK96:
Fix #386

Maybe this should not be default behaviour, but only availble when a flag is set. I don't know how to do that yet.

### DIFF
--- a/services/RecipesService.php
+++ b/services/RecipesService.php
@@ -74,12 +74,6 @@ class RecipesService extends BaseService
 			throw new \Exception('Recipe does not exist');
 		}
 
-		$recipeResolved = $this->getDatabase()->recipes_resolved()->where('recipe_id', $recipeId)->fetch();
-		if ($recipeResolved->need_fulfilled == 0)
-		{
-			throw new \Exception('Recipe need is not fulfilled, consuming not possible');
-		}
-
 		$transactionId = uniqid();
 		$recipePositions = $this->getDatabase()->recipes_pos_resolved()->where('recipe_id', $recipeId)->fetchAll();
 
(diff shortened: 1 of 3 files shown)


berrnd (changes requested):
Well, entirely changing a current behaviour is just like never the way to go. There should be at least a warning like "do you really want to continue" - e.g. by simply adding another button in the existing popup on the recipe page when consuming a recipe, or something like that.

Beside that changing the general stock consume logic for that is a rather bad idea, have you tested your changes beforehand at any chance?

It doesn't work, since your new added parameter `consume_all_available` for `StockService->ConsumeProduct`, which defaults to `false`, isn't used in `RecipesService->ConsumeRecipe`.

Also the API endpoint `/recipes/{recipeId}/consume` needs to be taken into account for that change (maybe an optional new request parameter to allow consuming partially fulfilled recipes)...

And of course the confirmation message after consuming such a partially fulfilled recipe should also properly reflect what happened...

Sorry and please don't get me wrong, I'm glad for any help, but Pull Requests like this produce just more work than anything else...

JJK96:
I tested it on my server, but I see that I made a mistake when applying the changes in Git. Let me fix it

JJK96:
I don't intend to cause you more work. Feel free to ignore it!

berrnd:
> Feel free to ignore it!

When this state is meant to be the final state for that, then I can only do that - see my review comment above for any remarks.

PR #2135, opened 2023-02-12
Feature: Stock Purchase Metrics

raupie:
Feature Request: Stock "Purchases" metrics

- group by Product vs Product Group
- Add pie charting via [canvasjs](https://canvasjs.com/jquery-charts/)
- Add improved "filter-by-date" via [daterangerpicker](https://www.daterangepicker.com/)

I added a few demo entries to add some values for `today` vs `yesterday` for daterangepicker filter testing.

The controller I built utilizes a custom `sql` statement as the `lessql` library doesn't support `groupby` for the sum calculations needed.

Related: https://github.com/grocy/grocy/issues/64
Related: https://www.reddit.com/r/grocy/comments/10r51sk/analytics_reporting_capabilities/


raupie:
I also believe my `daterangepicker` is an upgrade that can be applied on all the date range filter drop downs.

controllers/StockController.php
```diff
@@ -515,6 +515,53 @@ public function Stockentries(\Psr\Http\Message\ServerRequestInterface $request,
...
+		if (isset($request->getQueryParams()['byGroup']))
+		{
+			$sql = "
+			SELECT product_group_id as id, product_group as name, sum(quantity * price) as total
+			FROM product_purchase_history
+			where $where
+			GROUP BY product_group
+			ORDER BY product_group
+			";
+		} else {
```
berrnd, line 542: Would be great if the current code style could be used, e.g. new lines after curly braces, camel case variable names and so on ([ref for PHP code](https://github.com/grocy/grocy/blob/master/.php-cs-fixer.php)) - but I can also take care of that whenever merging that.

public/viewjs/metrics.js
```diff
@@ -0,0 +1,96 @@
...
+		verticalAlign: "bottom"
+	},
+	data: [{
+		type: "pie",
+		showInLegend: true,
+		toolTipContent: "<b>{label}</b>: ${y} (#percent%)",
+		indexLabel: "{label}",
+		legendText: "{label} (#percent%)",
+		indexLabelPlacement: "outside",
+		valueFormatSTringt: "#,##0.##",
```
berrnd, line 28: If that's about number formatting, should respect locale as any other place a number is shown ([ref](https://github.com/grocy/grocy/blob/91700e7dae4b35c32c2483ce0837a714df87a398/public/js/grocy.js#L633-L670)).

public/viewjs/metrics.js
```diff
@@ -0,0 +1,96 @@
...
+
+if (urlParams.get('start_date')) start_date = moment(urlParams.get('start_date')) ;
+if (urlParams.get('end_date')) end_date = moment(urlParams.get('end_date'));
+
+$('#daterange-filter').daterangepicker({
+	showDropdowns: true,
+	startDate: start_date,
+	endDate: end_date,
+	locale: {
+		"format": 'MM/DD/YYYY',
```
berrnd, line 60: grocy uses fixed ISO-8601 notation for any date and time input/display ([ref](https://github.com/grocy/grocy#input-shorthands-for-date-fields)).

public/viewjs/metrics.js
```diff
@@ -0,0 +1,96 @@
...
+		"firstDay": 1
```
berrnd, line 61: Should respect [`CALENDAR_FIRST_DAY_OF_WEEK`](https://github.com/grocy/grocy/blob/91700e7dae4b35c32c2483ce0837a714df87a398/config-dist.php#L25-L28) (and maybe [`CALENDAR_SHOW_WEEK_OF_YEAR`](https://github.com/grocy/grocy/blob/91700e7dae4b35c32c2483ce0837a714df87a398/config-dist.php#L30-31)) as any other place a calendar is shown.

public/viewjs/metrics.js
```diff
@@ -0,0 +1,96 @@
...
+	ranges: {
+		'Today': [moment(), moment()],
+		'Yesterday': [moment().subtract(1, 'days'), moment().subtract(1, 'days')],
+		'Last 7 Days': [moment().subtract(6, 'days'), moment()],
+		'Last 14 Days': [moment().subtract(13, 'days'), moment()],
+		'Last 30 Days': [moment().subtract(29, 'days'), moment()],
+		'This Month': [moment().startOf('month'), moment().endOf('month')],
+		'Last Month': [moment().subtract(1, 'month').startOf('month'), moment().subtract(1, 'month').endOf('month')],
+		'This Year': [moment().startOf('year'), moment().endOf('year')],
+		'Last Year': [moment().subtract(1, 'year').startOf('year'), moment().subtract(1, 'year').endOf('year')]
```
berrnd, line 72: Range label localization is missing (I guess).

views/layout/default.blade.php
```diff
@@ -734,6 +736,8 @@ class="col content-text">
 	@if(!empty($__t('bootstrap-select_locale') && $__t('bootstrap-select_locale') != 'x'))<script src="{{ $U('/node_modules', true) }}/bootstrap-select/dist/js/i18n/defaults-{{ $__t('bootstrap-select_locale') }}.js?v={{ $version }}"></script>@endif
 	<script src="{{ $U('/node_modules/jquery-lazy/jquery.lazy.min.js?v=', true) }}{{ $version }}"></script>
 	<script src="{{ $U('/node_modules/nosleep.js/dist/NoSleep.min.js?v=', true) }}{{ $version }}"></script>
+	<script src="{{ $U('/node_modules/daterangepicker/daterangepicker.js?v=', true) }}{{ $version }}"></script>
+	<script src="{{ $U('/node_modules/canvasjs/dist/jquery.canvasjs.min.js?v=', true) }}{{ $version }}"></script>
```
berrnd, line 740: When a library is only needed on a single page, it's not needed to include it globally (like [here](https://github.com/grocy/grocy/blob/91700e7dae4b35c32c2483ce0837a714df87a398/views/calendar.blade.php#L7-L16)).

views/stockmetricspurchases.blade.php
```diff
@@ -0,0 +1,121 @@
+@extends('layout.default')
+
+@section('title', $__t('Stock Metrics: Purchases'))
+@section('activeNav', 'stockmetricspurchases')
+@section('viewJsName', 'metrics')
```
berrnd, line 5: `viewJsName` should be named like the view, so `stockmetricspurchases`.

berrnd (changes requested):
First: Thanks a lot!

I've tried to have a quick look first practically how that works - was unable to do so, since the view (I guess) `product_purchase_history` is missing.

&nbsp;
> Add pie charting via canvasjs

What's the big benefit over using or what's not possible by utilizing the already included library [Chart.js](https://www.chartjs.org/) (as also mentioned [there](https://www.reddit.com/r/grocy/comments/10r51sk/comment/j72se6s/?utm_source=reddit&utm_medium=web2x&context=3))?

&nbsp;
> I also believe my `daterangepicker` is an upgrade that can be applied on all the date range filter drop downs.

There are no other places currently where a date range needs to be selected, only single date or time or date-time inputs are used. What would be the practically benefit of replacing [tempus-dominus](https://github.com/tempusdominus/bootstrap-4), which also means rewriting all currently used customizations / integrations?

public/viewjs/metrics.js
raupie: I refactored to the `chartjs` library. We can resolve this one.

controllers/StockController.php
raupie: I agree. I'll review and adjust the style to the preferred standard.

public/viewjs/metrics.js
raupie: Do you have any quick examples how to pull from the declared config the local data but in javascript?

public/viewjs/metrics.js
raupie: What do you mean? I believe it's in the `localizations/strings.pot` these new UI elements that require translation elements.

views/layout/default.blade.php
raupie: Ah, yes. No need to declare it globally. I fixed this.

views/stockmetricspurchases.blade.php
raupie: Got it. Easy fix.

raupie:
I like this doughnut label plugin for chartjs. 

https://www.npmjs.com/package/chartjs-plugin-doughnutlabel

I want to display the complete sum of the entire range in the center which is easy to do. The part I'm struggling is how to implement the users declared local but in javascript. I think most of the time the php blade template engine is handling the `$` formatting for example. 

Just not sure how I could implement this with javascript. 

```
totalAmount = (parseFloat(totalAmount) + parseFloat(itemTotal)).toFixed(2);
===
doughnutlabel: {
				labels: [
					{
						text: totalAmount,
						font: {
							size: 30,
							weight: 'bold',
						},
					},
					{
						text: 'total',
					}
				]
			}
```

raupie:
I did find it difficult to adopt chartjs as 2.8.0 is very old. I would be interested in taking a stab after all of this and look into updating this library. Seems it has a bunch of new things. Finding help online was challenging this many versions behind.

raupie:
> I've tried to have a quick look first practically how that works - was unable to do so, since the view (I guess) `product_purchase_history` is missing.

I have the new view in my migrations 0215.sql file. It's in my MR.

> > Add pie charting via canvasjs
> 
> What's the big benefit over using or what's not possible by utilizing the already included library [Chart.js](https://www.chartjs.org/) (as also mentioned [there](https://www.reddit.com/r/grocy/comments/10r51sk/comment/j72se6s/?utm_source=reddit&utm_medium=web2x&context=3))?

There isn't. I agree, let's refactor to chartjs. Which seems the version that's used is 2.8.0 which is a few major versions behind. I will submit an issue and attempt an upgrade down the road once metrics is implemented. I like chartjs better now that I've spent some time with it.

> > I also believe my `daterangepicker` is an upgrade that can be applied on all the date range filter drop downs.
> 
> There are no other places currently where a date range needs to be selected, only single date or time or date-time inputs are used. What would be the practically benefit of replacing [tempus-dominus](https://github.com/tempusdominus/bootstrap-4), which also means rewriting all currently used customizations / integrations?

Oh, I saw a few places where the dropdown date picker was an option. All the stock journal and stock entry pages have a `date range` drop down selection. Mine is way more dynamic and allows for a custom date range which would be super helpful and not limited to set ranges without customizations.

public/viewjs/metrics.js
berrnd: `Grocy.CalendarFirstDayOfWeek` for example [here](https://github.com/grocy/grocy/blob/91700e7dae4b35c32c2483ce0837a714df87a398/views/layout/default.blade.php#L99) - any place where a calendar / date-time-picker is used uses that...

public/viewjs/metrics.js
berrnd: Haven't tested it so far due to the missing view, but in the current state, I don't think `Today`, `Yesterday`, `Last 7 Days` and so on will be translated...

berrnd:
The sneak peak looks good, great! 👍

&nbsp;
> I have the new view in my migrations 0215.sql file. It's in my MR.

However it's still not here: https://github.com/raupie/grocy/tree/FR-stock-purchase-metrics/migrations

However, there is no need to hurry in general - I've converted this to draft for now, feel free to mark it as "Ready for review" again, when I should have a look for testing. Thanks again for your time working on this!

public/viewjs/metrics.js
raupie: corrected to `YYYY-MM-DD`

Perhaps a future issue to make this a localization preference via `config.php` setting?

controllers/StockController.php
raupie: Fixed :heavy_check_mark:

public/viewjs/metrics.js
raupie: Excellent, `daterangepicker` has an options flag for both. :heavy_check_mark: 

https://www.daterangepicker.com/#options

public/viewjs/metrics.js
raupie: Yep, I see what you mean now. How would I "test" the translations? 

I'm adding the keys to the primary `strings.pot` in the root of `localization` folder. 

Then I'm duplicating the keys:values in `de/strings.po` and flushing the cache `rm -rf data/viewcache/*`

config.php -> `DEFAULT_LOCALE=de`

raupie:
The missing migrations file. Due to my local git global ignore for not commit sql to git :face_with_spiral_eyes: 

Forced it in there for the review when I finish up the rest of the edits. Mystery solved :heavy_check_mark:

public/viewjs/metrics.js
berrnd: > Perhaps a future issue to make this a localization preference via `config.php` setting?

No, this is by design as [mentioned in README](https://github.com/grocy/grocy#input-shorthands-for-date-fields).

public/viewjs/metrics.js
berrnd: If you want to test it locally, then of course this would be a way to do it - I never do that since if the `__t()` function or whatever is broken, most likely also the untranslated text won't display.

rainforest1155:
I'm just getting started with Grocy and among other things, I also wanted to use it to keep track of other expenses, like take out from restaurants and purchasing non consumables like books. For this I found the user entity feature and I'm wondering if the reporting function you're working on could also be applied to items in custom user entities?
What would I need to keep in mind when creating user fields for this to be applicable? Do the fields need to have specific names, e.g. for the price? 
Thanks!

berrnd:
@rainforest1155 Please keep it to the topic - this here is about starting with stock analysis things by creating intial reports about product prices, nohting else. Thanks.

berrnd (approved):
I've finally manged to have a closer look at that - and renamed / moved around / streamline some stuff - thanks again!

raupie:
This is awesome! Thanks.

raupie:
Looks like there is an issue. 

Pulled master, can't see reports. 404 on the chartjs-plugin

Issue #2137, opened 2023-02-20
Feature Request: PWA


devkral:
Ever thought about making the grocy frontend a PWA?
This would eliminate the need for separate apps and is desktop compatible

Note: this needs some work if your frontend is not client side. Otherwise you basically just need a serviceworker, a webmanifest and chrome/chromium lighthouse (ideal for testing that the project is PWA conformant)

berrnd:
If you mean offline functionality and stuff: This won't be something I'll add for grocy. Consider using one of the community contributed apps for that.

If you mean only the "Add to home screen" functionality: Yes, thought about it and grocy has that since ever ([code ref 1](https://github.com/grocy/grocy/blob/91700e7dae4b35c32c2483ce0837a714df87a398/views/layout/default.blade.php#L31-L32), [code ref 2](https://github.com/grocy/grocy/blob/master/public/img/appicons/site.webmanifest)). I use that daily to access my shopping list, chores and more on mobile.

devkral:
sry, my bad: overlooked the install button

PR #2139, opened 2023-02-20
docs: link to frontends, document PWA capability

devkral:
The PWA install button changes from browser to browser, so just document, that this functionality is available (people can search it)

see #2137

Also link to the companion apps, so people can direct their questions to the right maintainers.


berrnd:
Thanks, but that's basically already covered by this section:

https://github.com/grocy/grocy/blob/91700e7dae4b35c32c2483ce0837a714df87a398/README.md?plain=1#L21-L23

And the average user doesn't know what a "PWA" is, so brings practically nothing than confusion.

But thanks again anyways.

PR #2322, opened 2023-08-22
stock_current view now uses cached QUCs -> Recipes consume 50% faster

alkuzman:
During the release of the cache__quantity_unit_conversions_resolved the change on the stock_current view was left out.


berrnd (approved):
👍

PR #2355, opened 2023-10-09
Store estimated chore execution date at chore execution and show it in journal

wasinix:
I missed the possibility in Grocy, to see if a chore had been done at the estimated time, or if it has been done late or in advance.
So I tried, out of curiosity, if I could implement this on my own.
This is the result and for me it is working.
Not sure if it fulfills your quality criteria and if you are happy about contributions like this and find it useful.

BR,
wasinix


berrnd (changes requested):
Thanks, don't know yet if we generally need that - technical remark:

Translations are not maintained and therefore not managed here ([README ref](https://github.com/grocy/grocy#localization)). So please revert the changes in all `*.po` files and only add the new strings in [`strings.pot`](https://github.com/grocy/grocy/blob/master/localization/strings.pot) (this also happens automatically when [`MODE = dev`](https://github.com/grocy/grocy/blob/c9215a9a4e929eaf324c593158e0e80277761c6f/config-dist.php#L20)).

services/ChoresService.php
```diff
@@ -187,11 +187,13 @@ public function TrackChore(int $choreId, string $trackedTime, $doneBy = GROCY_US
 			}
 		}
 
+		$estimatedExecutionTime = $this->getDatabase()->chores_current()->where('chore_id', $choreId)->min('next_estimated_execution_time');
```
berrnd, line 190: Why is `min('next_estimated_execution_time')` needed here? `chores_current` already only contains one row per chore...

berrnd (changes requested):
If we add that, I guess the info how late or early the corresponding execution was tracked would also be useful. Maybe indicated by a small red/green icon and more info in a tooltip or something similar.

services/ChoresService.php
wasinix: This has been in the begin a lazy copy/paste from function GetChoreDetails.
I later noticed the min and was wondering a bit about it and thought it will not do any harm and better be on the safe side, there might be reason for the "min"

berrnd: > if you are happy about contributions like this

Please don't get me wrong, but reviewing a contributed copy & paste job, without really understanding what's happening, takes probably fourth the time as if I would have implemented this on my own after the corresponding feature request was posted...

I'll maybe give this another try, but likely not very soon, thanks again.

wasinix: Understandable and totally fine for me.
Will close this PR and create a feature request instead the next time.
Sorry if I wasted your time and nevertheless thanks for this nice tool and sharing it to the public.

PR #2412, opened 2023-12-24
Added 'category' filter to tasks view.

geoffwright240:
The 'Tasks' view had an empty space for another default filter along the header bar. I have a lot of categories and it simplifies presentation to filter according to the task category without adding a lot of unnecessary clutter to the view.

image

I would be happy to add this as a configurable option but I wanted to check first whether we could just include it by default. Happy to take feedback or suggestions.

### DIFF
--- a/public/viewjs/tasks.js
+++ b/public/viewjs/tasks.js
@@ -41,6 +41,7 @@ $("#category-filter").on("change", function()
 	{
 		value = "";
 	}
+
 	tasksTable.column(tasksTable.colReorder.transpose(6)).search(value).draw();
 });
 
--- a/public/viewjs/tasks.js
+++ b/public/viewjs/tasks.js
@@ -42,7 +42,7 @@ $("#category-filter").on("change", function()
 		value = "";
 	}
 
-	tasksTable.column(tasksTable.colReorder.transpose(6)).search(value).draw();
+	tasksTable.column(tasksTable.colReorder.transpose(3)).search(value).draw();
 });
 
 $("#clear-filter-button").on("click", function()
--- a/public/viewjs/tasks.js
+++ b/public/viewjs/tasks.js
@@ -41,10 +41,6 @@ $("#category-filter").on("change", function()
 	{
 		value = "";
 	}
-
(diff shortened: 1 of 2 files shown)


public/viewjs/tasks.js
```diff
@@ -34,12 +34,28 @@ $("#status-filter").on("change", function()
...
+$("#category-filter").on("change", function()
+{
+	var value = $(this).val();
+	if (value === "all")
+	{
+		value = "";
+	}
+
+	// Transfer CSS classes of selected element to dropdown element (for background)
+	$(this).attr("class", $("#" + $(this).attr("id") + " option[value='" + value + "']").attr("class") + " form-control");
```
berrnd, line 46: What's this? You don't set a `class` attribute below for the `<option>`'s and task categories doesn't even have colors...

berrnd (commented):
Thanks.

public/viewjs/tasks.js
geoffwright240: You're right. I was just following the same pattern that we have in the 'status' filter. Let me look at it more carefully and it can probably be removed.

$("#status-filter").on("change", function()
{
	var value = $(this).val();
	if (value === "all")
	{
		value = "";
	}

	// Transfer CSS classes of selected element to dropdown element (for background)
	$(this).attr("class", $("#" + $(this).attr("id") + " option[value='" + value + "']").attr("class") + " form-control");

	tasksTable.column(tasksTable.colReorder.transpose(5)).search(value).draw();
});

geoffwright240:
Thanks so much for this great application! Hopefully that fix propagates to the original PR. Let me know if there are any other issues and/or what else can I do to help.

berrnd (approved):

berrnd:
> Let me know if there are any other issues and/or what else can I do to help.

Everything on this project is done publicly, so just have a look at all the [open Bugs](https://github.com/grocy/grocy/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and [Feature Requests](https://github.com/grocy/grocy/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) and feel free to do whatever you want.

Issue #1980, opened 2022-08-26
Bug: Automatic add to shopping list wrong unit


tieu1991:
I use the automatic add to shopping list when stock is below minimum stock quantity. The problem is when that happen it add the minimum stock quantity to the shopping list instead of the default buy quantity. Here is an exemple of what happen.

Product : Eggs
Default purchase unit : pack
Default stock unit: piece
Factor purchase to stock : 12
Minimum stock unit : 4

When the number of eggs in the stock fall below 4, 4 eggs is added to the shopping list. I would expect 1 pack of eggs to be added to the shopping list. 

Grocy version 3.3.1

berrnd:
Not a bug.

Duplicate of #1355 (and #1615, and #1694, and #1837 - please check existing request before opening new ones).

Issue #1741, opened 2022-01-10
Bug: Default quantity added to shopping list using REST API


andreheuer:
I have created a product that has a different purchase QU (e.g. bottle) than default QU (e.g. ml, and 1 bottle = 1000ml). When adding this item to the shopping list using the UI, the purchase QU is selected by default (bottle).

However, the REST api does not support to select a QU during adding items to a shopping list using /stock/shoppinglist/add-product API. The expectation is (from my side) that the purchase QU (bottle) will be used by adding an item to a shopping list (as in the UI), but the default QU (ml) is used. So, if I am adding 1 item of the product to the shopping list, it adds 1ml to the shopping list instead of 1 bottle. From my point of view, it should either use the purchase QU as default or that the QU can be also selected by the REST api call.

This applies to the demo instance and is also tested with 3.1.3.

Happy to discuss.

berrnd:
Just to clarify generally: On any API endpoint any amount is related to the products QU stock.

For UI things a duplicate of #1355.

> the REST api does not support to select a QU during adding items to a shopping list using /stock/shoppinglist/add-product API

Not true. The UI uses the API for everything, so the best and most simple way to check how things are done is to utilize your Browsers developer tools and inspect the requests - sample:
```
curl -X 'POST' \
  'https://demo-prerelease.grocy.info/api/stock/shoppinglist/add-product' \
  -H 'accept: */*' \
  -H 'Content-Type: application/json' \
  -d '{
  "product_id": 27,
  "list_id": 1,
  "product_amount": 5,
  "qu_id": 12
}'
```

andreheuer:
I have checked the documentation of the API which then seems not to be up to date:
![image](https://user-images.githubusercontent.com/10393020/148740216-aa662f8b-a4a3-4a94-a886-594d17a8112d.png)
There is no qu_id parameter mentioned.

Issue #1694, opened 2021-11-23
Feature Request: Enhancement of shopping list


fideco1981:
Just started using Grocy, and I love it!

One feature request is this:

When an item is below minimum stock, when you add it to the shopping list, it will only add amount missing from minimum stock.
Eg. you have 4 eggs minumum stock, you use two eggs, two eggs will be added to the shopping list. 

I'd like to be able to add "phurcase amount" to the products, so that I can add eg. 6 eggs automatic when below minimum.

berrnd:
Duplicate of #1355.

Issue #1385, opened 2021-02-25
Bug: Adding product to shopping list via stock overview uses wrong units


wrybind:
When adding products to the shopping list via the stock overview the system uses a different unit than the user selected.

1.	Create product baking powder with “quantity unit stock” set to piece and “Default quantity unit purchase” to pack with a “Factor purchase to stock quantity unit” of 20.
2.	Go to stock overview and click “Add to shopping list” for baking powder
3.	Add 1 pack (pack is correctly displayed as the default unit)
4.	Result: 20 pieces are added to shopping list
 
I chose that I want to add the unit “pack” but the system uses “pieces” instead. 
 
When a pack of baking powder is already on the shopping list, adding another one as described will lead to “two packs” which is correct. 
 
The problem does not appear when adding products to the shopping list from the shopping list view.
 
The “add products that are below defined min. stock amount” way of adding to the shopping list has the same behavior, so that solving this issue might help enhancement #1355 too.

PR #2416, opened 2024-01-01
convert buying amount to set qu_id_purchase

viktor-c:
## Prerequisite
  * sqlite uihelper_shopping_list -> already saves the qu_id for added products
  * but AddMissingProductsToShoppingList from StockService does not use it

## Path to add products to shopping list from shoppinglistView:
* Add products that are below defined min stock amount
and
* Product is __not__ on the shopping list
  * missing amount will be converted to qu_id_purchase
  * ceil to next integer
  * add missing amount _equivalent_ to qu_id_purchase
    * equivalent because the database holds the amounts in qu_id_stock
* Product __is__ on the shopping list
  * if we are missing an amount greather than what we already have on the list
  * check to see how many we need to achieve minimum stock
  * and calculate like above the _equivalent_ amount in qu_id_purchase needed to buy

## Examples
* Example 1, Eggs
  * Stock 35 Piece. Minimum Stock 30 Piece, 1 Pack = 10 Piece
  * consume 31, remaining 4, amount_missing=26
  * -> add to shopping list _equivalent_ amount of 30
  * will be shown as __3 Packs__
* Example 2, Eggs
  * Stock 35 Piece. Minimum Stock 30 Piece, 1 Pack = 10 Piece
  * consume 31, remaining 4, amount_missing 26, already on the shopping list 5 Piece
  * -> add to shopping list 26 ↑ 30 ≡ 3 Packs
* Example 3, Eggs
  * Stock 35 Piece. Minimum Stock 46 Piece, 1 Pack = 10 Piece
  * consume 31, remaining 4, amount_missing 42, already on the shopping list 3 Packs
  * -> add to shopping list 42 ↑ 50 ≡ 5 Packs

Solves: #1355 
@berrnd I can understand it does not seem to bother you to buy a pack of eggs when you see you need 4 or 5, but many users in the last 3 years seem to stumble upon this (#1385, #1694, #1741, #1980 and so on).

Other than that I wish all the contributors a Happy New Year and stay healthy!

### DIFF
--- a/services/StockService.php
+++ b/services/StockService.php
@@ -32,14 +32,27 @@ class StockService extends BaseService
 
 			$alreadyExistingEntry = $this->getDatabase()->shopping_list()->where('product_id', $missingProduct->id)->fetch();
 
+			$equivalentAmountToAdd = $amountToAdd;
+			if ($product->qu_id_stock != $product->qu_id_purchase){
+				$productQuConversion = $this->getDatabase()->cache__quantity_unit_conversions_resolved()->where('product_id = :1 AND from_qu_id = :2 AND to_qu_id = :3', $product->id, $product->qu_id_stock, $product->qu_id_purchase)->fetch();
+				if ($productQuConversion){
+					$factor = $productQuConversion->factor;
+					/// we need to find the next integer from qu_id_stock(amount_missing) * factor
+					// which is ceil(qu_id_stock(amount_missing) * factor) * qu_id_purchase
+					// but the shopping list always has the number needed to buy as stock unit, so convert back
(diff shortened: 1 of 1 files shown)


berrnd:
By using `ceil` you always loose information and that can never be a solution, just like in #2417. And also just like there, a simple "UI only solution" to solve this problem would be more than enough - it's only about displaying amounts, the math behind is already correct.

Don't get me wrong, but this here looks mostly like a copy & paste job rather than really understanding what you did - which is more than fine, but reviewing something like that wastes hours on my side for nothing at then end...

Issue #2091, opened 2023-01-02
Sensible order quantities on shopping list?


danq8:
Hi.
Really like Grocy, but am having real issues when trying to get realistic order quantities on a shopping list.

Example 1:
(Using a private pre-release demo)
Steps to reproduce:
1) Remove all milk from stock.
2) Add 21 servings of Chocolate sauce.
3) Add missing items to shopping list.
Result: 4.2 bottles of milk on shopping list.
Is there any way to get the sensible result of 5 bottles?

Example 2:
(Found on my own instance running version 3.3.2 of Grocy on home assistant)
I have a recipe that needs two cloves of garlic.
The only quantity I can buy in my local store is a bag with 3 Bulbs of garlic.
So I created a unit of measure of 1 garlic clove,
another unit of measure as 1 garlic bulb with approximately 8 cloves in a bulb.
and finally a pack of 3 Bulbs equivalent to the product.
I created a product as 3 Bulbs of garlic.
Then added a recipe that needed 2 cloves of garlic to the menu planner, and clicked the button to add products not in stock to the shopping list.
I've tried every variation I can think of, but never get the right result.
From example 1, I guess there's work that needs to be done if Grocy is meant to order quantities that are available in a store from the menu planner? Or I could just not be getting the config right, I can't find any help to get it working the way I would expect it to work..
Thanks for taking a look.

Cheers

Dan.

berrnd:
Duplicate of #1355 (and and #1615, and #1694, and #1980) - please check existing requests before opening new ones.

&nbsp;
> Result: 4.2 bottles of milk on shopping list.
> Is there any way to get the sensible result of 5 bottles?

Personal unpopular opinion: grocy does the math (right) and picking 5 bottles when there are 4.2 on my shopping list is not that hard practically.

PR #2417, opened 2024-01-01
convert and ceil to next integer the amount to order for the recipe

viktor-c:
This PR means to address this cosmetic issue #2091
Now instead of showing 4,2 bottles of Milk, it will show 5 bottles in the shopping list.

### DIFF
--- a/services/RecipesService.php
+++ b/services/RecipesService.php
@@ -41,7 +41,7 @@ class RecipesService extends BaseService
 					$conversion = $this->getDatabase()->cache__quantity_unit_conversions_resolved()->where('product_id = :1 AND from_qu_id = :2 AND to_qu_id = :3', $recipePosition->product_id, $recipePosition->qu_id, $product->qu_id_stock)->fetch();
 					if ($conversion != null)
 					{
-						$toOrderAmount = $toOrderAmount * $conversion->factor;
+						$toOrderAmount = ceil($toOrderAmount * $conversion->factor);
 					}
 				}
 


viktor-c:
I see there will be a problem for meat for example:
Should buy 1,1 kg meat, this will put 2 kg on the shopping list.
There are some quantity units which cannot be divided (e.g. bottles) and ones that should be left fractional. 
What do you think would be a better place to introduce an option to specify this: 
  - when defining quantity units, have a boolean option: fractional
or 
  - as a property of the product: disable rounding in shopping list
when checked it will allow buying a fraction of this unit and disable ceiling in the shopping list

berrnd:
This doesn't address a cosmetic issue, this completely ignores / removes fractional amounts when adding missing ingredients to a shopping list. Maybe it fits you, but this can never be called a solution for the underlying problem / feature request. A _cosmetic_ thing would be doing that only for the UI (which is already possible by utilizing [custom JS](https://github.com/grocy/grocy#adding-your-own-css-or-js-without-to-have-to-modify-the-application-itself) if you want that).

viktor-c:
> A _cosmetic_ thing would be doing that only for the UI (which is already possible by utilizing [custom JS](https://github.com/grocy/grocy#adding-your-own-css-or-js-without-to-have-to-modify-the-application-itself) if you want that).
Is an UI only solution of displaying the amount as above worth looking into or is it doomed to be rejected?

berrnd:
> Is an UI only solution of displaying the amount as above worth looking into or is it doomed to be rejected?

Nothing was ever and will ever be rejected when it generally works and doesn't only cover one (your) specific use case. This here was rejected because it removes information (fractional amounts), as already pointed out above.

viktor-c:
> > Is an UI only solution of displaying the amount as above worth looking into or is it doomed to be rejected?
> 
> Nothing was ever and will ever be rejected when it generally works and doesn't only cover one (your) specific use case. This here was rejected because it removes information (fractional amounts), as already pointed out above.

I see in your mentioned issues a lot of CustomJs code. Maybe somehow off topic, can you put together a set of "snippets" which could then be activated from the UI. Say you have one for ceiling amount in the shopping list, you have one to enable a checkbox as default and so on, like gathering some of the useful ones. By describing this in the _Readme.md_ maybe you can avoid so many opened issue.

berrnd:
> can you put together a set of "snippets" which could then be activated from the UI

Of course I technically could, but if one can't open a text file and copy & pasting a snippet found somewhere, than custom JS is not what one needs. Beside that, custom JS is _custom_ - it will never be used for a built-in solution. But of course, like for just like anything related to UI, JS is most likely needed for that - but there is not only custom JS in Grocy, but also built-in JS ([ref1](https://github.com/grocy/grocy/tree/master/public/js), [ref2](https://github.com/grocy/grocy/tree/master/public/viewjs)).

&nbsp;
> By describing this in the Readme.md maybe you can avoid so many opened issue.

Yes, I also thought one time that a README prevents the same over and over again (frustrating) reports. Life thought me something different, especially in 2023 - too many people are unable to read READMEs in reality.

PR #2429, opened 2024-01-08
Change Openapi descriptions for easier use of auto-generated api client

jan-karwowski:
Hi!

Thanks for making this awesome household management tool!

I am currently writing some helper programs which interact with grocy api. I am doing this in Java and generating the client api bindings with swagger.  During that I ran into some problems regarding the fact that openapi description does not exactly match the actual api. 

- in some types it is stated that field type is `date-time` even if the field contains only a `date` -- this PR changes format form `date-time` do `date`,
- the `ProductDetailsResponse` type contains `product_barcodes` field. Openapi says that it contains a `ProductBarcode` item while actual type is an array of such items.
- the `spoiled` field in stock entry in api description is declaread as a boolean. Sematically it is a boolean but the values returned are 0 or 1, so it cannot be parsed as a boolean.

Besides that I found using the generated bindings for the generic entity api rather hard to use. Every time I fetch some data, for instance a product or quantity unit, etc. I need to check the type of a returned object and cast it into the requested entity type (this is because java is a statically typed language). A part of this PR (commits 641be7ee30d4f9d8547e231de39367e5fd2442a0 and 6599da09e7c37c827651c3be94a9f6df8b485c6c) changes the description of the generic entity api from a single `/objects/{entity}` endpoint that has an entity type parameter to many endpoints where there is separate api endpoint for `/object/XXX` for each object.  This way each of new endpoints has a single possible return type, so no casting is needed.

This PR changes **only openapi description**. No changes to actual API are made so it should not break any existing programs using the API. (In the PR I tried to avoid API breaks at all cost even if IMO it is better to change some things in API responses).

The changes in commits 641be7ee30d4f9d8547e231de39367e5fd2442a0 and 6599da09e7c37c827651c3be94a9f6df8b485c6c will change the auto-generated bindings for anyone who uses such and therefore it may break things. If this is a problem I can submit another PR excluding those two commits.

This PR does not solve all the problems I had with auto-generated bindings. I had one more: `row_created_timestamp` is not formatted in accordance to JSON schema for `date-time` as the JSON Schema spec requires that `date-time` contains a timezone. Currently I've done a workaround on a client side as changing it in grocy may break the programs that already use the API.

### DIFF
--- a/controllers/OpenApiController.php
+++ b/controllers/OpenApiController.php
@@ -100,6 +100,68 @@ class OpenApiController extends BaseApiController
 				array_push($spec->components->schemas->ExposedEntity_NotIncludingNotListable->enum, $value);
 			}
 		}
+
+		$entity_types = array(
+			"products" => "Product",
+			"chores"  => "Chore",
+			"product_barcodes" => "ProductBarcode",
+			"batteries" => "Battery",
+			"locations" => "Location",
+			"quantity_units" => "QuantityUnit",
+			//			"quantity_unit_conversions" => ,
+			"shopping_list" => "ShoppingListItem",
+			// "shopping_lists" => ,
+			"shopping_locations" => "ShoppingLocation",
+			//			"recipes" => ,
+			// "recipes_pos",
+			// "recipes_nestings",
+			"tasks" => "Task",
+			"task_categories" => "TaskCategory",
+			//			"product_groups",
+			//"equipment",
+			"api_keys" => "ApiKey",
+			//"userfields",
+			//"userentities",
+			//"userobjects",
+			//"meal_plan",
+			"stock_log" => "StockLogEntry",
+			"stock" => "StockEntry",
+			///			"stock_current_locations" => ,
+			"chores_log" => "ChoreLogEntry"
+			//"meal_plan_sections",
+			//"products_last_purchased",
+			//"products_average_price",
+			//"quantity_unit_conversions_resolved",
+			//"recipes_pos_resolved"
+		);
+		// non-generic entity api
+		foreach(array_unique(array_merge($spec->components->schemas->ExposedEntity_NotIncludingNotListable->enum,
+										 $spec->components->schemas->ExposedEntity_NotIncludingNotEditable->enum)) as $value) {
+			$listObjectsPath = unserialize(serialize(clone $spec->paths->{"/objects/{entity}"}));
+
+			$type = null;
+			if(array_key_exists((string)$value, $entity_types)) {
+				$type = (object) array('$ref' => "#/components/schemas/" . $entity_types[$value]);
+			} else {
+				$type = (object) array("type" => "object");
+			}
+			$listObjectsPath->get->responses->{"200"}->content->
+													   {"application/json"}->schema->items = $type;
+			$listObjectsPath->post->requestBody->content->
+													   {"application/json"}->schema = $type;
+
+			if(!in_array($value, $spec->components->schemas->ExposedEntity_NotIncludingNotListable->enum)) {
+				unset($listObjectsPath->get);
+			}
+			if(!in_array($value, $spec->components->schemas->ExposedEntity_NotIncludingNotEditable->enum)) {
+				unset($listObjectsPath->post);
+			}
+
+			// TODO: update summary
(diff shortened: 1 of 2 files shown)


berrnd:
It's nowhere stated that Grocy's OpenAPI specification is meant for feeding automatic API client generation tools.

It's defined purpose is
- serving the [/api](https://demo-prerelease.grocy.info/api) page (Swagger UI)
- validation for various API related things ([example code ref](https://github.com/grocy/grocy/blob/99f448dd6476dda1ea71037c8af1daa5254c64af/controllers/GenericEntityApiController.php#L298-L301))

and nothing else.

So needs thorough testing for the later point and a detailed review of [/api](https://demo-prerelease.grocy.info/api) if the documentation there (which of course can be always improved) still covers everything at least as before, especially for generic endpoints returning different schemas on purpose. Breaking API changes related to any of that won't be accepted in any case. Thanks anyway.

berrnd (changes requested):
The `data-time` / `date` fixes are ok.

The ones related `ProductDetailsResponse` / `product_barcodes` too.

&nbsp;
> the `spoiled` field in stock entry in api description is declaread as a boolean. Sematically it is a boolean but the values returned are 0 or 1, so it cannot be parsed as a boolean.

It's the same for each and every boolean field in Grocy's API, not only this one. So if, then change all and also add an explanation since the next question would be what other numbers mean then. Even that it's theoretically correct, practically any [`truthy`](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) value is also just fine, even if maybe auto-generating tools can't handle that out of the box.

&nbsp;
> A part of this PR (commits 641be7e and 6599da0) changes the description of the generic entity api from a single /objects/{entity} endpoint that has an entity type parameter to many endpoint

This won't be merged. Makes the /api page lengthy as hell, doesn't improve its readability and is then also inconsistent with other endpoints (e.g. `GET / PUT /userfields/{entity}/{objectId}`.

jan-karwowski:
> It's nowhere stated that Grocy's OpenAPI specification is meant for feeding automatic API client generation tools.
> 
> It's defined purpose is
> 
>     * serving the [/api](https://demo-prerelease.grocy.info/api) page (Swagger UI)
> 
>     * validation for various API related things ([example code ref](https://github.com/grocy/grocy/blob/99f448dd6476dda1ea71037c8af1daa5254c64af/controllers/GenericEntityApiController.php#L298-L301))

OK. My changes were focused at code generation stuff, so probably you should close this PR without merging.
What about such solution:  I will make another PR containing **only** commits regarding date/date-time and `product_barcodes`, so it can be easily merged. Are you ok with that?




> > the `spoiled` field in stock entry in api description is declaread as a boolean. Sematically it is a boolean but the values returned are 0 or 1, so it cannot be parsed as a boolean.
> 
> It's the same for each and every boolean field in Grocy's API, not only this one. So if, then change all and also add an explanation since the next question would be what other numbers mean then. Even that it's theoretically correct, practically any [`truthy`](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) value is also just fine, even if maybe auto-generating tools can't handle that out of the box.

The PR touched only things that I was using, to only this field. I agree that there are more fields with this problem. The json schema spec used by openapi spec explicitly says that `boolean` type can contain [either `true` or `false` and nothing else](https://json-schema.org/understanding-json-schema/reference/boolean), even if in JS there are many values that can be evaluated to true. Nevertheless if a code generation is not the purpose of the api description, I don't think that my change here is relevant.

berrnd:
> I will make another PR containing **only** commits regarding date/date-time and `product_barcodes`, so it can be easily merged. Are you ok with that?

Sure.

&nbsp;
> The json schema spec used by openapi spec explicitly says that `boolean` type can contain [either `true` or `false` and nothing else](https://json-schema.org/understanding-json-schema/reference/boolean), even if in JS there are many values that can be evaluated to true.

I already mentioned theory and practicability. Believe it or not, I also use statically typed languages like Java or C# kind of daily and it's by far not impossible there too, just like probably in any code language on this planet. Just not out of the box by the code generation tool of choice, sometimes you need to do something on your own I guess.

jan-karwowski:
Created  #2433 containing only the accepted changes. Closing this one.

PR #2432, opened 2024-01-09
Fix blank/missing items in spending report

### LINKED ISSUE #2431: Bug: Spending Report Blank/Missing Items When Product Groups Not Used
When viewing the spending report, if product groups are not linked to every product, the Spending Report can either be a) completely blank or b) missing items.

This issue is impactful because users may be relying on this feature for budgeting and to track their spending. In the case of the report being completely missing, this is at least obvious, but in the case of where not every product has a product group, it can cause the spending report to underreport which may not be noticeable to the end user.

The root cause of this is that we are using a simple inner join when joining to the product group table. This is an issue since product groups are an optional field and therefore products without a product field are inadvertently filtered out. We should be using a left join instead.

I have patched this and tested it. You will see a PR shortly.

----
Steps to Reproduce Issue:

1. Go to demo instance and view spending report.
2. Note a product currently in the table/in the pie chart.
3. Go back to the stock overview and remove the product group from one of those entries.
4. Go back to spending report and note that the product is now missing from the report. If you repeated this for all product instances (or just used a new prod install without linking product groups), you will be left with a blank report.

Torqu3Wr3nch:
Use left (outer) join when connecting product to optional product group. Resolves #2431.

### DIFF
--- a/controllers/StockReportsController.php
+++ b/controllers/StockReportsController.php
@@ -44,7 +44,7 @@ class StockReportsController extends BaseController
 			FROM products_price_history pph
 			JOIN products p
 				ON pph.product_id = p.id
-			JOIN product_groups pg
+			LEFT JOIN product_groups pg
 				ON p.product_group_id = pg.id
 			WHERE $where
 			GROUP BY p.id, p.name, pg.id, pg.name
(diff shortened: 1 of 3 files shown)


berrnd (approved):
👍

PR #2433, opened 2024-01-09
Fix data types in opeanapi description

jan-karwowski:
A subset of changes proposed in #2429

- the `ProductDetailsResponse` type contains product_barcodes field. Openapi says that it contains a `ProductBarcode` item while actual type is an array of such items.
- change several occurences of `date-time` to `date` in cases where only date is returned


berrnd (approved):
👍

Issue #2458, opened 2024-02-05
Bug: File Userfields: Uploading big files (= when the upload takes longer than a couple of seconds) doesn't work


PedroCPereiraDev:
Hello,

I created a UserField called Software for my Equipments, being this field a FILE type.
My intent is to save ZIP containing firmwares or other related small pieces of software.

Upon uploading the file successfuly, it shows as saved but cannot be retrieved or downloaded.

Attached I show prints of how it appears.

My expectation was that obviouly would not be displayed but that pressing the download button would work.

Other related bug is that I tried with other file types, and other file types work (I tried with a STL file) but then another strange interaction happaned. If for example you upload an STL file, once you go to the Equipments list, and click the TAB of the UserField File you created it automaticly downloads the STL file. I believe it should only download if you click the download button.
Attached I also send a photo, except for the auto download the UI when uploading an STL seems good to me.

I tried in a HASS addon instance, a UNRAID docker instance, and both directly via IP or via a NGinx Proxy to rule out network.
I tried also the pre-realease demo and the same behauvior happened.

I state this BUG as related to ZIP files, but I suspect it 

what_happens_when_downloading_using_the_button:
![what_happens_when_downloading_using_the_buton](https://github.com/grocy/grocy/assets/11192580/3cd84849-b137-43ad-95c2-5075eebb275c)

how_it_appear_in_the_list:
![how_it_appear_in_the_list](https://github.com/grocy/grocy/assets/11192580/b2017a17-74f4-48e5-a9b8-a5529d1c1697)

how_it_appears_in_edit_mode_after_saving:
![how_it_appears_in_edit_mode_after_saving](https://github.com/grocy/grocy/assets/11192580/d9539b46-f0bf-4b0e-856a-2b1c55e46707)

what_shows_when_the_file_is_an_STL:
![what_shows_when_the_file_is_an_STL](https://github.com/grocy/grocy/assets/11192580/4b0c6ce6-b12c-4f6d-81ad-5b19a2fb97dd)

PedroCPereiraDev:
After submitting the report I had an idea that maybe this bug (the ZIP part, not the auto-download) is not related to the file type itself but with the size of the file.

The ZIP I was using was 41MB of size, I just now tried a ZIP file of just 102KB and it worked (althought it still auto-downloads).

So the issue here is size related. I believe that even the PDF files normally uploaded if big enought to be 41MB would have problems.

PedroCPereiraDev:
My assumption is correct.

[...]

PR #2477, opened 2024-02-23
Stream uploads to file

bbx0:
Currently, file uploads are buffered in memory. For very large uploads this can cause a memory exhaustion and I noticed the following error:

```text
NOTICE: PHP message: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 98566176 bytes) in /var/www/grocy/packages/slim/psr7/src/Stream.php on line 357
```

This pull request changes the logic to stream an upload directly into the target file (using a chunk size of 32 KB). Now the maximum file size is only controlled by the likes of `client_max_body_size` and no longer affected by the PHP `memory_limit`.

Please just let me know if anything shall be revised, or you would like to receive the pull request in a different format.

### DIFF
--- a/controllers/FilesApiController.php
+++ b/controllers/FilesApiController.php
@@ -109,9 +109,26 @@ class FilesApiController extends BaseApiController
 			}
 
 			$fileName = $this->checkFileName($args['fileName']);
-			$data = $request->getBody()->getContents();
 
-			file_put_contents($this->getFilesService()->GetFilePath($args['group'], $fileName), $data);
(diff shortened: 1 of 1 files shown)


berrnd:
Thanks, but just a reality check: Maybe a memory limit of ~128 MB is not from 2024, but more from 1995 or something like that.

berrnd (approved):
This only solves partly the known problems about big file upload in Grocy (also ref #2458).

A real solution would be using client side chunking.

bbx0:
> Thanks, but just a reality check: Maybe a memory limit of ~128 MB is not from 2024, but more from 1995 or something like that.

Very true :) It's just the default value. Thanks for the quick follow-up and all your work!

Issue #1626, opened 2021-09-26
Feature Request: OAuth Authentication


quqkuk:
The issue is self-explanatory. I'd like to log into grocy using my Nextcloud credentials, preferably by using OAuth (which is open and can be used with other platforms).
I could work on it, although I can't guarantee that I'll have something done quickly (mainly because my PHP knowledge is rusty).
Thank you for your feedback in advance!

berrnd:
Authentication is done by a configurable middleware, see [data/config.php](https://github.com/grocy/grocy/blob/04d826943cd39720a4591d9879b1f4f87658ec0b/config-dist.php#L72-L74).

Some others use this for example with [Authelia](https://www.authelia.com/) (and `ReverseProxyAuthMiddleware` in grocy), maybe have a look at the [subreddit](https://www.reddit.com/r/grocy).

About the Nextcloud integration: Duplicate of #406.

PR #2489, opened 2024-03-04
Simple OAuthMiddleware 

bboehmke:
This PR implements a simple OAuthMiddleware. I had a couple of issue with the ReverseProxyAuthMiddleware in combination with the API keys. So I tried to implement a minimal OAuth variation instead.

I tested the implementation with [authentik](https://goauthentik.io/) but it should work with any OAuth2 provider.

I know that there already was a discussion about OAuth in #1626.
So if you don't like to add this to grocy, it would be great to have at least 1616b41ea99e9dd77cd53821d7f297b45e79ff82 (I will create a separate PR if needed).
Otherwise the required redirect is not working in the OAuthMiddleware.

PS: I have not done any PHP programming since multiple years. Let me know if anything should be changed.

### DIFF
--- a/config-dist.php
+++ b/config-dist.php
@@ -96,6 +96,15 @@ Setting('LDAP_BIND_PW', ''); // Password for the above account
 Setting('LDAP_USER_FILTER', ''); // Example value "(OU=grocy_users)"
 Setting('LDAP_UID_ATTR', ''); // Windows AD: "sAMAccountName", OpenLDAP: "uid", GLAuth: "cn"
 
+// Options when using OAuthMiddleware
+Setting('OAUTH_CLIENT_ID', '');
+Setting('OAUTH_CLIENT_SECRET', '');
+Setting('OAUTH_SCOPES', 'openid profile');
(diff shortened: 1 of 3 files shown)


berrnd:
> So if you don't like to add this to grocy

That's the case - simply share it wherever you want and maintain it over the long run yourself (that's also the simple point why I don't accept that - I can't maintain stuff I'm personally not interested in and so far it never happened that random drive contributors came back when after one or two years something doesn't work with what was contributed).

&nbsp;
> it would be great to have at least https://github.com/grocy/grocy/commit/1616b41ea99e9dd77cd53821d7f297b45e79ff82 (I will create a separate PR if needed)

Not needed, simply use the whole power of PHP - doing an immediate redirect in plain is as simple as:
```php
header('Location: https://example.com');
exit();
```

Thanks anyways.

bboehmke:
no issue

> header('Location: https://example.com');
> exit();

thanks for the hint. I will rework it and maintain it on my own

PR #2491, opened 2024-03-10
Add link for opening Barcode Buddy

TheDodger:
As I struggled getting a customisation by custom_js.html done, I thought maybe such a feature could be of interest for more than one person with a tablet in single app mode.

I tried to integrate it as cleanly as I could. While I was at it, I aligned the menu items by changing icons to fixed width.

Please don't yell at me. :)

### DIFF
--- a/changelog/75_UNRELEASED_xxxx-xx-xx.md
+++ b/changelog/75_UNRELEASED_xxxx-xx-xx.md
@@ -58,7 +58,7 @@
 
 - Optimized that file uploads are not completely buffered in memory before writing them to disk (thanks @bbx0)
 - Fixed again that some dialogs were not properly (too small) displayed (this time mostly affecting Chrome/Edge)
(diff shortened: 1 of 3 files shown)


berrnd:
None of any of the community contributed apps/tools will ever be integrated into Grocy directly.

If we start that, tomorrow will bring 100 additional requests integrating links for whatever third party web app one can think of. Exactly for this level of customization [custom JS](https://github.com/grocy/grocy?tab=readme-ov-file#adding-your-own-css-or-js-without-to-have-to-modify-the-application-itself) was invented and it's as easy as cooking some rice to use that to add a simple link to the sidebar - if not for you, then maybe simply ask the internet (the [r/grocy subreddit](https://www.reddit.com/r/grocy) would be an appropriate place for that).

&nbsp;
> While I was at it, I aligned the menu items by changing icons to fixed width.

Mixing completely unrelated things is always a good thing, so I'm simply going to ignore that here.

PR #2492, opened 2024-03-10
Align settings (wrench) menu entries

TheDodger:
Untangled it from the other pull request.


TheDodger:
Would it be okay if I also include the same changes to the user menu here or would you like an additional PR for that?

TheDodger:
Thumbs up for okay or thumbs down for additional PR would suffice. :)

berrnd (approved):

berrnd:
> Would it be okay if I also include the same changes to the user menu

Of course, doesn't make any sense to only care about some icons when having a look at that - but superseded by https://github.com/grocy/grocy/pull/2492/commits/d9db158c2d5138f1d2b55dc6606b93673fd04112.

PR #2501, opened 2024-03-17
Also copy treat_opened_as_out_of_stock when copying a product

TheDodger:
Copy pasted a bit, again. Couldn't find a ticket for that, so no reference.


berrnd (approved):
👍

PR #2502, opened 2024-03-17
Add battery description column to batteries overview

TheDodger:
Hi Bernd,

I could either open a ticket, asking if you left this out intentionally, or simply take it on and create a pull request asking the exact same question.

Btw. I can't really decide if this column does make any sense at this place, as it could take up a vast amount of space, at least in my case. So maybe it would make sense if it's hidden by default.

Thanks in advance for your opinion.
Manu


berrnd:
> if you left this out intentionally

Yes.

This breaks consistency with any other overview pages. Also adding only the HTML part isn't enough (this e.g. completely breaks the table when the user has reordered or hidden columns before). Enforcing a new column to be always visible just because you want it isn't the way to go, most of the time adding additional minor stuff hidden by default is what's expected.

Please do more extensive testing beforehand, not everything can be solved by blindly copy & pasting stuff. If reviewing / completing a pull request takes for me longer than doing the corresponding task myself at the end, consider adding a regular Feature Request instead - I hate wasting time and even writing this response here took longer than properly implementing this new column.

&nbsp;
> Btw. I can't really decide if this column does make any sense at this place, as it could take up a vast amount of space, at least in my case. So maybe it would make sense if it's hidden by default.

Obviously you've noted that yourself, so why even let me review this state, it's nothing than wasting time - great. Beside that, columns in any table can be reordered by the user (if they're added properly).

PR #2522, opened 2024-04-15
Added alternative quick actions

Logianer:
## What?
image that shows the stockoverview page with the alternative design enabled

- Enable 'Use alternative quick actions' under `/stocksettings`
- The 3 default Quick Actions will be replaced by two Quick actions that each open a dialog (The Consume-dialog and the transfer dialog)

## How?
- Added a "Quick Action Component" `views/components/quickactions.blade.php` that handles the buttons in the first table column (excluding the dropdown)
- New UserSetting svalue `use_alternative_actions_on_stock_overview_page` that controls if the alternative design is enabled.

## Why?
It bothered me, that I always had to go through the dropdown menu to get access to the "advanced" settings for consuming, so I added this feature to enable this view.

### DIFF
--- a/localization/de/strings.po
+++ b/localization/de/strings.po
@@ -2802,3 +2802,6 @@ msgstr "Für den angegebenen Barcode wurde nichts gefunden"
 
 msgid "Configure colors"
 msgstr "Farben konfigurieren"
+
+msgid "Use alternative quick actions"
+msgstr "Alternative Schnellaktionen benutzen"
--- a/public/viewjs/stocksettings.js
+++ b/public/viewjs/stocksettings.js
@@ -51,3 +51,8 @@ $("#stock_default_consume_amount_use_quick_consume_amount").on("click", function
 		$("#stock_default_consume_amount").removeAttr("disabled");
 	}
 });
+
+if (BoolVal(Grocy.UserSettings.use_alternative_actions_on_stock_overview_page))
+{
+	$("#use_alternative_actions_on_stock_overview_page").prop("checked", true);
+}
--- a/views/components/quickactions.blade.php
+++ b/views/components/quickactions.blade.php
@@ -1,0 +1,50 @@
+@if(boolval($userSettings['use_alternative_actions_on_stock_overview_page']))
+<a id="product-{{ $currentStockEntry->product_id }}-consume-button"
+	class="permission-STOCK_CONSUME btn btn-success btn-sm product-consume-button show-as-dialog-link @if($currentStockEntry->amount_aggregated == 0) disabled @endif"
+	href="{{ $U('/consume?embedded&product=' . $currentStockEntry->product_id ) }}">
+	<i class="fa-solid fa-utensils"></i> {{ $__t('Consume') }}
+</a>
+<a id="product-{{ $currentStockEntry->product_id }}-consume-button"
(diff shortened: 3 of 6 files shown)


berrnd:
That's fine that you like it more like that and we've discussed those buttons here maybe 10 times per year or so. However, if implementing each and everyone's preferred way of having those buttons, we'll end up in having two trillions of "alternative quick actions" or something like that. All about that depends on your very personal way of using Grocy and since Grocy doesn't enforce a specific workflow, there are endless possibilities what can fit you, but most likely not everyone else at just the same time.

[Custom JS](https://github.com/grocy/grocy?tab=readme-ov-file#adding-your-own-css-or-js-without-to-have-to-modify-the-application-itself) is the way to customize such details without having to maintain your complete own fork, but of course that's also doable.

Thanks anyways.

localization/de/strings.po
```diff
@@ -2802,3 +2802,6 @@ msgstr "Für den angegebenen Barcode wurde nichts gefunden"
 
 msgid "Configure colors"
 msgstr "Farben konfigurieren"
+
+msgid "Use alternative quick actions"
```
berrnd, line 2806: Translations are managed via Transifex ([README ref](https://github.com/grocy/grocy?tab=readme-ov-file#localization)), so the only file to be edited here is [`strings.pot`](https://github.com/grocy/grocy/blob/master/localization/strings.pot).

public/viewjs/stocksettings.js
```diff
@@ -51,3 +51,8 @@ $("#stock_default_consume_amount_use_quick_consume_amount").on("click", function
 		$("#stock_default_consume_amount").removeAttr("disabled");
 	}
 });
+
+if (BoolVal(Grocy.UserSettings.use_alternative_actions_on_stock_overview_page))
+{
+	$("#use_alternative_actions_on_stock_overview_page").prop("checked", true);
```
berrnd, line 57: New user settings also have to be added to [`config-dist.php`](https://github.com/grocy/grocy/blob/master/config-dist.php), otherwise a "Warning: Undefined array key" will be thrown at many places (please do extensive tests before submitting a PR) unless each user changed this settings once (= each and everyone, existing updated instances and new ones).

views/stocksettings.blade.php
```diff
@@ -207,8 +207,21 @@ class="form-check-input custom-control-input user-setting-control"
 
 		@endif
 
+		<div class="form-group">
```
berrnd, line 210: Since this all affects only the /stockoverview page, the more appropriate group would be "Stock overview" and not "Common". Also a little help icon explaining shortly what this is about wouldn't hurt.

views/components/quickactions.blade.php
```diff
@@ -0,0 +1,50 @@
+@if(boolval($userSettings['use_alternative_actions_on_stock_overview_page']))
```
berrnd, line 1: This completely drops the current quick consume/open buttons (and doesn't move them to the dropdown/more menu or something like that). The new buttons also don't have tooltips like the current ones do. Dropping existing functionality, also when it's just for an "alternative", can't be called a solution just like ever.

berrnd (commented):
Even though I've already decided to not merge this, I just gave it a little spin.

PR #2524, opened 2024-04-16
Invalid / Spam / Removed

Sanji78:
Invalid / Spam / Removed


berrnd:
Please don't play games with me. This is simply a newly created account of @StefanoGiu.

We've had enough (private) conversion lately, I'm not reviewing anything of you. Just annoying that you're still harassing me even though I've clearly stated several times that I don't want any further contact with you, wow.

Issue #2186, opened 2023-03-31
Feature Request: Optimize tracking button handling on the chores overview page


karwosts:
I'm using monthly chores for a couple things, but the logic for tracking the chore does not feel right. 

For example I have a monthly chore due on the 1st of every month. It is now March 31, and my next due date is April 1. If I "track chore" today, the day before it is due, the due date does not advance, the only way I can "track" an April chore is to track it during the month of April, otherwise it keeps resetting the next date to April 1. 

I feel like the logic for this is not quite right, and it makes chores that are due in the early days of the month impossible to track early. 

If I had a chore due on the 31st, I could track that chore as complete any time in the weeks leading up to it, and it would advance to the next month, but I can't do that with a chore due on the 1st. 

Some possible ideas for improvement:
 - allow a configurable date in the monthly chore for which is considered to track for the next month, instead of the current month
 - or, if next chore due date is in month X, and it is tracked in month X-1, then move the next due date to month X+1.

karwosts:
I'm thinking of something like a chore to pay a bill monthly that is due on the 1st. It's a chore that needs to be done once a month, before the due date. It does not need to be done _on_ the due date, just something done once a month before it is due. 

Perhaps this is just not something that can be mapped to a chore. Repeating daily / 30 days is no good because it will wander over time and not keep the due date of the 1st.

Right now I have a custom script that is spawning a Task each month to remember to do this. Was hoping I could remove that and just treat it as a recurring chore instead, but there is no existing chore that can handle this type of schedule. 

If the due date was on the 31st it would work fine, since then anytime I track it would be done for the month. Just feels weird that I can't have same functionality for a chore due earlier in the month. 

Anyway, I'm disappointed you see it that way, but thanks anyway for acknowledging and for all your hard work.

berrnd:
> If the due date was on the 31st it would work fine, since then anytime I track it would be done for the month. Just feels weird that I can't have same functionality for a chore due earlier in the month.

Feels weird (for you), works practically just fine - I still just don't see the problem.

And also just another reality check related to the example: I guess paying a monthly bill due on the 1st is naturally most of the time not payed on that 1st, but on a day or two or whatever before to meet the condition on the receiver side. So it makes even more sense to have the corresponding task set to be due on that last day of each month (also ref #816).

For the sake of not disappointing you I've changed the following:
- Changed the handling of the tracking buttons on the chores overview page:
[...]

Issue #2311, opened 2023-08-15
Feature Request: Flip chore execution function


lagerea:
Since 4.0 the chore execution of clicking/pressing the green button in the web view or the chore column(mobile) defaults to chore execution at the scheduled time, and the more options (kebab) menu top option is to track chore execution now. 

From a UX standpoint this functionality should be swapped as the only time a chore should be tracked at the time of scheduled that isn't the present is when it was not tracked but was executed, e.g. I cleaned the dishes, but I forgot to execute the chore tracking until the next day. This is the exception to intended or expected behavior which would be to execute/clear the chore in the present or for it to be completed late, resetting the time required for it to be completed again.

berrnd:
As you already said, this was just flipped for v4.0.0. What should be done from an "UX standpoint" and what is "intended or expected" as always depends on personal preference, not on law. There were a lot requests to change it to like it is now, so sounds more the exception is how you expect it (or as you were used to before v4.0.0, adapting to changes is maybe sometimes the harder part), which is of course ok.

However, this won't be changed back and forth for every release from now one, how it's now makes definitely sense - the green button tracks the corresponding chore keeping the desired schedule.

If that doesn't fit your use case, consider using [custom JS](https://github.com/grocy/grocy#adding-your-own-css-or-js-without-to-have-to-modify-the-application-itself) to change it to behave like however you prefer it.

Here is a snippet (for `data/custom_js.html`) to swap the button / menu item functionality:
```html
<script>
if (Grocy.View == "choresoverview")
{
	$(".track-chore-button").addClass("now");
	$(".dropdown .track-chore-button").removeClass("now");
}
</script>
```

maniackcrudelis:
A simple option would be way easier for everyone to adjust the behavior of this button instead of breaking the previous behavior...
I was myself wondering why my chores were suddenly acting so weird !

berrnd:
Follow-up: #2424 was created to track adding some configuration options to make possible what was mentioned here.

berrnd:
There is now this new settings (chores settings / top right corner settings menu) to swap the "Track next chore schedule" / "Track chore execution now" buttons/menu items on the chores overview page:

![grafik](https://github.com/user-attachments/assets/b8e2846d-c0f5-4f14-9bd1-d709014e8aec)

PR #2537, opened 2024-05-16
feat: configurable behaviour of tracking button on /choresoverview

Worly:
I started using the Chores feature, but got blocked when I noticed that the default green button on /choresoverview tracks an execution of the chore on the next schedule. Following the discussion here https://github.com/grocy/grocy/issues/2186 I see how useful that could be. But unfortunately, it doesn't work as expected for me in 99% of cases.

For example if I have a chore "Clean toilet" setup to be scheduled every 7 days, and I am lazy and do it after 13 days, I don't want it to be scheduled again for tomorrow. The toilet is clean now and needs to be cleaned again in 7 days. Pretty much all of my tasks are like that, so it's cumbersome to open the context menu every time to use the `Track chore execution now` option.

So this PR adds a feature flag to change which of those two behaviours is the default of the green button, and the other one is added as a context menu option.


Worly:
I'm sorry, I failed to find all those others discussions.

> Don't get why it's so often insisted to do everything on [/choresoverview](https://demo-prerelease.grocy.info/choresoverview), maybe simply discover [/choretracking](https://demo-prerelease.grocy.info/choretracking), unleashing endless time selection possibilities.

It's just for ease of use. While doing chores around the house, it's a bit cumbersome to open /choretracking to find the chore which I just completed. Especially since I did the chores because it's overdue on the /choresoverview, so it's already there on the top.

> https://github.com/grocy/grocy/issues/2311#issuecomment-1678780624 even provides an immediate solution to get that by simply utilizing [custom JS](https://github.com/grocy/grocy?tab=readme-ov-file#adding-your-own-css-or-js-without-to-have-to-modify-the-application-itself).

Great! This works for me :D 

I'll close this PR then since it could definitely be implemented better (if it even gets implemented). For example per chore.

PR #2555, opened 2024-07-25
Pull new Finnish translations from Transifex

IWVS:
Pull new Finnish translation files from Transifex now that I have 100% translated them there


berrnd:
Thanks for translating Grocy.

Pulling translations from Transifex is done regularly before a new release is shipped, no need to provide PRs for individual files.

Issue #2560, opened 2024-08-04
Bug: Adding note-only entry to shopping list via API fails


foo-git:
I'm working on a grocy add/consume barcode scanner based on a MicroPython ESP32 chip. While investigating the Grocy API regarding shoppinglist entries, I found a mismatch of the actual API call and the stated behaviour as in #245 by your first comment.

Using the prerelease environment, I tried to add a new entry without any product_id, but it failed. For reference, the JSON request body:
`{
  "list_id": 1,
  "product_amount": 5,
  "note": "This is the note of the shopping list item..."
}`

Grocy response code is 400 with body:
`{
  "error_message": "No product id was supplied"
}`

In the [StockAPIController.php](https://github.com/grocy/grocy/blob/5a2600209f38df1ae78f7689c9e0f45a7e82fea0/controllers/StockApiController.php#L217) I found the thrown exception. In this file, it looks like there's no option for a note-only entry as the exception gets thrown everytime no product_id is set.

As the Android app and the GUI allow entries with emptry product_id it is possible to do this, I'm just not sure where the problem lies exactly. Maybe a clearer documentation on the API call might be helpful.

berrnd:
> As the Android app and the GUI allow entries with emptry product_id it is possible to do this,

Grocy's web frontend uses the exact same API too.

&nbsp;
> Maybe a clearer documentation on the API call might be helpful.

Maybe simply doing the corresponding action on the web frontend, while having a look at the behind-the-scenes at the same time, answers this question itself in the most obvious, easy and fast way I can think of.

Btw: I'm allergic to reporting a simple usage question as a bug, please be so kind and simply ask if something is unclear. Feel more than free to improve any documentation, which is not super crystal clear for you.

&nbsp;
> In the [StockAPIController.php](https://github.com/grocy/grocy/blob/5a2600209f38df1ae78f7689c9e0f45a7e82fea0/controllers/StockApiController.php#L217) I found the thrown exception.

That's not the route used by the "Add item" button on [/shoppinglist](https://demo-prerelease.grocy.info/shoppinglist) (however you came to that wrong conclusion), it's [that one](https://github.com/grocy/grocy/blob/5a2600209f38df1ae78f7689c9e0f45a7e82fea0/controllers/GenericEntityApiController.php#L11-L73) instead.

&nbsp;
So simply do it the right way and everything will work - minimal example:

```
POST /objects/shopping_list

{
  "shopping_list_id": 1,
  "note": "xxynoteonly"
}
```

PR #2561, opened 2024-08-05
Update grocy.openapi.json to include developer tool hint and note-only shoppinglist entry

foo-git:
Thanks for your detailed help in #2560. To improve the documentation, I propose the following changes (hopefully I understand your project structure enough to make them). Please comment if improvements are necessary.

Add general usage/developing hint at the top.
Add hint about adding note-only shoppinglist entries.

### DIFF
--- a/grocy.openapi.json
+++ b/grocy.openapi.json
@@ -2,7 +2,7 @@
 	"openapi": "3.1.0",
 	"info": {
 		"title": "Grocy REST API",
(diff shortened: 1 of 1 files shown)


berrnd:
This is mostly very specific to your particular use case, we can't document each and every single one on [/api](https://demo-prerelease.grocy.info/api). From my real world experience with not only this public Open Source project, people don't read stuff anyways. And all the others can simply search the planet's web, finding #2560 and more references easy and fast.

Thanks anyways.

PR #2564, opened 2024-08-09
Give recipe modal in meal plan space

adaschma:
This is a tiny css addition to give recipe modal in meal plan space for Ingredients/Preparation being side-by-side.

After:

Bildschirmfoto vom 2024-08-09 13-28-30
Bildschirmfoto vom 2024-08-09 13-29-04

Before:
Bildschirmfoto vom 2024-08-09 13-29-43

### DIFF
--- a/public/css/grocy.css
+++ b/public/css/grocy.css
@@ -278,6 +278,13 @@ a:not([href]) {
 	z-index: 99998;
 }
 
+/* Give recipe modal in meal plan space for Ingredients/Preparation side-by-side */
+@media (min-width: 992px) {
+	.modal-xl {
+		max-width: 1100px;
+	}
+}
+
 /* Third party component customizations - DataTables */
 .dataTable td {
 	vertical-align: middle !important;


berrnd:
By default, side by side ingredients/preparation view is already there and it even fits for [/mealplan](https://demo-prerelease.grocy.info/mealplan) on standard 1080p resolution:

grafik

Since Grocy uses Bootstrap (v4.x), their default breakpoints are used for kind of everything when it comes to fit any screen size.

Doc ref: https://getbootstrap.com/docs/4.6/layout/grid/

So what you essentially do here is to redefine the xl breakpoint from 1140px to 1100px (btw. in any place, not specific to the meal plan recipe modal, so [once again](https://github.com/grocy/grocy/pull/1628) not very well tested).

Very minor tweak and very great if that's fits better your personal use case / screen size where you want to run Grocy's meal plan. However, it doesn't make any sense to include this - or how we should handle fitting everyone's very personal screen size? I have an idea: Simply use [custom CSS](https://github.com/grocy/grocy?tab=readme-ov-file#adding-your-own-css-or-js-without-to-have-to-modify-the-application-itself).

Thanks anyways.

adaschma:
I see you made up your mind. I would still like to add, that it doesn't change the break point, but uses the width when it makes a noticeable difference. Just to make sure we understand each other, correctly: This change is rejected on a fundamental level and not because of anything technical that can be fixed (because you would rather just keep bootstraps defaults). I don't agree with your choice, but it is yours to make.

berrnd:
>  when it makes a noticeable difference

On your particular screen / viewport size, not in general. If you think you know better the breakpoints to use in general, I suggest to improve Bootstrap itself.

&nbsp;
> This change is rejected on a fundamental level

Grocy should fit generally everyone and not only you, that's the reason, you don't need to understand that.

&nbsp;
I'm tired of those discussion of base principles. Use Grocy, or not. Simply utilize [custom CSS](https://github.com/grocy/grocy?tab=readme-ov-file#adding-your-own-css-or-js-without-to-have-to-modify-the-application-itself) to make it pixel perfect on your very personal screen size, or not. In any case, please don't waste my time again with such useless and just not thought-to-the-end stuff (again: you change the modal-xl max-width for each and every place, not only the meal plan as you've mentioned), it was the second time.

Issue #2484, opened 2024-03-01
Feature Request: Allow filenames with non-ASCII characters


wonderfulShrineMaidenOfParadise:
## How to reproduce

1. Save a file as `ビール.jpg`
2. Edit master data, product `Beer`, and upload the file `ビール.jpg` as the picture
3. Save
4. See the error

berrnd:
Something that currently works as defined can't be a bug, even that you expect non-ASCII filenames should be possible to be used.

This is the code part you need to extend if you want that - feel free to start working on that right away:

https://github.com/grocy/grocy/blob/e01e0f3abe85945f8fac3ee2117dcbeb4a4e90aa/helpers/extensions.php#L224-L323

berrnd:
> I will take a shot at adjusting the code.

Thanks, but it failed miserably: https://github.com/grocy/grocy/pull/2569#pullrequestreview-2246138099

PR #2569, opened 2024-08-19
Update extensions.php

githubber:
Updated the regex pattern to allow any Unicode character in the filename.
I admit that I used ChatGPT 4o mini to help me since my regex-foo is weak. But this looks like it should work.

### DIFF
--- a/helpers/extensions.php
+++ b/helpers/extensions.php
@@ -223,12 +223,14 @@ function GetUserDisplayName($user)
 
 function IsValidFileName($fileName)
 {
-	if (preg_match('=^[^/?*;:{}\\\\]+\.[^/?*;:{}\\\\]+$=', $fileName))
-	{
-		return true;
-	}
+    // Pattern allows any Unicode character except '/', '?', '*', ':', ';', '{', '}', '\'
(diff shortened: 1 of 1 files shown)


berrnd:
> I admit that I used ChatGPT 4o

Had that before, spent more than an hour of testing and it turned to work even more bad than before. I won't do that again. So please provide the specific example file names you have tested this against. Especially keep [directory traversal](https://en.wikipedia.org/wiki/Directory_traversal_attack) in mind while doing that.

berrnd (commented):
Unrelated to the RegEx pattern there's more around and you obviously haven't tested this just once with the in #2484 provided example file name `ビール.jpg`:

https://github.com/user-attachments/assets/3b45d851-5f27-4f3e-96a6-5b15723df290

Can't say how frustrating it is to review such stuff, wasted another half an hour of my evening, thanks for that. If you don't really have the passion to fix or improve something, just don't do it. Asking an AI chatbot while not using a single own brain cell is probably the worst way of all.

berrnd:
Turns out to be completely unrelated to the backend check (which already handles non-ASCII characters fine, so just another thing you haven't even tested / thought about for a millisecond beforehand I guess) - superseded by c21090f5220bd5bc449d2f329f06bd454634860f.

PR #2574, opened 2024-09-01
Make date in backup filename ISO 8601 compliant

stefan-muc:
By making the date in backup filename ISO 8601 compliant (year-month-day) an alphabetical sort of the backups automatically will be a chronological sort of the backups. This is more convenient.

As the automatic deletion of backups older than 60 days relies on the modified date of the file (and not the filename), this will continue to work without problems.

See also: https://www.xkcd.com/1179/

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -9,7 +9,7 @@ set -e
 shopt -s extglob
 pushd `dirname $0` > /dev/null
 
-backupBundleFileName="backup-`date +%d-%m-%Y-%H-%M-%S`.tgz"
+backupBundleFileName="backup_`date +%Y-%m-%d_%H-%M-%S`.tgz"
 echo Making a backup of the current installation in ./data/backups/$backupBundleFileName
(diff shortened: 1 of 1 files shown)


berrnd (approved):
👍

berrnd:
> See also: https://www.xkcd.com/1179/

Thanks for the quick lesson - never heard about ISO 8601 at all so far ([I guess)](https://github.com/grocy/grocy?tab=readme-ov-file#input-shorthands-for-date-fields).

There is never a rule or law on anything related to Grocy, each and every little detail is at the end up to personal preference - I think it's good to never forget that.

stefan-muc:
Yeah sure it's personal preference too, that's why I mentioned a cartoon to keep the humor - but I prefer ISO 8601 for the reasons mentioned and think this is the most intuitive one.

Thanks for the head up to the input shorthands, I didn't know that until now, will make entering due dates easier for me!

Issue #2595, opened 2024-10-25
Feature Request: Exclude self-produced product prices from spending reports


danielr18:
_Originally reported at: https://www.reddit.com/r/grocy/comments/1gbtrb8/should_products_created_from_recipes_affect_the/_

Currently, when a self produced product is created from a recipe, it affects the spending report as if it were a purchased item. This can lead to inaccurate spending reports.

### Proposed solution
Exclude self-produced product prices (those created from recipes) from spending reports.

berrnd:
Done by #2596 / 5b48004449dd02ba22eec86ce42e7bc5f24e80bf.

PR #2596, opened 2024-10-25
Exclude self-produced product prices from spending reports

danielr18:
Possible solution to #2595 which I think achieves what you mentioned in Reddit

> Spontaneously I'd also say self produced products should simply be excluded from that report

The spending report uses the `products_price_history` view. By adding the `transaction_type` to that view, we can exclude the `self-production` transactions from the spending report queries.

### Testing
I tested the spending report view to check it was working as expected, and the product price history endpoint by visiting the Purchase/Inventory views.

The endpoint that returns the products price history only returns a subset the properties, so there wasn't any change introduced to it. For example, if you go to the inventory and select the produced product, you would still see its price.

```
$rows = $this->getDatabase()->products_price_history()->where('product_id = :1', $productId)->orderBy('purchased_date', 'DESC');
        foreach ($rows as $row)
        {
            $returnData[] = [
                'date' => $row->purchased_date,
                'price' => $row->price,
                'shopping_location' => FindObjectInArrayByPropertyValue($shoppingLocations, 'id', $row->shopping_location_id)
            ];
        }
```

And there's also the products_last_purchased view which wasn't changed either. I guess one could argue that the price for self produced transactions could be ignored in there, but I didn't want to add that to the discussion when the issue is about the spending report.

```sql
DROP VIEW products_last_purchased;
CREATE VIEW products_last_purchased
AS
SELECT
    1 AS id, 
-- Dummy, LessQL needs an id column
    sl.product_id,
    sl.amount,
    sl.best_before_date,
    sl.purchased_date,
    sl.location_id,
    sl.shopping_location_id,
    IFNULL((SELECT price FROM products_price_history WHERE product_id = sl.product_id ORDER BY purchased_date DESC LIMIT 1), 0) AS price
FROM stock_log sl
....
```

### DIFF
--- a/controllers/StockReportsController.php
+++ b/controllers/StockReportsController.php
@@ -10,6 +10,7 @@ class StockReportsController extends BaseController
 	public function Spendings(Request $request, Response $response, array $args)
 	{
 		$where = "pph.transaction_type != 'self-production'";
+
 		if (isset($request->getQueryParams()['start_date']) && isset($request->getQueryParams()['end_date']) && IsIsoDate($request->getQueryParams()['start_date']) && IsIsoDate($request->getQueryParams()['end_date']))
 		{
 			$startDate = $request->getQueryParams()['start_date'];
--- a/controllers/StockReportsController.php
+++ b/controllers/StockReportsController.php
@@ -14,12 +14,12 @@ class StockReportsController extends BaseController
 		{
 			$startDate = $request->getQueryParams()['start_date'];
 			$endDate = $request->getQueryParams()['end_date'];
-			$where = " AND pph.purchased_date BETWEEN '$startDate' AND '$endDate'";
+			$where .= " AND pph.purchased_date BETWEEN '$startDate' AND '$endDate'";
 		}
 		else
 		{
 			// Default to this month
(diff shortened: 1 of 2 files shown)


berrnd (approved):
The [first state](https://github.com/grocy/grocy/pull/2596/commits/83c1cf01606540c6257645da84b4a235a2cf89fb) provided here didn't work, so obviously it wasn't tested once. I hate that, it wastes my time, it's frustrating af. If you want my time on reviewing stuff, please make 110% sure to test it thoroughly before (!) submitting. If you don't have the passion to do that, please better don't provide a PR here in the future.

Obviously, while I already had a look, you noted that yourself, since the [second provided state](https://github.com/grocy/grocy/pull/2596/commits/be513122fa434f86d98d54f42fa59d050a65ef16) works. Thanks anyways.

danielr18:
Thanks for reviewing it.

And you are right. I created this in a rush without much of a reason.
I was figuring out how to run it locally after I created the PR before figuring out I could just edit the file in the docker container.

If I make any contribution in the future, I'll make sure to test it before putting it out for review.

Issue #2156, opened 2023-03-03
Feature Request:  Add a "Note" field to the Comsume entry


starcomjing:
Hi , It would be nice to have a Note field in the consume entry (like in the purchase entry),So it could appear in the stock journal.  Would be useful to put a note on why I comsume that product or that quantity.

![image](https://user-images.githubusercontent.com/126898405/222772128-d4b0b7e9-e439-4a3d-8e0e-ae485dd74399.png)
_* image of the Note field in the Purchase entry_

TheCDC:
Perhaps this note field could also be automatically populated when the stock entry is consumed by a recipe similar to how a shopping list item automatically gets a note when you add a recipe to the list.

xfze:
Not only I would like to bump this up as it is a feature request that would be very useful for my use case, as I can offer a dirty workaround while/if it's not implemented:

you can purchase a negative quantity via the API and add a note, it will lower the stock by the desired amount and the note will be visible on the journal. Big downside is that will be registered as a "purchase" (with a negative value though - you can test it on the API sandbox).

I don't know if this should be viewed as a bug and there is the risk that it will be "fixed" in newer versions...

nevertheless even if this feature is not included in the main GUI I would love it to just be included in the API (adding a "note" parameter to the request body schema of the consume POST method similarly to the one already in the purchase method).

smreeves:
Maybe this would be the same for a note girls when a recurring schedule is completed? For example, I did yard work this week. I checked my gutters since we have trees and they were good except for the south side. Is like to note that so I can see it when I do yard work next week. Or am I doing this wrong?

berrnd:
> for a note for when a recurring schedule is completed

You're talking about chores, completely off-topic here (please be so kind and avoid spamming threads in the future) and possible since years, simply create a Userfield for the entity `chores_log`.

GMOogway:
thanks @me-ssaxena, we need this feature. @berrnd

PR #2621, opened 2024-12-16
#2156 added note field to consume page

me-ssaxena:
As part of the ticket, one part is to add note field to the consume page. Have added that functionality.

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -315,8 +315,13 @@ class StockApiController extends BaseApiController
 				$allowSubproductSubstitution = $requestBody['allow_subproduct_substitution'];
 			}
 
+			$note = null;
+			if (array_key_exists('note', $requestBody)) {
+				$note = $requestBody['note'];
+			}
+
 			$transactionId = null;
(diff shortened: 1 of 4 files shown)


me-ssaxena:
hey @berrnd - as part of the requirement I have added the note field to the consume page. I am new to both grpcy and PHP development, so please check if everything looks okay. Thanks.

berrnd:
This needs a little more thinking about the general concept.

Let's recap what the note field of stock entries / on the purchase page is for / why it was added:
It's there to add additional info for a stock entry, often used as an alternative approach for parent/child products to distinguish same, yet different products. E.g. only having and using a generic "Chocolate" product, but mentioning the exact type/brand there.

So of course this note needs to be reflected on all stock journal lines related to the corresponding stock entry, to be able to trace what happened to that specific stock entry over its lifecycle.

What you do now here is to simply overwrite that useful information on the consume transaction, making it impossible to identify what specific stock entry it was on the stock journal's consume action - sorry, but that's way too short-sighted.

&nbsp;
> I am new to both grpcy and PHP development

Don't get me wrong, I appreciate your efforts, but life has thought me that exactly this constellation + doing some copy paste work leads to more work (and often frustration) than if I do the corresponding "work" myself, after thinking about an end-to-end concept. If you're new to Grocy as a user, this becomes even more difficult as you rarely have the complete vision/app in your head. So thanks again, but this is not the way to add this particular feature - of course you can use it on your instance like so if that works for your personal use case - that's the magic of Open Source.

PR #2624, opened 2024-12-25
Replaced all occurences of "Are you sure to" with "Are you sure you want to"

Forceu:
"Are you sure to" is not grammatically correct, therefore replacing it with the correct version


berrnd (commented):
Localizations are not managed here, but via Transifex ([README ref](https://github.com/grocy/grocy?tab=readme-ov-file#localization)) - means all `*.po` files are managed externally.

&nbsp;
> "Are you sure to" is not grammatically correct

Maybe, but it's widely used on this planet and everyone get what's meant. Changing localization strings just for the sake of changing it creates unnecessary translation work on each and every language - for nothing better at the end.

Thanks anyways.

Forceu:
> Maybe, but it's widely used on this planet and everyone get what's meant.

No, it is not widely used in this form, as it is not correct. Yes, people will know what you mean, but it is like saying "Möchten das löschen?" vs "Möchten Sie das löschen?" in German.

berrnd:
> No, it is not widely used

Google around, check other software and come to the conclusion of the opposite.

&nbsp;
> but it is like saying "Möchten das löschen?" vs "Möchten Sie das löschen?" in German.

Der Vergleich hinkt komplett bzw. ist einfach falsch. Both is grammatically correct ([ref](https://english.stackexchange.com/a/222327)).

&nbsp;
Lately a lot of way too minor details, mainly all about some kind of fetishes on visual and other stuff, come up here - Grocy is about helping with the practically relevant stuff. IMHO time wasting and frustrating when afterwork from my side (and here from the whole translation community; this results in 26 strings to be re-translated across 32 languages) is required. That said, I did the changes technically correct in 5f26e914a20bdf16885089a834993c60049462ec. Guten Rutsch.

PR #2625, opened 2024-12-25
Changed dialog size to display API key delete dialog correctly

Forceu:
The api key was too long for the dialog

Before:

image

After:

image

### DIFF
--- a/public/viewjs/manageapikeys.js
+++ b/public/viewjs/manageapikeys.js
@@ -34,6 +34,7 @@ $(document).on('click', '.apikey-delete-button', function(e)
 	bootbox.confirm({
 		message: __t('Are you sure to delete API key "%s"?', objectName),
 		closeButton: false,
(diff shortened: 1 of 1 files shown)


berrnd (changes requested):
If you want to change/improve stuff, please be so kind and test it thoroughly, everything else creates triple work on my end.

Since the whole Grocy web frontend is responsive, that's of course entirely dependent on your viewport size and even with `size: 'large'` it looks at the end the same on common smartphone resolutions:

grafik

So improvement ideas in this (very minor) area from my side:
- Either wrap the text or (IMHO better)
- display the description instead or additionally to the key itself (the description option was added some releases ago).

berrnd:
Superseded by 0270f6aa0899a8c1b8a2622f2de3a23efa6a0e76.

Forceu:
Thank you! :)

Issue #156, opened 2019-02-28
Feature Request: pushbullet notifcations


dirknina:
Any thought of adding push bullet or other types of notifications to send to phones instead of logging in via web ui.  For Example you can push your grocery list to listed phones on a set day schedule or do it manualy via UI.  Really love the program its a time saver.

berrnd:
Great idea, maybe needs some definitions first what (the shopping list or really the stock-list?) to send and when (only manually or also automatically)...

Is also related to #142 maybe.

dirknina:
ya something like #142  that for where we can send the grocery list on ( x day every 2 weeks ).  as for his idea of the chores maybe a reminder push ( X time before event ) ?  just some ideas

cbrodine:
It might be interesting to setup something like MQTT support or similar so that it can be integrated with Node-Red or Homeassistant and you can make your own self-hosted automatons?  Then a person could integrate with pushbullet (other other systems) themselves on whatever schedule they wanted.

CyberRad:
SMS or email would be something I would like to see for this. That way there is not an additional application to manage. I was thinking of a configurable notification that could send a complete grocery list and or items that were added to a grocery list. This with a configurable frequency with options like daily or weekly digest, when an item is added or quantity updated.

SebRut:
> It might be interesting to setup something like MQTT support or similar so that it can be integrated with Node-Red or Homeassistant and you can make your own self-hosted automatons? Then a person could integrate with pushbullet (other other systems) themselves on whatever schedule they wanted.

I'm actually working on a homeassistant integration at [SebRut/pygrocy](https://github.com/SebRut/pygrocy) and [custom-components/grocy](https://github.com/custom-components/grocy) which is using the REST API to fetch data.

markusressel:
Thanks to the work of @SebRut I was able to quickly build a telegram bot for Grocy. Take a look if you want: https://github.com/markusressel/grocy-telegram-bot
Although functional, it is still very much a work in progress though, so issues, contributions etc are welcome.

HydrelioxGitHub:
I'm a user of telegram and it could be great to have the bot inside grocy instead to have two distinct system.

markusressel:
I think keeping the bot a separate project makes more sense in terms of maintainability, but it would certainly be an option to integrate simple (telegram) notification support into Grocy. For telegram its a s simple as a GET request and other platforms probably aren't much more difficult.

skluthe:
[...]

Issue #1293, opened 2021-01-26
Feature request: Automatic testing (initially focus on lint + styling)


mountainsandcode:
First of all, thanks for the hard work on this @berrnd - think it’s a brilliant tool already!

In order to facilitate development, I was thinking it might be helpful to add:
- Automatic code linting to check for errors, typos, etc, which could also be used to enforce the style guide (this could help prevent things like #1260)
- Potentially, in the future, doing some automated testing, though I think this is less of a priority

Happy to try to help on setting this up - if you think it’s helpful. If you do, and you have some ideas, happy to try to consider those in a possible PR!

berrnd:
Thanks for the suggestion.

See https://github.com/grocy/grocy/issues/1114#issuecomment-720150440 and the there referenced other issues/comments about "dev topics".

I aready use (PHP) code linting, the changes which caused #1260 were not done by me. In general those things maybe cause 0.001 % of the reported bugs, all other things are UI glitches or simply logic errors. Both is hard to tackle by automatic tests, at least it's a lot of work, probably more then building/fixing the feature/problem itself - so out of scope for grocy for me.

PR #2643, opened 2025-01-16
More powerful webhooks

mgarratt:
Firstly, thank you for all the effort that has gone into Grocy!

I've only recently come across it, and one thing I would like is the ability to sync shopping list items to my to-do list. Looking around, the usual approach for something like that would be to build a separate application which uses the API, but I'd like to find a more succinct solution than polling for changes.

I saw there was already a webhook implementation for the label printer, and I thought perhaps this could be expanded to send various events to whatever system a user configures. This would also allow things like #156 (push notifications) where the receiving system supports webhooks.

At the moment, this is thoroughly untested and needs a lot of extra work, such as storing webhooks in the DB and adding a UI, but I wanted to share early and see if it's something you think would be a helpful addition to Grocy.


berrnd:
Sounds great, but also like a bigger rewrite/changeset.

If you want to go on that, feel free to do so, but please do really thorough tests and everything before having my review on it. From my nearly 8 years experience here and elsewhere, bigger contributed changes always tend to use a lot of after-work on my side, since just not everything was considered and tested - passion is the key.

mgarratt:
Apologies for taking a while to reply. I'm putting together a list of things to test before and after to be sure it all works. Do you have any opinions on adding automated tests?

berrnd:
> Do you have any opinions on adding automated tests?

Yes: #1293

PR #2661, opened 2025-01-31
Make labels on login form point to correct inputs

ibizaman:
I noticed this because I'm writing a playwright test for Grocy and the `get_by_label` selector was not working. Now it's working.

### DIFF
--- a/views/login.blade.php
+++ b/views/login.blade.php
@@ -15,7 +15,7 @@
 			novalidate>
 
 			<div class="form-group">
-				<label for="name">{{ $__t('Username') }}</label>
+				<label for="username">{{ $__t('Username') }}</label>
 				<input type="text"
 					class="form-control"
 					required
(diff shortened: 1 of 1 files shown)


berrnd:
Very minor and pretty much practically irrelevant, thanks anyways.

&nbsp;
> I'm writing a playwright test for Grocy

If you're looking for that this should land here just a quick heads up to avoid potentially unnecessary work: #1293

ibizaman:
Yeah, it's as minor as it gets. That being said, now when the user clicks on the Username or Password labels, the respective input gets selected.

TBH I wasn't necessarily planning on upstreaming the tests, especially since they're on NixOS. Just so you know, my goal with the tests is to declaratively setup Grocy - LDAP integration and then test I can login with a user created on the LDAP server.

Thanks for accepting the patch anyway!

berrnd (approved):

PR #2669, opened 2025-02-06
Fix Location Content Sheet filter checkbox

DMouse10462:
Previously, the query parameter functioned as expected, but the checkbox always appeared checked. Additionally, clicking the checkbox would always result in the "unchecked" state despite remaining visually checked.

This adds the required context to the controller and an if directive to the checkbox's checked attribute in the view to match.

If you don't care for my particular style/implementation, feel free to just close this out and implement it how you feel is best. It's a two-liner and isn't worth anyone's time to go through rounds of iteration over. Just offering a fix and pointing out the issue.

### DIFF
--- a/controllers/StockController.php
+++ b/controllers/StockController.php
@@ -93,7 +93,8 @@ class StockController extends BaseController
 			'products' => $this->getDatabase()->products()->where('active = 1')->orderBy('name', 'COLLATE NOCASE'),
 			'quantityunits' => $this->getDatabase()->quantity_units()->orderBy('name', 'COLLATE NOCASE'),
 			'locations' => $this->getDatabase()->locations()->orderBy('name', 'COLLATE NOCASE'),
-			'currentStockLocationContent' => $this->getStockService()->GetCurrentStockLocationContent(isset($request->getQueryParams()['include_out_of_stock']))
+			'currentStockLocationContent' => $this->getStockService()->GetCurrentStockLocationContent(isset($request->getQueryParams()['include_out_of_stock'])),
+			'includeOutOfStockProductsAtTheDefaultLocation' => isset($request->getQueryParams()['include_out_of_stock'])
 		]);
 	}
 
--- a/views/locationcontentsheet.blade.php
+++ b/views/locationcontentsheet.blade.php
@@ -40,7 +40,7 @@
 		<input class="form-check-input custom-control-input"
 			type="checkbox"
 			id="include-out-of-stock"
-			checked>
+                        @if(!$includeOutOfStockProductsAtTheDefaultLocation) checked @endif>
 		<label class="form-check-label custom-control-label"
 			for="include-out-of-stock">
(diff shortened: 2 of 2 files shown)


berrnd:
That's only a very theoretical problem. Everything works as defined, keeping a random URL query params in sync is nothing a user cares about, the desired thing needs to work, here query parameters are used to make that happen, but adding extra logic to make those URL params more pretty is practically irrelevant, hence nothing Grocy needs. Thanks anyways.

DMouse10462:
@berrnd,

Perhaps I didn't describe the problem correctly - it is not theoretical, the filter's behavior is *actually* stuck in the "unchecked" behavior after being clicked once in its current state on the frontend:

Before Change (checked state at beginning, unchecked state for all clicks after the first):
chrome_5WBsIsp2Sp

After change, the checkbox behaves correctly. Please check the functionality in current master; this is *not* purely cosmetic or a matter of query parameters being "in sync" in some immaterial way.

Perhaps it was my mistake for being too into the technical details of the change and not describing the actual issue at hand.

Thanks,
David

DMouse10462:
Note that I didn't include the URL in the GIF screenshot - I don't care what the URL looks like either so long as it behaves correctly. The problem is that it doesn't, and the behavior suffers _for that reason._

If I had the address bar in the GIF, it would start with no query parameters. Once the checkbox is clicked once, the ?include_out_of_stock=true parameter is applied. Every future click both keeps that parameter in place _and has the effects of that query parameter in place._

The current workaround *for the user* is to strip the ?include_out_of_stock=true from the end of the URL every time they want to "re-check" the box (even though it looks checked, which is confusing for users). The URL query parameters are usually "nothing a user cares about" until they break functionality.

DMouse10462:
Corrected the title and changed the initial comment to mention that the "backend functionality" works as expected, but the frontend does not.

berrnd:
>  it is not theoretical, the filter's behavior is actually stuck in the "unchecked" behavior after being clicked once in its current state on the frontend

When something in a web browser doesn't work, the golden rule of webdev is to simply check another browser first.

My go-to browser Firefox doesn't have this problem:

https://github.com/user-attachments/assets/2d44bc78-c922-4df1-b525-8abf614de08a

In can reproduce it in latest Chrome/Edge.

What here happens on changing that checkbox is a `window.location.reload()` ([code ref](https://github.com/grocy/grocy/blob/ebcc9a422e24a5673a33cee738c56dd172d53fe7/public/viewjs/locationcontentsheet.js#L27)) and browsers seem to behave differently on what they do with existing input values.

So that's the real problem, that needs investigation, then a conclusion which is the bad behaving browser, then a corresponding change fixing this (very minor) problem. The most minor "problems" are the ones I spend the most time here lately, interesting...

berrnd:
Superseded by ce5be6645e3477f3a47414a7857a71a345e28301.

PR #2692, opened 2025-03-04
FIX: Issue with __image_url when it contains HTTP params (ie, ?<params>)

joaoc:
I had an error using a Plugin to obtain the product information from the store where I bought it.

The issue is that the store API returns the image URL with some size HTTP params.
Eg: "productX.jpg?sw=180&sh=180"

This was messing with grocy. 
The image is saved OK (with the added params on its filename), but it can't be used.

This commit make sure Grocy only uses the base URL when creating the filename for the product image from the __image_url. 

When the __image_url contains HTTP params, it will split the URL on the '?' character and only use the 1st part.

### DIFF
--- a/services/StockService.php
+++ b/services/StockService.php
@@ -626,7 +626,7 @@ class StockService extends BaseService
 					{
 						$webClient = new Client();
 						$response = $webClient->request('GET', $pluginOutput['__image_url'], ['headers' => ['User-Agent' => 'Grocy/' . $this->getApplicationService()->GetInstalledVersion()->Version . ' (https://grocy.info)']]);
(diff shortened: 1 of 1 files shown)


berrnd:
> The image is saved OK (with the added params on its filename), but it can't be used.

Then that's the real problem and should be addressed.

&nbsp;
> it will split the URL on the '?' character and only use the 1st part.

Most likely this will catch most of the mentioned "problems", however it's definitely a little bit short thought and by far not really better than before - if, why not really parse the URL fully at that place.

&nbsp;
> I had an error using a Plugin

A plugin can also take care of what you mentioned here.

berrnd:
Superseded by 2a3e23ded468bf4544bac359a1a6c08cd10ae7a3.

PR #2694, opened 2025-03-05
fix(openapi): volatile is not an array, it's an object

bertrand-sifre:
While developing a basic application using OpenAPI, I noticed that volatile did not have the correct type.


berrnd (approved):
👍

PR #2704, opened 2025-03-20
Added option to include details in the webhook sent to label printers

### LINKED ISSUE #2318: Feature Request: Send more fields to printer webhook to allow for customized labels
Hi,

Would it be possible to send more data to the webhook when printing a stock entry label?

My preferred solution would be to send as much information as possible and let the webhook decide what to do with it. Maybe all the information visible on the "purchase" page, as well as product settings like parent, due date type, ...?

However, if that's not possible, the fields I'm most interested in are:
- Category
- Due date type
- The quantity unit the label refers to
- The date the stock entry was purchased
- The date the stock entry was opened

Thank you!

Best regards,
Tobias

_Edit: Added "as well as product settings like parent, due date type, ...?" and "- The date the stock entry was purchased / opened"_

gergof:
Added a flag allowing to append details to the webhook body sent to label printers.

This would close #2318


berrnd:
> Added a flag

Definitely not needed. If the webhook receiver doesn't need the information (haven't checked in-depth if that's really everything #2318 talks about), it can simply ignore it. More options more than most often cause confusion and lead to a lot support effort.

gergof:
Removed configuration option

berrnd (commented):
Have you tested this once? It doesn't work at all like you provided it, see 9fd976509cf065160901ab4100741d4356463fcd.

Sadly it's always the same with contributions, so much after work needed... :/

PR #2706, opened 2025-03-21
Replaced barcode scanning library with ZXing

### LINKED ISSUE #2705: Feature Request: Camera barcode scanner: allow reading QR/Datamatrix codes
I've made some research and Quagga2 allows using external readers.

- For QR codes: https://github.com/ericblade/quagga2-reader-qr
- For datamatrix codes: https://github.com/hackathi/quagga2-reader-datamatrix

These packages could be included with minimal effort and then it should be possible to scan QR and Datamatrix codes as well.

For grocycodes Code128 barcodes occupy a lot of space, so only Datamatrix or QR is a viable option for smaller labels, yet currently they can not be scanned with the web app's camera scanner.

gergof:
Replaced Quagga2 barcode scanner with ZXing library.

Fixes #2705


berrnd (approved):
Seems to work, thanks.

PR #2707, opened 2025-03-21
Improve the update shell script

zero-24:
Hi,

i have here an improved update shell script adding additional log information as well as making sure the update script is beeing marked executable after the update has been run too
image

image

I could also move the settings like backup folder and release URL into a dedicated setting file. And thereby also could make it possible to configure to move the backup path outside of the web root, by default we of course would still fallback to the data folder. But please let me know when you want me to include this change into my PR or you say its to complicated.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -1,32 +1,33 @@
 #!/bin/bash
+echo "[$(date '+%d/%m/%Y %H:%M:%S')] Start updating Grocy"
 
 GROCY_RELEASE_URL=https://releases.grocy.info/latest
-
-
-echo Start updating Grocy
 
 set -e
 shopt -s extglob
 pushd `dirname $0` > /dev/null
 
 backupBundleFileName="backup_`date +%Y-%m-%d_%H-%M-%S`.tgz"
-echo Making a backup of the current installation in ./data/backups/$backupBundleFileName
+echo "[$(date '+%d/%m/%Y %H:%M:%S')] Making a backup of the current installation in ./data/backups/$backupBundleFileName"
 mkdir -p ./data/backups > /dev/null
 touch ./data/backups/$backupBundleFileName
 tar -zcvf ./data/backups/$backupBundleFileName --exclude ./data/backups . > /dev/null
 find ./data/backups/*.tgz -mtime +60 -type f -delete
 
-echo Deleting everything except ./data and this script
+echo "[$(date '+%d/%m/%Y %H:%M:%S')] Deleting everything except ./data and this script"
 rm -rf !(data|update.sh) > /dev/null
 
-echo Downloading latest release
+echo "[$(date '+%d/%m/%Y %H:%M:%S')] Downloading latest release"
 rm -f ./grocy-latest.zip > /dev/null
 wget $GROCY_RELEASE_URL -q --show-progress -O ./grocy-latest.zip > /dev/null
 
-echo Unzipping latest release
+echo "[$(date '+%d/%m/%Y %H:%M:%S')] Unzipping latest release"
 unzip -o ./grocy-latest.zip > /dev/null
 rm -f ./grocy-latest.zip > /dev/null
 
 popd > /dev/null
 
-echo Finished updating Grocy
+echo "[$(date '+%d/%m/%Y %H:%M:%S')] Make sure the new update.sh script is executable"
+chmod +x ./update.sh
+
+echo "[$(date '+%d/%m/%Y %H:%M:%S')] Finished updating Grocy"


berrnd:
> adding additional log information

The only thing I can see is added timestamps per "echo line". The whole process takes seconds unless you have a really slow as hell machine (in your example it was 6 seconds) - I don't see the point of that.

&nbsp;
> I could also move the settings like backup folder and release URL into a dedicated setting file.

No. What's the point of making things more complicated than reality needs it? The release URL doesn't change, I'm the only one on earth who provides and ships releases for Grocy.

&nbsp;
> thereby also could make it possible to configure to move the backup path outside of the web root

We have the directory of Grocy and everything Grocy needs and works with should be below this directory. Having a directory outside introduces another path where proper permissions are needed, a point a lot of people already struggle with. Again: Why making stuff more complicated than needed? If you need that, do it (e.g. simply use your own customized copy of `update.sh` to do Grocy updates), but that doesn't mean it's cool for everybody on this planet.

&nbsp;
> making sure the update script is beeing marked executable after the update has been run

That's maybe the only useful point here, but also let's face first the facts on that too: Grocy ships by default as a `.zip` file. I thought permission information is not stored in ZIPs, but [seems I'm wrong](https://unix.stackexchange.com/a/313685). So why the `update.sh` doesn't have proper permissions after unpacking needs to be checked and optimized (maybe it's simply about that I built those ZIPs on Windows). However, that's why [README explicitly contains a hint](https://github.com/grocy/grocy?tab=readme-ov-file#how-to-update) about that and since everybody can read, I think that's pretty much enough.

&nbsp;
Thanks anyways.

zero-24:
Hi thanks for the quick response.

>The only thing I can see is added timestamps per "echo line". The whole process takes seconds unless you have a really slow as hell machine (in your example it was 6 seconds) - I don't see the point of that.

That echo lines have the reason that you can see what happend when writing the output to log files. But its also fine for me to not apply my proposal here.

> No. What's the point of making things more complicated than reality needs it? The release URL doesn't change, I'm the only one on earth who provides and ships releases for Grocy.

Yes maybe a bad example, the main idea was more to allow to configure the backup path. By default of course nothing would be changed but i do accept that its not a feature you want to include.

> That's maybe the only useful point here, but also let's face first the facts on that too: Grocy ships by default as a .zip file. I thought permission information is not stored in ZIPs, but [seems I'm wrong](https://unix.stackexchange.com/a/313685). So why update.sh doesn't have proper permissions after unpacking needs to be checked and optimized (maybe it's simply about that I build those ZIPs on Windows). However, that's why [README explicitly contains a hint](https://github.com/grocy/grocy?tab=readme-ov-file#how-to-update) about that and since everybody can read, I think that's pretty much enough.

Yes I saw that and it was the first reason I did the PR against this repo. Could that might be taken out of this PR and this would be the only applied change to the default script? Wouldnt it make it just easier for everyone to make sure its executable all the time?

Thanks

berrnd:
> Wouldnt it make it just easier for everyone to make sure its executable all the time? Wouldnt it make it just easier for everyone to make sure its executable all the time?

Of course, that's why I mentioned that. If, then tackle the root problem, so why the `x` flag isn't already in the release ZIP.

Lastly remember that this script isn't used by the vast majority of Grocy users, since the vast majority run it via Docker or the HA Add-on or Grocy Desktop, so practically more than minor, also since updating Grocy isn't a daily and neither a monthly task.

zero-24:
> Of course, that's why I mentioned that. If, then tackle the root problem, so why the x flag isn't already in the release ZIP.

Sure I'm happy to take a look into that. Where can I find the build script that you are using to build an grocy release didnt saw a build folder in here? You mentiond that you are building the release on windows are you using WSL and therfor an .sh script for that or an cmd script to build the zip file?

And is the x set on your local git repo that you are using to build the package?

Thanks

berrnd:
> Where can I find the build script that you are using

Nothing is hidden here, every single bit is public and [that's the batch file](https://github.com/grocy/grocy/blob/master/.devtools/create_release_package.bat) used to populate a release ZIP.

&nbsp;
> are building the release on windows are you using WSL

No, I like it simple: Check out this repo, [install dependencies](https://github.com/grocy/grocy/blob/master/.devtools/install_dependencies.bat), execute that mentioned `create_release_package.bat` - that's it.

&nbsp;
> And is the x set on your local git repo

Windows doesn't have such a flag on files, so I haven't cared so far and doubt it (otherwise it would already work, which is exactly the not working point we're talking about here).

zero-24:
Thanks for the information, I have tried some things and took a look into docs and it seams to be the case as mentiond that the problem is between windows and linux or better the filesystems as NTFS does not support the +x attribute so extracting the zip will result into the default values for the chmod to be used which of couse does not include the +x attribute.

image

I would recommend one of the following workarounds for this issue.
- Move the build script to linux or WSL and make sure the chmod +x is run before the build script started.
- Run the chmod +x on the update script itself as proposed here

berrnd:
> Move the build script to linux or WSL

That's not going to happen.

&nbsp;
> Run the chmod +x on the update script itself as proposed here

Doesn't change anything for the first (manual) unpacking.

&nbsp;
Conclusion: The hint about that in README was enough for nearly a decade and so it will most likely also work for the next ten years.

zero-24:
> That's not going to happen.

I can understand that wanted to list it as possible solution only.

> Doesn't change anything for the first (manual) unpacking.

I agree but once it has been done the first time an update will not reset it back to not executable.

> Conclusion: The hint about that in README was enough for nearly a decade and so it will most likely also work for the next ten years.

Sure will accept it my only goal here was to help and propos an improvement to the process but its your project and of course I will accept when you dont want to adjust the script as mentiond.

Thanks for taking the time looking into this request even when you are rejecting my proposal. :)

PR #2716, opened 2025-04-10
Optimizing website

BDZ-41:
I've detected two sites where a variable is loading but never used.
This is $quantityUnitConversionsResolved, it will also be ok to remove from querying the database for the data but for now I'm trying to understand all the code.

I've observed that when a user has defined lots of quantity units (with their conversions), this variable takes a lot of space (in my case it reduced the page load from 35mb to 300k just by removing it).

I'll look the use cases, for a future refactoring, because it's not optimal to have the quantity units table by every single product and download it every time.

### DIFF
--- a/views/mealplan.blade.php
+++ b/views/mealplan.blade.php
@@ -73,7 +73,6 @@
 	var weekRecipe = {!! json_encode($weekRecipe) !!};
 
 	Grocy.QuantityUnits = {!! json_encode($quantityUnits) !!};
-	Grocy.QuantityUnitConversionsResolved = {!! json_encode($quantityUnitConversionsResolved) !!};
 
 	Grocy.MealPlanFirstDayOfWeek = '{{ GROCY_MEAL_PLAN_FIRST_DAY_OF_WEEK }}';
 </script>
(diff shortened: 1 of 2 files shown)


berrnd:
> where a variable is loading but never used

Entirely not correct. Please be so kind and test changes thoroughly before requesting my review on it - not doing that is essentially wasting my time and I don't like that too much.

The meal plan page needs it because it's required for the product amount picker component when adding a product to the meal plan, your proposed changes breaks adding products to the meal plan. I don't add expensive (in terms of computing power/time needed) stuff without thinking.

&nbsp;
> 35mb to 300k just by removing it
> because it's not optimal to have the quantity units table by every single product

When you want Grocy to handle unit conversions, unit conversion factors are needed. Having a JSON representation of 35 MB sounds odd. I would review my unit definitions and their conversions. Having a lot of data always ends up in, well, handling a lot of data unless magic is involved.

BDZ-41:
Ah, ok, sorry, I didn't know you can add a product to the meal plan, I thought it was only recipes.
I've deleted many quantity units on my instance I'm not using so I could reduce the size of that variable.

But anyway I'm looking into optimize the way it's used, 'cause I believe that variable stores each qu that can be used for every product, but if the table is big, or there are a lot of products it's more optimal to just have the table and check in execution time which of them can be used.

PR #2728, opened 2025-05-06
README.md: add required php modules for installation

wonderfulShrineMaidenOfParadise:
Fix the following errors:
Unable to run Grocy: PHP module 'iconv' not installed, but required.
Unable to run Grocy: PHP module 'tokenizer' not installed, but required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,16 @@ See the website for more installation guides and troubleshooting help. &rarr; [h
 ### Platform support
 
 - PHP 8.2 or 8.3 (with SQLite 3.34.0+)
-  - Required PHP extensions: `fileinfo`, `pdo_sqlite`, `gd`, `ctype`, `intl`, `zlib`, `mbstring`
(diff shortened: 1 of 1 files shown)


berrnd:
Those are core extension, so can't be missing unless you use a somehow fucked up PHP release.

wonderfulShrineMaidenOfParadise:
They are not included in Alpine by default:
https://gitlab.alpinelinux.org/alpine/aports/-/blob/dd0c489d6fcc9f8b5151b51117843678258ba823/community/php84/APKBUILD#L140
https://gitlab.alpinelinux.org/alpine/aports/-/blob/dd0c489d6fcc9f8b5151b51117843678258ba823/community/php84/APKBUILD#L171

berrnd:
> They are not included in Alpine PHP by default

Sounds stupid and they should check what PHP says/defines about core extensions.

Issue #2761, opened 2025-06-29
Bug: Infinite redirect loop if /data/viewcache is not writable


Loveangel1337:
New install, I don't like to give permissions for my service users to write anywhere they don't need, therefore I don't have +w everywhere.

Trying to get to / gives me a TOO_MANY_REDIRECT error in Chrome, and indeed, the / route redirects to /.

Traced the issue back to this https://github.com/grocy/grocy/blob/master/app.php#L66
```
if (!file_exists($hashCacheFile))
{
	EmptyFolder($viewcachePath);
	touch($hashCacheFile);

	if (function_exists('opcache_reset'))
	{
		opcache_reset();
	}

	// Schema migration happens on the root route, so redirect to there
	header('Location: ' . (new UrlManager(GROCY_BASE_URL))->ConstructUrl('/'));
	exit();
} 
```

Nothing checks that the file exists after the touch, therefore it loops back onto itself and tries to create the file again.

berrnd:
> I don't like to give permissions for my service users to write anywhere they don't need

Well, [README explicitly mentions](https://github.com/grocy/grocy?tab=readme-ov-file#how-to-install):
_Ensure that the `data` directory is writable_

So you haven't read that and that's not a bug.

Loveangel1337:
While that's true, and I missed that bit in the README:
- The data folder is not the data/viewcache folder
- The data/viewcache directory is included in the zip file when there is creation logic present in the app.php file so it gets setup with my initial rights on, I would have done a chmod g+w data, which would have not worked.

Therefore, either the README needs to specify that the data directory *and the data/viewcache* need to be writable, or the viewcache folder needs to not be included in the zip so that it gets created by the service user. But IMHO there is also no reason in the current version that the web user should have write access to the .htaccess or to the plugin directory either.

berrnd:
The `data` folder stores everything Grocy creates, arbitrary sub folders can be created - not something special, kind of every application which stores user data needs to have write access somewhere - README doesn't list any specific sub folder, it's about that whole folder and that wasn't an issue for ten thousands of users over the last decade.

No need to waste more time here, simply read a project's README next time before installing and applying paranoid security restrictions for no practical reason.

&nbsp;
> the web user should have write access to the .htaccess or to the plugin directory either

[...]

PR #2762, opened 2025-06-29
Add exit condition is the hashcacheFile fails to create to avoid a redirect loop

Loveangel1337:
Solves #2761

### DIFF
--- a/app.php
+++ b/app.php
@@ -66,7 +66,11 @@ $hashCacheFile = $viewcachePath . "/$hash.txt";
 if (!file_exists($hashCacheFile))
 {
 	EmptyFolder($viewcachePath);
-	touch($hashCacheFile);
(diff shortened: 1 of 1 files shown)


app.php
```diff
@@ -66,7 +66,11 @@
 if (!file_exists($hashCacheFile))
 {
 	EmptyFolder($viewcachePath);
-	touch($hashCacheFile);
+	if!(touch($hashCacheFile))
+	{
+		// We failed to create the cache file, probably because of a missing write permission, bail!
+		exit('Unable to write to the viewcache directory.');
```
jackschmidt: Maybe better as `exit("Unable to write to the data/viewcache directory.\n");` ... specify the directory a bit more and include a newline for people who are using command line / curl to debug.

PR #2770, opened 2025-07-12
Support associating a Shopping List with a Store

### LINKED ISSUE #1357: Feature request: Shopping List add Products for default store
I would like to add a shopping list for every store.
Now i would like to add every product below min stock level to the according shopping list as set in the masterdata default store field.

Occhioverde:
Good morning!

Since this feature would greatly simplify the effort of adopting Grocy in my family, I tried to put together a PR that closes #1357 allowing to associate a Shopping List with a specific Store and then pull products that are below the specified amount only if linked to the same one.

In detail, this PR:
- Adds a new (optional) `Associated Store` field to Shopping Lists;
- Changes the `/stock/shoppinglist/add-missing-product` API route to support a new (optional) `check_default_shopping_location` body parameter that, if set to `true`, only adds to the list products from the associated Store;
- Adds a new frontend button that calls the `/stock/shoppinglist/add-missing-product` endpoint with the `check_default_shopping_location` body parameter enabled.

Let me know if the code looks good enough or if changes are needed before merging.

### DIFF
--- a/controllers/StockApiController.php
+++ b/controllers/StockApiController.php
@@ -20,13 +20,18 @@ class StockApiController extends BaseApiController
 			$requestBody = $this->GetParsedAndFilteredRequestBody($request);
 
 			$listId = 1;
-
 			if (array_key_exists('list_id', $requestBody) && !empty($requestBody['list_id']) && is_numeric($requestBody['list_id']))
 			{
 				$listId = intval($requestBody['list_id']);
 			}
 
-			$this->getStockService()->AddMissingProductsToShoppingList($listId);
+			$checkDefaultShoppingLocation = false;
+			if (array_key_exists('check_default_shopping_location', $requestBody) && filter_var($requestBody['check_default_shopping_location'], FILTER_VALIDATE_BOOLEAN) !== false)
+			{
+				$checkDefaultShoppingLocation = boolval($requestBody['check_default_shopping_location']);
+			}
+
+			$this->getStockService()->AddMissingProductsToShoppingList($listId, $checkDefaultShoppingLocation);
 			return $this->EmptyApiResponse($response);
 		}
 		catch (\Exception $ex)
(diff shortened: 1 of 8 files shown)


berrnd (changes requested):
Too complicated and not complete as described in #1357, maybe what you personally need but when doing it, please do it from A to Z.

1. No new option on shopping list level is needed. Keep it simple, extend the shopping list option "Automatically add products that are below their defined min. stock amount to the shopping list" in a way to select "Store list" (or something like that) and then simply use (or create if missing) a shopping list named the same as the corresponding store.
2. No new frontend button is needed, simply extend the existing logic/places. The aforementioned setting automates adding stuff below min. stock and this totally ignores what you added (please test stuff end to end before requesting my review on something).

berrnd:
Or even a more simple idea:

Change nothing on the existing below min. stock handling and simply introduce "virtual shopping lists" per store which displays then the products across all shopping lists per (product default) store. This would then also cover manually added products, regardless on which shopping list they currently are.

Occhioverde:
First of all, thank you very much for your time in reviewing this.

I personally thought that adding a new option without changing the behavior of the existing and consolidated ones was a better choice to avoid disrupting the workflow of users that are not interested in this feature, but your proposal of adapting the "Automatically add products that are below their defined min. stock amount to the shopping list" sounds equally good to me, even for my own use case.

Just to be sure of what has to be done before starting to code, I made a quick mockup of the "Shopping list settings" page with new mutually exclusive checkboxes that would allow switching from the "old" behavior of the option ("Add everything to the following list") to the new one that I'm implementing ("Automatically sort products to different lists according to their default store"):

image

Do you think that something like this would be more adeguate to the scope of the project?

berrnd:
I [already pointed out](https://github.com/grocy/grocy/pull/2770#issuecomment-3066687133) what I would do for that and I doubt I'll merge something else.

PR #2788, opened 2025-08-03
fix triggers using double quotes, errors on SQLite 3.41+

jandemter:
You used double quotes around the message of RAISE() in a couple of triggers, this does not work on SQLite 3.41 and newer (tested on 3.50.2). I think I fixed it successfully for all triggers (did a "select * from sqlite_master where type = 'trigger';"), but I only tested where I first faced the problem, on /product/new, triggers below "-- other triggers affected" untested so far.

### DIFF
--- a/migrations/0254.sql
+++ b/migrations/0254.sql
@@ -1,0 +1,202 @@
+-- recreate some triggers that use double quotes, error on sqlite 3.41+
+
+DROP TRIGGER prevent_self_nested_recipes_INS;
+CREATE TRIGGER prevent_self_nested_recipes_INS BEFORE INSERT ON recipes_nestings
+BEGIN
+SELECT CASE WHEN((
+	SELECT 1
+	FROM recipes_nestings
+	WHERE NEW.recipe_id = NEW.includes_recipe_id
+	)
+	NOTNULL) THEN RAISE(ABORT, 'Recursive nested recipe detected') END;
+END;
+
+DROP TRIGGER prevent_self_nested_recipes_UPD;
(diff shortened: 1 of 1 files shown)


berrnd:
Please provide the exact relevant environment parameters where this is a problem.

https://www.sqlite.org/quirks.html#dblquote states:

> As of SQLite 3.41.0 (2023-02-21) SQLITE_DBCONFIG_DQS_DDL and SQLITE_DBCONFIG_DQS_DML are disabled by default in the [CLI](https://www.sqlite.org/cli.html). Use the ".dbconfig" dot-command to reenable the legacy behavior if desired.

PHP PDO (used here) isn't CLI.

berrnd:
I did some tests, it is how the SQLite changelog states it - by default double quoted string literals are disabled in CLI environments, not anywhere else and therefore not in PHP PDO. Of course you can set all those parameters to your very personal likings anywhere. Here only the (version specific) defaults are cared about, so this is essentially irrelevant. So it would be interesting how this was a problem in Grocy for you at all.

However, I agree that single quoted string literals make more sense in SQL in general. Since reviewing if you really includeded all relevant triggers in the current version takes longer than extracting the list quickly myself, I'll do that. Also past migrations should be corrected, even that SQLite obviously only checks that at execution time.

Superseded by ba2cb6ce080625f3e0ddf311c42107004b7f92ba.

Thanks for the hint.

jandemter:
Thanks! The triggers don't run via PDO, but directly in SQLite, I guess that's what made this a problem. Environment was PHP 8.3.23, SQLite 3.50.2, standard build params from FreeBSD ports.

PR #2794, opened 2025-08-15
Enhance product card chart colors for better visibility

DeepCoreSystem:
Improved the color scheme of product card charts to enhance visibility and user experience.


berrnd (approved):
👍

PR #2805, opened 2025-09-13
Replace "command" tests with regex and fix broken commands

RyanSchaefer:
Fixes https://github.com/grocy/grocy/issues/2806

Replaces "command" detection logic with regex and fixes bug where user could not input commands of greater than length 4.

[BUG]
Cannot trigger the following:
YYYYMMDD gets expanded to the proper ISO-8601 notation

[Fix Evidence]
https://github.com/user-attachments/assets/bbf0991d-39a1-421f-96f8-db5177c3c372

[BUG]
Cannot trigger the following:
YYYYMMe or YYYYMM+ gets expanded to the end of the given month in the given year in proper notation

[Fix Evidence]
https://github.com/user-attachments/assets/37a6459d-e1f0-4b97-bbc4-296ba87055df

[BUG]
Cannot trigger the following past 100:
[+/-]n[d/m/y] gets expanded to a date relative to today, while adding (+) or subtracting (-) the number of days/months/years, in proper notation

[Fix Evidence]
https://github.com/user-attachments/assets/26177a00-9c95-4f0e-980f-adb41e665c48

### DIFF
--- a/public/viewjs/components/datetimepicker.js
+++ b/public/viewjs/components/datetimepicker.js
@@ -150,7 +150,7 @@ Grocy.Components.DateTimePicker.GetInputElement().on('keyup', function(e)
 		Grocy.Components.DateTimePicker.SetValue(moment('2999-12-31 23:59:59').format(format), inputElement);
 		nextInputElement.focus();
 	}
-	else if (value.length === 4 && !(Number.parseInt(value) > centuryStart && Number.parseInt(value) < centuryEnd)) // Shorthand for MMDD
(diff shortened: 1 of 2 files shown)


RyanSchaefer:
Need to update datetimepicker2. Will add that to this.

berrnd:
Superseded by 4c5fbffd2503d59e7f603aa677195b58f9df8260, which keeps the existing logic while fixing the only reproducible shorthand problem. Even though you personally maybe like regex expression testing more, there is no need to switch to that only because of that.

RyanSchaefer:
Okay man. It's clear you don't want other people contributing to this project. Best of luck to you.

Issue #129, opened 2019-01-02
Import/Export


Braintelligence:
It would be awesome if we had a simple import/export option.
A CSV for every data model would suffice in my case.

Use-case is backup purposes and system-agnosticism. (Especially important if we want to move to a different database engine.)

berrnd:
Thanks for all your ideas! :)

The database is a not encrypted SQLite database with, in my opinion, self explaining table and column names - so that's already an option if you know a little bit SQL...

Braintelligence:
Yeah, thanks for pointing it out. I consider that a valid option if a frontend-solution won't be considered until I need it.

CyberRad:
Is there a way to restore from a backed up directory of the database and files? I only ask in this thread as I would think that you could back these items up and compress them. Even going as far as scheduling this functionality. Then within the GUI or using a documented procedure put the files in place and resume where this backup left off. I am sure there is a little more to it than that but other projects I have tested for do it this way. (Plex for example)
Really interested in this as well as users from the LinuxServers/Unraid Docker community are too. Makes  it less nerve racking to spend time inputting the data and getting the family used to the tool.

berrnd:
Everything is saved in the `/data` directory, the database itself is a single SQLite file (`/data/grocy.db`), so backing up that directory and restoring it is enough.

When using [`update.sh`](https://github.com/grocy/grocy/blob/master/update.sh) to update it makes a backup of the whole installation in `/data/backups` as a `.tgz` archive and keeps it there for 60 days.

Stoicmaverick:
It would be much more convenient and accessible if there was a GUI backup and download option such as within Home Assistant. If I'm going to "Try something real quick" it feels so much safer to be able to one-click download a current full backup that can be one-click restored if I manage to nuke something or mess up all my locations somehow.

colonelmattyman:
It would also help users share their food databases.  It is a huge PITA populating everything.

cjramseyer:
It is not possible to create a "backup"/"snapshot" from one HA instance, (coipy the snapshot to another instance), and use the restore function.
The DB for the addons is not available outside of an HA snapshot.
Having to manually re-enter everything is a huge issue.

Please add export/import functions, or at least a function to import the DB from another instance of the grocy addon

AndrewHoover:
[...]

PR #2813, opened 2025-09-25
Add download SQL database button

Luc-Mcgrady:
Adds an option to download `data/grocy.db` directly from the browser for a quick backup.
image

Related #129


Luc-Mcgrady:
I think that maybe a full Import/Export page might be a better solution. I'll try it later if that's ok?

PR #2814, opened 2025-09-27
Support `data:` URI images for external barcode lookup plugins

GammaC0de:
My external data provider returns image data as a `data:image/png;base64,xxx...` URI instead of a URL.
This PR adds support for handling such case and also provides a mechanism for plug-in writers to return an image even when the image URL is not publicly accessible.


berrnd (approved):
👍

Thanks.

PR #2843, opened 2025-11-29
Forbid empty username header for reverse proxy authentication

GeoffreyFrogeye:
Consider the case where a Grocy user wants to use reverse proxy authentication OR authentication with an API key.

By default, authenticating reverse proxy will unconditionally perform user authentication, before allowing a request to be passed to Grocy which will verify an API key, rendering them virtually useless. So the proxy needs to be configured not to perform authentication in certain cases, for example the presence of the `Grocy-Api-Key` HTTP header, or the path being a subdirectory of `/api/`.

It can however be tricky to configure this though. For example, nginx does not allow conditional authentication, or conditional header setting. In those cases, when the condition is met, the username header would still be returned, albeit empty. Grocy will then create an user with empty username (not normally possible). It default to having all permissions, and while most are removable, reading chores/stock/etc. do not seem to be, and would thus be public.

This returns an error when the username is empty, to accomodate for those cases.

### DIFF
--- a/middleware/ReverseProxyAuthMiddleware.php
+++ b/middleware/ReverseProxyAuthMiddleware.php
@@ -43,9 +43,14 @@ class ReverseProxyAuthMiddleware extends AuthMiddleware
 			if (count($username) !== 1)
 			{
 				// Invalid configuration of Proxy
-				throw new \Exception('ReverseProxyAuthMiddleware: ' . GROCY_REVERSE_PROXY_AUTH_HEADER . ' header is missing or invalid');
+				throw new \Exception('ReverseProxyAuthMiddleware: ' . GROCY_REVERSE_PROXY_AUTH_HEADER . ' header is missing');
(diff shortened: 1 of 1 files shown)


GeoffreyFrogeye:
Yeah, I spent quite some time trying to configure my reverse proxy for this use case and still haven't managed to do it, hence why I resorted to patch grocy's code. I thought it would be ok given Grocy already validates the "empty username" at other places, [including this middleware](https://github.com/grocy/grocy/blob/68b4abfac42f80523d9aba22b5a5efd5a5d62211/middleware/ReverseProxyAuthMiddleware.php#L34), and it might prevent future people like me from not realizing they had `/api` open to the public for months 😅.

But I understand this is putting maintenance burden where it's not really supposed to be, so no problem if you just close. If I really can't make nginx conditionally set the header I'll just keep the patch in my package manager.

PR #2850, opened 2025-12-18
Frontend - CSS adjustments

mrtncode:
Hey, I've made a few changes to the design to give the front end a better look (especially in light mode). I'm not sure if design changes are desired. Designs are always very subjective.
If it's not desired, I'll gladly close the PR and continue using the design in my personal CSS theme file :)

I look forward to your reply.

### DIFF
--- a/public/css/grocy.css
+++ b/public/css/grocy.css
@@ -8,7 +8,6 @@ body {
 }
 
 .content-text {
-	color: #444;
 	font-size: 1rem;
 }
 


MarzyCoder:
Thank you for your feedback :))
I will just using it as my personal css theme file. Maybe I will share it on GitHub also :) 
Are there any other plans to revamp the front end at some point? In my opinion, the design is a bit outdated, but that's probably just my view 😁

berrnd:
> Are there any other plans to revamp the front end at some point?

This is a hobby, not a business - I don't have a strict time schedule for any of my hobbies, hope that's the case for yours too.

&nbsp;
> In my opinion, the design is a bit outdated, but that's probably just my view 😁

As said and not only true for this specific project: 14561 users, 15897 opinions on any single design decision. Btw: Grocy uses [Bootstrap](https://getbootstrap.com/) as the base for everything design, the most widely used framework for such things, at least on this planet. If you think they do design wrong and you can do better, you should do it there to help humanity as a whole. In general I focus on efficient data input and output since that what Grocy (or any ERP system) is about, not on fancy looking rounded borders today and edgy once tomorrow, just because that's the next big hype introduced by whoever. And just to remember again: This is just a random hobby project, primarily created to support my very own needs, published to the world for free to see if it's useful for others too, which it evidently is.

MarzyCoder:
Everything's fine :) I didn't mean to be mean or anything. I understand, of course, that it's a hobby project. I'm just used to different designs (e.g., from Home Assistant). 😊
But I can tell that things are a bit different here, and that's totally fine.

berrnd:
> I'm just used to different designs

To be honest and to conclude this: In this PR you added some space here and there, made some corners more round and changed some colors slightly. This doesn't change anything to the better or worse nor it reinvents the wheel. Just very minor layout adjustments you obviously visually like better, which is fine and for which I invented the option for [custom CSS](https://github.com/grocy/grocy?tab=readme-ov-file#adding-your-own-css-or-js-without-to-have-to-modify-the-application-itself) years ago.

Maybe you're a design person, I know and understand that they have to change visual stuff just for the sake of changing something, due to otherwise it looks to others they don't do anything. When you do that multiple decades at scale you get design clutter and inconsistency everywhere, check out Windows for a great example on that all.

Fun fact: I wanted to switch my Home Automation stuff to HA not only once since it's a great system, but stepped back due to its very inefficient UI everywhere. So everyone has a different focus, some prefer round edges and fancy design stuff, for others getting things done in an efficient way is the primary goal. Grocy is and will ever be about the later.

PR #2866, opened 2026-01-31
Fix 'Due Soon Days' User Settings for Chores and Batteries

trevortabaka:
I recently set up Grocy for the first time and noticed changing the 'Due soon days' for Chores and Batteries did not persist like for other features.

I was able to reproduce on the public demo pages for both Stable and Preview, and searched but did not see any existing Issues or Pull requests referencing it here on GitHub.

The reproduction steps were:
1. Select wrench settings icon in upper-right corner
2. Choose "Batteries settings" (or "Chores settings")
3. Change the numerical value for "Due soon days"
4. Refresh the page (or select "OK")
5. The value is reset to what it was before

Thanks for creating Grocy and taking the time to review!

### DIFF
--- a/views/batteriessettings.blade.php
+++ b/views/batteriessettings.blade.php
@@ -17,7 +17,7 @@
 
 		@include('components.numberpicker', array(
 		'id' => 'batteries_due_soon_days',
-		'additionalAttributes' => 'data-setting-key=" batteries_due_soon_days"', 'label'=> 'Due soon days',
+		'additionalAttributes' => 'data-setting-key="batteries_due_soon_days"', 'label'=> 'Due soon days',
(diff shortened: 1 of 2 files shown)


berrnd:
Well, at the end it is about those two spaces. However, there is more around and it would be cool to look at the whole thing: The HTML markup is broken, so the PHP-CS-Fixer rules / formatting failed ([ref](https://github.com/grocy/grocy/blob/master/.php-cs-fixer.php)). This needs to be corrected otherwise those spaces will be there again after applying those rules.

berrnd:
Superseded by #2867 and 471f21e9925f6b0131ab4add3e5e6ca5190012c6.

Issue #686, opened 2020-04-01
Feature: Add ingredients of recipe to shopping list, even if they are already in stock


nanderv:
Hi,

I want to use Grocy to keep track of my personal storage of groceries. It seems to be great for that (much better than my current 'Excelsheet of doom'), but it lacks one feature.

I'm living in a student house, with some flatmates. We cook for each other, and then share the costs. To make this easier, it's nice to just buy the groceries for a cooking session, and put the receipt on the system we use for keeping finances fair (we're Dutch ;))

I'd therefore like to have a button that says: add all the ingredients of this recipe to a shopping list, regardless of whether I already have enough in stock.

Ps. if this button already exists, I'm sorry, I tried it out in the demo, and couldn't find it.

mcdoogs:
What if the 'put missing products on shopping list' button became more like a general 'put ingredients on shopping list' button? It could show all ingredients on the checklist, but only check off the ones not in stock. A check/un-check all option could be added for ease of use, especially helpful in this use case where you are adding all ingredients.

nanderv:
That sounds like a fix that could work, indeed.

Kryptortio:
This might also solve the issue I noticed, if one item is marked as "Not enough in stock" but the rest is in stock I still can't use the add button for what's missing. So the check-mark suggestion might be a good idea and/or possibly an add button for each recipe row. 

Maybe consume should be allowed even is something is missing as well because you could decide to skip it or replace it with something else.

berrnd:
@Kryptortio off-topic things are untrackable, this here is about what the title says, nothing else.

&nbsp;
> if one item is marked as "Not enough in stock" but the rest is in stock I still can't use the add button for what's missing

Should of course work, please provide a reproducible example and open a separate issue if that's reproducible.

I've created an example, here: https://test-lkistxe5hglv622ujwtx3c.demo-prerelease.grocy.info/recipes?recipe=4
One ingredient is missing and not already on the shopping list, the button "Put missing products on shopping list" can be used.

&nbsp;
> Maybe consume should be allowed even is something is missing as well because you could decide to skip it or replace it with something else.

Already requested/tracked in #548.

9Lukas5:
Is this feature being considered anytime to be implemented? 👀

9Lukas5:
> > Is this feature being considered anytime to be implemented?
> 
[...]

PR #2868, opened 2026-01-31
Recipe: allow adding full recipe required ingredient amount

9Lukas5:
As you asked in #2864, I extended the existing popup.

This modifies the bootbox on to add recipe ingredients to the shopping list in a way, that

- button to open bootbox is always enabled
- popup lists all* ingredients, independent of fulfillment state
- two new quick-select buttons to check only missing or all ingredients
- radio button selection to define if the full or only missing amount shall be added to the shopping list

*This still excludes ingredients which have the stock check disabled in a recipe.
In order to get this information, the RecipesController was modified that it adds the data from the recipe ingredient setting as new additional field in the recipePositionsResolved.

I hope this is now the right direction to approach this feature?
Additionally, I put it into a single commit and didn't use the conventional commits style, sorry that was just by habit.

If you want me to change anything else, please let me know.

Example Images:
Bildschirmfoto vom 2026-01-31 22-23-33
Bildschirmfoto vom 2026-01-31 22-05-06

By default it checks only the missing ingredients and only the missing amount, reflecting the same behavior as the current one.

closes https://github.com/grocy/grocy/issues/686

### DIFF
--- a/controllers/RecipesApiController.php
+++ b/controllers/RecipesApiController.php
@@ -16,13 +16,19 @@ class RecipesApiController extends BaseApiController
 
 		$requestBody = $this->GetParsedAndFilteredRequestBody($request);
 		$excludedProductIds = null;
+		$ignoreStock = false;
 
 		if ($requestBody !== null && array_key_exists('excludedProductIds', $requestBody))
 		{
 			$excludedProductIds = $requestBody['excludedProductIds'];
 		}
 
-		$this->getRecipesService()->AddNotFulfilledProductsToShoppingList($args['recipeId'], $excludedProductIds);
+		if ($requestBody !== null && array_key_exists('ignoreStock', $requestBody))
+		{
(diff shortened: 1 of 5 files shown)


berrnd:
Just quickly looking over this:
- Changed new translation source/base strings are still missing
- Looping over all $recipePositionsResolved and executing another SELECT per row is performance wise the worst thing ever
- The added new API parameter is missing in grocy.openapi.json
- Radio buttons use a different style than all the other ones throughout Grocy
- Displaying an amount without a unit doesn't make sense at all (100 meat are missing, 100 killolinos I guess)
- Code style still doesn't align
- Most likely more

Sorry, but when reviewing and talking about it back and forth takes already twice as much time as doing the corresponding thing myself, something is wrong. Thanks again anyways, but we should stop that there, at least I doubt I will review anything like that again.

Feel free to continue this on your Fediverse thread. Will lock this here since I just don't want to waste any more time.

Issue #481, opened 2019-12-31
Feature Request: Sublocations


bartlibert:
I'd like to be able to divide a location into several sublocations, e.g. several drawers in the refrigerator. I tried making each drawer a separate location, but then they all get printed on separate pages in the overview, while I'd like to have one location per page, but with subdivisions per sublocation.

E.g 

**Refrigerator** 
_Top drawer_ 
- 1kg spinach
- Spaghetti sauce
- ...

_Middle drawer_ 
- 500 g spinach
- lamb chops
- ...

I tried to do this with user defined field, but I did not find a way

hannahswain:
I've solved this by using a structure in the location name, i.e. `Kitchen > Fridge`, or `Kitchen > Spice rack`. That keeps things grouped together nicely in the UI and on the print out.

Stoicmaverick:
I posted the request for #671 , and this might be an option if nesting more than one location is possible. For instance my 100k resisters might be in "Workbench - Drawer 2 - Compartment Column B - Compartment Row 4, which might be displayed simply as "Workbench Drawer 2 Compartment B-4"

t-gergely:
I need recursive locations, too, so I can migrate my inventory to Grocy from Piwigo. I have building/room/cupboard/shelf/side/box etc.

jose1711:
Found http://www.twisterrob.net/project/inventory/ which seems to focus solely on home inventory. Maybe this can serve as a model/inspiration too.

DieKatzchen:
The way the Magic Home Inventory works is kind of clever. All items are also locations, all locations are also items, so you can put anything inside anything. That way I can put the remote in a drawer, but I can also put the rechargeable batteries in the remote. Then you can search for an item and it will tell you where it is. If I could do this in Grocy it could also tell me when I put those batteries in and I could attach some notes like maybe for a universal remote I could put in the manual in case I ever need to reprogram it. Stuff like that.

Saturo78:
A simple ERP warehouse logic would be perfect.
In the Location there are storage spaces.
100-01-01-01 stands  for Location 100-shelf 1-level 1- storage space 1
This Code could be printed as QR-Code an placed near the storage space. and can be used for the Transfer Transaction.

berrnd:
@Saturo78 this is a about parent/child locations or something similar, definitely not about attaching barcodes to locations which is a duplicate of #221 (please keep it to one topic per request).

mgh2:
[...]

PR #2869, opened 2026-02-01
481 sublocations

miicha:
Add sublocation suppport, as mentioned here: https://github.com/grocy/grocy/issues/481
It might not yet be complete...

### DIFF
--- a/migrations/0255.sql
+++ b/migrations/0255.sql
@@ -144,3 +144,42 @@ BEGIN
 		SELECT 1 FROM descendants WHERE id = NEW.parent_location_id
 	) NOTNULL) THEN RAISE(ABORT, 'Circular location hierarchy detected') END;
 END;
+
+-- Trigger to inherit is_freezer from parent on INSERT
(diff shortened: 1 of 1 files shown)


berrnd:
Looked a bit over the already done changes and can already say: Without a very in-depth explanation of why every single change is needed, this won't work for the reasons just mentioned. And even with that, getting into it will most likely need more time than doing it myself.

It looks also like a lot of copy-paste-adapt work. Re-using existing and maybe not the best approaches to solve such kind of complex things is likely not the way I would pick, but I have neither the time nor the motivation to get into all this right now either way. Consider maintaining this in your own project / fork.

miicha:
Thanks for your explanations. I can totally understand your point.
I'll try to find some time to minimize the amount of changes. Maybe I can find a better way to implement it, but first I would need to dive deeper in the codebase and this might take some while.
Should have porbably started with a smaller feature to get familiar with the code, but this one is most interesting to me...

PR #2873, opened 2026-02-08
Adapt stock_current view to add stock_ids

detobel36:
Currently, using the API endpoints, we are unable to retrieve the list of stock IDs...     
The goal of this pull request is to adapt the view to add a new column that will aggregate all stock ids. This way, in the `/stock` endpoint, we are now able to know the ID of the stock entry and thus query `/stock/entry/{entryId}`.

Before that, it was not possible to know this stock entry id.

Only line 6 and 36 were added to the existing view:
```sql
        GROUP_CONCAT(s.id) AS stock_ids,
```

### DIFF
--- a/migrations/0256.sql
+++ b/migrations/0256.sql
@@ -1,0 +1,54 @@
+DROP VIEW IF EXISTS stock_current;
+CREATE VIEW stock_current
+AS
+SELECT
+        pr.parent_product_id AS product_id,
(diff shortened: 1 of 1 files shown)


berrnd:
This project does not accept code contributions (ref [Contributing](https://github.com/grocy/grocy?tab=contributing-ov-file#readme)).

